PR #58, opened 2020-12-22
fix: IndexedDB failing to fallback in private mode on Firefox

CDDelta:


CDDelta:
Fixed in commit ca76a5911a02d6f6ca43c4824fff329402f12d95

PR #71, opened 2021-01-27
nitpicks from fix: folder child paths not updated after rename

jdaev:


CDDelta:
We try and prefix commits that won't be squashed with standard labels that describe their reason. Please see [here](https://github.com/ardriveapp/ardrive-web/blob/dev/CONTRIBUTING.md) :)

PR #79, opened 2021-02-01
feat: file type thumbnails

jdaev:


lib/pages/drive_detail/components/drive_detail_data_table.dart
```diff
@@ -111,3 +116,27 @@ DataRow _buildFileRow({
         ),
       ],
     );
+
+Icon _buildFileIcon(FileEntry file) {
+  final extension = file.name.split('.').last;
```
CDDelta, line 121: Instead of checking the extension could you parse the file `contentType` instead?

CDDelta (dismissed):

lib/pages/drive_detail/components/drive_detail_data_table.dart
jdaev: There doesn't seem to be any content type or mime on FileEntry. Figured this was the simplest approach.

PR #81, opened 2021-02-03
feat: surface file network confirmation status in table view

CDDelta:
This PR surfaces the confirmation status of files on Arweave in the drive detail table view.


jdaev (approved):
Probably gonna have to refactor to avoid conflicts when merging mobile view later.

CDDelta:
> Probably gonna have to refactor to avoid conflicts when merging mobile view later.

Shouldn't be too bad right? Only `_buildFileRow()` was changed significantly.

PR #83, opened 2021-02-04
fix: mobile support for the folder and file data table

jdaev:
PR changes the datatable into a listview


lib/pages/drive_detail/components/drive_detail_data_list.dart
```diff
@@ -0,0 +1,83 @@
+part of '../drive_detail_page.dart';
+
+Widget _buildDataList(BuildContext context, DriveDetailLoadSuccess state) =>
+    ListView(
+      // sortColumnIndex: DriveOrder.values.indexOf(state.contentOrderBy),
+      // sortAscending: state.contentOrderingMode == OrderingMode.asc,
```
CDDelta, line 6: nit: remove comments here

lib/pages/drive_detail/components/drive_detail_data_list.dart
```diff
@@ -0,0 +1,83 @@
+part of '../drive_detail_page.dart';
+
+Widget _buildDataList(BuildContext context, DriveDetailLoadSuccess state) =>
+    ListView(
```
CDDelta, line 4: The Figma includes separators on the ListView. We should probably have that in here too.

lib/pages/drive_detail/drive_detail_page.dart
```diff
@@ -30,59 +31,122 @@ class DriveDetailPage extends StatelessWidget {
...
+                                    Text(
+                                      state.currentDrive.name,
+                                      style:
+                                          Theme.of(context).textTheme.headline5,
+                                    ),
+                                    SizedBox(
+                                      height: 16,
+                                    ),
+                                    Align(
+                                      alignment: Alignment.centerRight,
```
CDDelta, line 126: Looks like `DriveDetailActionRow` is supposed to be right-aligned here but doesn't show up as so?

Anyway, we should look into where else we can put these actions as opposed to having them in the header like on desktop. Perhaps in a bottom sheet that pops up?

Discuss in Slack.

lib/misc/misc.dart
```diff
@@ -1,5 +1,6 @@
 export 'resources.dart';
 
 const kDriveNameRegex = kFolderNameRegex;
-const kFolderNameRegex = r'^[^\/\\\<\>\:\"\?]+(?<!\.)$';
-const kFileNameRegex = r'^[^\/\\]+(?<!\.)$';
+const kFolderNameRegex = r'^[^\/\\\<\>\:\"\?\*]+$';
+const kFileNameRegex = r'^[^\/\\\*]+$';
+const kTrimTrailingRegex = r'[^\s \.]+$';
```
CDDelta, line 6: Is this the best way we can implement these Regexes? If so, please add a comment explaining why here.

CDDelta (changes requested):

lib/misc/misc.dart
jdaev: It was the only way that worked. I tried doing some of it in a Validator.any function inside the form but that kept throwing an invalid type error.

CDDelta: Sorry, I meant to add a comment in the code explaining why we're doing this so we know in the future. Hopefully, we can swap over when Safari supports negative lookbehinds.

jdaev: Okay

lib/pages/drive_detail/components/drive_detail_data_list.dart
```diff
@@ -41,43 +39,53 @@ Widget _buildDataList(BuildContext context, DriveDetailLoadSuccess state) =>
...
 }) =>
-    ListTile(
-      onTap: () => onPressed(),
-      selected: selected,
-      leading: Padding(
-        padding: const EdgeInsetsDirectional.only(end: 8.0),
-        child: const Icon(Icons.folder),
-      ),
-      title: Text(folder.name),
+    Column(
```
CDDelta, line 48: You should be able to do [ListView.separated](https://api.flutter.dev/flutter/widgets/ListView/ListView.separated.html) right?

jdaev: Sure

jdaev: Think this is simpler though. I'd have to merge the subfolders and files into a single list to use it in an itembuilder.

CDDelta: True. You should be able to remove the `Column` widget and flatten the tree a bit by only returning a `ListTile` from `buildFolderListTile()` and `_buildFileListTile()` and then using `intersperseOuter()` on the resulting Widget list.

jdaev: Alright

jdaev (commented):
Changes done

lib/pages/drive_detail/components/drive_detail_data_list.dart
```diff
@@ -3,39 +3,43 @@ part of '../drive_detail_page.dart';
...
+                  final bloc = context.read<DriveDetailCubit>();
+                  if (file.id == state.selectedItemId) {
+                    bloc.toggleSelectedItemDetails();
+                  } else {
+                    await bloc.selectItem(file.id);
+                  }
+                },
+              ),
+            )
+            .intersperseOuter(Divider()),
```
CDDelta, line 42: Could you do `intersperseOuter()` on the entire data widget list instead? Doing it like this can lead to inconsistent results with views that have folders but no files.

jdaev: Done

CDDelta: Sorry. I miswrote here. We should use `intersperse()` instead of `intersperseOuter()`. I believe that's more in line with general UI designs?

Beyond that. LGTM.

CDDelta:
As soon as you make the change with `intersperse()`, we can merge. Sorry bout that

CDDelta (approved):

PR #84, opened 2021-02-07
ui: update USD converted cost display

CDDelta:
This PR updates the upload dialog to reduce the number of decimals for the USD converted upload cost.


jdaev (approved):
Looks good.

PR #86, opened 2021-02-07
feat: direct public drive attach

jdaev:
Pass name through share link and attach drive automatically.

### DIFF
--- a/lib/blocs/drive_attach/drive_attach_cubit.dart
+++ b/lib/blocs/drive_attach/drive_attach_cubit.dart
@@ -56,35 +56,33 @@ class DriveAttachCubit extends Cubit<DriveAttachState> {
       }
     });
     if (driveName != null && driveName.isNotEmpty) {
-      submit(autoDriveId: initialDriveId, autoDriveName: driveName);
+      form.control('driveId').value = initialDriveId;
+      form.control('name').value = driveName;
+      submit();
     }
   }
 
-  void submit({String autoDriveId, String autoDriveName}) async {
-    if ((autoDriveId == null) && (autoDriveName == null)) {
-      form.markAllAsTouched();
+  void submit() async {
+    form.markAllAsTouched();
 
-      if (form.invalid) {
-        return;
-      }
+    if (form.invalid) {
+      return;
     }
 
     emit(DriveAttachInProgress());
 
     try {
-      final String driveId = autoDriveId ?? form.control('driveId').value;
-      final driveName =
-          autoDriveName ?? form.control('name').value.toString().trim();
+      final String driveId = form.control('driveId').value;
+      final driveName = form.control('name').value.toString().trim();
 
       final driveEntity = await _arweave.getLatestDriveEntityWithId(driveId);
-      if ((autoDriveId == null) && (autoDriveName == null)) {
-        if (driveEntity == null) {
-          form
-              .control('driveId')
-              .setErrors({AppValidationMessage.driveNotFound: true});
-          emit(DriveAttachFailure());
-          return;
-        }
+
+      if (driveEntity == null) {
+        form
+            .control('driveId')
+            .setErrors({AppValidationMessage.driveNotFound: true});
+        emit(DriveAttachFailure());
+        return;
       }
 
       await _driveDao.writeDriveEntity(name: driveName, entity: driveEntity);
--- a/lib/blocs/drive_attach/drive_attach_cubit.dart
+++ b/lib/blocs/drive_attach/drive_attach_cubit.dart
@@ -56,27 +56,31 @@ class DriveAttachCubit extends Cubit<DriveAttachState> {
     });
   }
 
-  void submit() async {
-    form.markAllAsTouched();
+  void submit({String autoDriveId, String autoDriveName}) async {
+    if ((autoDriveId == null) && (autoDriveName == null)) {
+      form.markAllAsTouched();
 
-    if (form.invalid) {
-      return;
+      if (form.invalid) {
+        return;
+      }
     }
 
     emit(DriveAttachInProgress());
 
     try {
-      final String driveId = form.control('driveId').value;
-      final driveName = form.control('name').value.toString().trim();
+      final String driveId = autoDriveId ?? form.control('driveId').value;
+      final driveName =
+          autoDriveName ?? form.control('name').value.toString().trim();
 
       final driveEntity = await _arweave.getLatestDriveEntityWithId(driveId);
-
-      if (driveEntity == null) {
-        form
-            .control('driveId')
-            .setErrors({AppValidationMessage.driveNotFound: true});
-        emit(DriveAttachFailure());
-        return;
+      if ((autoDriveId == null) && (autoDriveName == null)) {
+        if (driveEntity == null) {
+          form
+              .control('driveId')
+              .setErrors({AppValidationMessage.driveNotFound: true});
+          emit(DriveAttachFailure());
+          return;
+        }
       }
 
       await _driveDao.writeDriveEntity(name: driveName, entity: driveEntity);
@@ -111,6 +115,7 @@ class DriveAttachCubit extends Cubit<DriveAttachState> {
   @override
   void onError(Object error, StackTrace stackTrace) {
     emit(DriveAttachFailure());
+    print(stackTrace);
     super.onError(error, stackTrace);
   }
 }
--- a/lib/blocs/drive_attach/drive_attach_cubit.dart
+++ b/lib/blocs/drive_attach/drive_attach_cubit.dart
@@ -22,6 +22,7 @@ class DriveAttachCubit extends Cubit<DriveAttachState> {
 
   DriveAttachCubit({
     String initialDriveId,
+    String driveName,
     @required ArweaveService arweave,
     @required DriveDao driveDao,
     @required SyncCubit syncBloc,
@@ -54,6 +55,9 @@ class DriveAttachCubit extends Cubit<DriveAttachState> {
         form.control('driveId').updateValue(initialDriveId);
       }
     });
+    if (driveName != null && driveName.isNotEmpty) {
+      submit(autoDriveId: initialDriveId, autoDriveName: driveName);
+    }
   }
 
   void submit({String autoDriveId, String autoDriveName}) async {
@@ -115,7 +119,6 @@ class DriveAttachCubit extends Cubit<DriveAttachState> {
   @override
   void onError(Object error, StackTrace stackTrace) {
     emit(DriveAttachFailure());
-    print(stackTrace);
     super.onError(error, stackTrace);
   }
 }
--- a/lib/blocs/drive_detail/drive_detail_cubit.dart
+++ b/lib/blocs/drive_detail/drive_detail_cubit.dart
@@ -53,47 +53,51 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
       {@required String path,
       DriveOrder contentOrderBy = DriveOrder.name,
       OrderingMode contentOrderingMode = OrderingMode.asc}) {
-    emit(DriveDetailLoadInProgress());
+    try {
+      emit(DriveDetailLoadInProgress());
 
-    unawaited(_folderSubscription?.cancel());
+      unawaited(_folderSubscription?.cancel());
 
-    _folderSubscription =
-        Rx.combineLatest3<Drive, FolderWithContents, ProfileState, void>(
-      _driveDao.driveById(driveId: driveId).watchSingleOrNull(),
-      _driveDao.watchFolderContents(driveId,
-          folderPath: path,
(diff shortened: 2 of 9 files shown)


jdaev:
@CDDelta Any ideas on how to improve this?

lib/blocs/drive_attach/drive_attach_cubit.dart
```diff
@@ -111,6 +115,7 @@ class DriveAttachCubit extends Cubit<DriveAttachState> {
   @override
   void onError(Object error, StackTrace stackTrace) {
     emit(DriveAttachFailure());
+    print(stackTrace);
```
CDDelta, line 118: Do we need to print this? Does `flutter_bloc` not do it properly with `super.onError()`?

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -53,47 +53,51 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
...
+                currentFolder: folderContents,
+                contentOrderBy: contentOrderBy,
+                contentOrderingMode: contentOrderingMode,
+              ),
+            );
+          }
+        },
+      ).listen((_) {});
+    } catch (e, stackTrace) {
+      super.onError(e, stackTrace);
```
CDDelta, line 99: Instead of calling `super.onError()` directly let's override `onError()` here like in some of the other blocs so that other methods in this bloc can use it if it comes up in the future.

lib/blocs/drive_share/drive_share_cubit.dart
```diff
@@ -28,7 +28,8 @@ class DriveShareCubit extends Cubit<DriveShareState> {
     // On web, link to the current origin the user is on.
     // Elsewhere, link to app.ardrive.io.
     final linkOrigin = kIsWeb ? Uri.base.origin : 'https://app.ardrive.io';
-    var driveShareLink = '$linkOrigin/#/drives/${drive.id}';
+    var driveName = drive.name.replaceAll(' ', '+').replaceAll('.', '');
+    var driveShareLink = '$linkOrigin/#/drives/${drive.id}?name=${driveName}';
```
CDDelta, line 32: nit: let's use `final` whenever possible. There should be a lint for this already enabled?

lib/blocs/drive_share/drive_share_cubit.dart
```diff
@@ -28,7 +28,8 @@ class DriveShareCubit extends Cubit<DriveShareState> {
     // On web, link to the current origin the user is on.
     // Elsewhere, link to app.ardrive.io.
     final linkOrigin = kIsWeb ? Uri.base.origin : 'https://app.ardrive.io';
-    var driveShareLink = '$linkOrigin/#/drives/${drive.id}';
+    var driveName = drive.name.replaceAll(' ', '+').replaceAll('.', '');
```
CDDelta, line 31: I believe `encodeQueryComponent()` might be a better fit for this: https://api.dart.dev/stable/2.10.5/dart-core/Uri/encodeQueryComponent.html

lib/blocs/drive_attach/drive_attach_cubit.dart
```diff
@@ -56,27 +56,31 @@ class DriveAttachCubit extends Cubit<DriveAttachState> {
...
 
       final driveEntity = await _arweave.getLatestDriveEntityWithId(driveId);
-
-      if (driveEntity == null) {
-        form
-            .control('driveId')
-            .setErrors({AppValidationMessage.driveNotFound: true});
-        emit(DriveAttachFailure());
-        return;
+      if ((autoDriveId == null) && (autoDriveName == null)) {
```
CDDelta, line 76: nit: are the brackets around the conditions required here? I'm not sure that they add to the readablity but just let me know.

lib/components/drive_attach_form.dart
```diff
@@ -26,6 +26,46 @@ Future<void> promptToAttachDrive(
...
+            } else if (state is DriveAttachFailure) {
+              // Close the progress dialog if the drive attachment fails.
+              Navigator.pop(context);
+            } else if (state is DriveAttachSuccess) {
+              Navigator.pop(context);
+              Navigator.pop(context);
+            }
+          },
+          builder: (context, state) {
+            context.read<DriveAttachCubit>().submit(
```
CDDelta, line 59: Instead of providing the drive name here can the cubit be provided with the name on construction? The cubit can then initiate the drive attachment without an external prompt on it.

We should then be able to share the same code path with `promptToAttachDrive()` (though you'll maybe want to rename it to `attachDrive()` instead). And hide the `DriveAttachForm` when the drive is being auto-attached. To also get the progress dialogues etc to show up, you'll want to break the `BlocConsumer` into a `BlocBuilder` in the form and a `BlocListener` in `attachDrive().

The attach process can take a while so I believe it would be best to have an indicator here.

CDDelta: The problem with your current method is you're calling `DriveAttachCubit.submit()` in a builder which runs again when Flutter rebuilds the widget tree after the drive attach is successful causing another cubit submission.

CDDelta (changes requested):

lib/blocs/drive_attach/drive_attach_cubit.dart
jdaev: It was useful for printing to the web console. I'll remove it

lib/blocs/drive_attach/drive_attach_cubit.dart
jdaev: It does add to the readability at least for me. I usually group conditions but if it's unnecessary I'll remove it.

lib/components/drive_attach_form.dart
jdaev: Alright. Thanks

lib/blocs/drive_attach/drive_attach_cubit.dart
CDDelta: Cool. Let's just keep it then.

lib/blocs/drive_share/drive_share_cubit.dart
```diff
@@ -28,13 +28,16 @@ class DriveShareCubit extends Cubit<DriveShareState> {
     // On web, link to the current origin the user is on.
     // Elsewhere, link to app.ardrive.io.
     final linkOrigin = kIsWeb ? Uri.base.origin : 'https://app.ardrive.io';
-    var driveName = drive.name.replaceAll(' ', '+').replaceAll('.', '');
-    var driveShareLink = '$linkOrigin/#/drives/${drive.id}?name=${driveName}';
+    final driveName = drive.name.replaceAll(' ', '+').replaceAll('.', '');
```
CDDelta, line 31: Does the Uri constructor here not encode `driveName` properly?

lib/blocs/drive_attach/drive_attach_cubit.dart
```diff
@@ -54,6 +55,9 @@ class DriveAttachCubit extends Cubit<DriveAttachState> {
         form.control('driveId').updateValue(initialDriveId);
       }
     });
+    if (driveName != null && driveName.isNotEmpty) {
+      submit(autoDriveId: initialDriveId, autoDriveName: driveName);
```
CDDelta, line 59: Instead of passing the constructor parameters through, could we set the form values and then just have `submit()` without parameters instead?

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -141,4 +141,10 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
     _folderSubscription?.cancel();
     return super.close();
   }
+
+  @override
+  void onError(Object error, StackTrace stackTrace) {
+    emit(DriveDetailLoadNotFound());
```
CDDelta, line 147: Lets not emit a NotFound state here. I believe NotFound leads to some side-effects in some BlocListeners that we don't want.

On further thought, can't you call `onError()` without `super`, without overriding `onError()` here?

CDDelta (changes requested):

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -53,47 +53,51 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
...
+                    drive.ownerAddress == await profile.wallet.getAddress(),
+                currentFolder: folderContents,
+                contentOrderBy: contentOrderBy,
+                contentOrderingMode: contentOrderingMode,
+              ),
+            );
+          }
+        },
+      ).listen((_) {});
+    } catch (e, stackTrace) {
```
CDDelta, line 98: I missed this before. What error are you trying to catch here? It doesn't look like there's anything new to catch.

If you really need to catch something, would it be possible to do a `handleError()` on the stream instead?

lib/blocs/drive_share/drive_share_cubit.dart
CDDelta: Looks like the Uri constructor does encode its inputs. It does encode the `path` at least, the `Copy Link` result is now broken as the `https://` portion of the URL is encoded wrongly.

I believe it'll be easier to revert to the old way the Uri was constructed and just do [`Uri.encodeQueryComponent()`](https://api.dart.dev/stable/2.10.5/dart-core/Uri/encodeQueryComponent.html)

lib/blocs/drive_detail/drive_detail_cubit.dart
jdaev: I don't think so

lib/blocs/drive_detail/drive_detail_cubit.dart
jdaev: There was an uncaught exception being triggered from here. Something to do with dart.js

lib/blocs/drive_share/drive_share_cubit.dart
```diff
@@ -28,7 +28,10 @@ class DriveShareCubit extends Cubit<DriveShareState> {
     // On web, link to the current origin the user is on.
     // Elsewhere, link to app.ardrive.io.
     final linkOrigin = kIsWeb ? Uri.base.origin : 'https://app.ardrive.io';
-    var driveShareLink = '$linkOrigin/#/drives/${drive.id}';
+    final driveName = drive.name.replaceAll(' ', '+').replaceAll('.', '');
```
CDDelta, line 31: If you're already doing Uri.encodeQueryComponent, there should be no need for `replaceAll()` etc right?

The current way will cause drive names with spaces to come out with +s when shared.

CDDelta (changes requested):

lib/blocs/drive_detail/drive_detail_cubit.dart
CDDelta: Just tried it out. Looks like you can.

lib/blocs/drive_detail/drive_detail_cubit.dart
CDDelta: Is catching this crucial to the PR/feature? We should only be catching errors like this if we want to show an error in the UI as the runtime will throw uncaught errors for us.

If you can find a way to reproduce this and file a bug, it can be looked into in a separate issue.

lib/blocs/drive_share/drive_share_cubit.dart
jdaev: Missed that. Thanks

lib/components/drive_attach_form.dart
```diff
@@ -10,38 +10,47 @@ import 'package:reactive_forms/reactive_forms.dart';
...
           arweave: context.read<ArweaveService>(),
           driveDao: context.read<DriveDao>(),
           syncBloc: context.read<SyncCubit>(),
           drivesBloc: context.read<DrivesCubit>(),
         ),
-        child: DriveAttachForm(),
+        child: BlocListener<DriveAttachCubit, DriveAttachState>(
+          listener: (context, state) {
+            if (state is DriveAttachInProgress) {
+              showProgressDialog(context, 'ATTACHING DRIVE...');
```
CDDelta, line 31: Does this not cause us to show two "ATTACHING DRIVE..." dialogs?

Is this possibly what's causing the entire app page to do that sliding animation after the drive is attached?

jdaev: Possibly. Should I remove the one in DriveAttachProgress?

CDDelta: Remove the one in the listener. You'll need to also then update the `Navigator.pop()` calls there so we don't over pop.

jdaev: Okay

CDDelta:
Perfect, that post-attach page jank is gone! LGTM!

CDDelta (approved):

PR #93, opened 2021-02-16
feat: allow users to rename drives

jdaev:


lib/blocs/drive_rename/drive_rename_cubit.dart
```diff
@@ -0,0 +1,124 @@
...
+        final driveEntity = drive.asEntity();
+
+        final driveTx = await _arweave.prepareEntityTx(
+            driveEntity, profile.wallet, driveKey);
+
+        await _arweave.postTx(driveTx);
+        await _driveDao.writeToDrive(drive);
+
+        driveEntity.txId = driveTx.id;
+
```
CDDelta, line 85: You need to set the `ownerAddress` on `driveEntity` here using `profile.address` or something similar to that.

lib/blocs/drive_rename/drive_rename_cubit.dart
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // Check that the current drive does not already have a drive with the target file name.
+    final drivesWithName = (await _driveDao.allDrives().get())
+        .where((element) => element.name == newDriveName);
+    final nameAlreadyExists = drivesWithName.isNotEmpty;
+
+    if (nameAlreadyExists) {
+      control.markAsTouched();
+      return {AppValidationMessage.nameAlreadyPresent: true};
```
CDDelta, line 113: We need to add a new `AppValidationMessage` like `driveNameAlreadyPresent`. Also maybe rename `nameAlreadyPresent` to something else like `fsEntryNameAlreadyPresent`.

CDDelta:
LGTM! Works perfectly. Just update the validation message and we should be good to merge.

agsuy (approved):
Approved. We could rework the UI later on

PR #99, opened 2021-02-25
Feature/handle sync fail on no new entities

jdaev:


agsuy (approved):
Should fix sync on no new entities. We detected a weird behavior on Firefox under Linux that should be revisited in the future

PR #108, opened 2021-03-09
ui: added version to app drawer

jdaev:


agsuy:
Please put it either left aligned or right aligned as it is currently centered.

agsuy (approved):
Seems good. I will test this on staging before going live.

PR #109, opened 2021-03-14
feat/251 Add more details to drive, folder & file details

jdaev:


agsuy (approved):
Phil tested this. After removing that dummy test we are good to go. Merging.

PR #117, opened 2021-04-09
 fix/231 Invalid file preview link 

jdaev:


agsuy (approved):

agsuy (changes requested):
We found an issue uploading files as that seems to trigger multiple dialogs on Firefox.

We need to check that our before merging this.

agsuy (commented):
Seems ok! Lets wait for the test results before merging

agsuy (approved):

PR #118, opened 2021-04-10
chore/277 upgrade dependencies and migrate to null safety

jdaev:
Upgrading the dependencies here will require the app to be null safe, otherwise it won't build.


agsuy (commented):
Seems ok!

Lets check build status first before merging.

PR #119, opened 2021-04-10
 feat/276 Abort HTTP requests with dio 

jdaev:
Adds an abortDownload function to FileDownloadCubit which closes the connection.


agsuy (approved):
Looking good

PR #128, opened 2021-04-21
Feat/271 toggle testnet by version

jdaev:


agsuy (commented):
Lets test it with our testnet before merging

agsuy (changes requested):
There is not point on having the toggle once u re logged in as now to change it u need to login, switch to test, use another wallet an login again.

Insted we should show it at login page so u can easily switch before login.

agsuy (approved):

PR #129, opened 2021-04-22
fix/279 Cannot see drives in drive list

jdaev:
Make appdrawer scrollable
Make datatable scrollable without hiding actions


agsuy (changes requested):
When you click an Drive at bottom the view goes straight to top and u can not see selected drive at all without scrolling again.

Scroll should remain static when you select your drive

agsuy (approved):
There is an edge case on bottom element but this is acceptable.

PR #131, opened 2021-04-25
refactor: replace dio with httpclient

jdaev:


agsuy (approved):
Seems ok. We will test it on staging env.

Merging

PR #133, opened 2021-04-27
fix: revert back to http

jdaev:


agsuy:
Images are not corrupted now.

PR #146, opened 2021-05-26
task: Changed gateway to gateway.amplify.host

jdaev:


agsuy:
did not work as expected

PR #149, opened 2021-06-07
nit: changed precision of ar displayed

jdaev:
Precision changed to 5 digits


agsuy (approved):
Working as expected merging to staging.

PR #150, opened 2021-06-07
ui/disable new button on insufficient ar balance

jdaev:


agsuy (approved):
Working as expected. Missing some wallet balance tests that will be performed on staging.

Merging.

PR #151, opened 2021-06-07
fix: incorrect textfield label color RM-300

jdaev:
Incorrect colors caused by breaking change in ThemeData secondaryColor property. 
Use colorscheme instead of secondary color.


agsuy (approved):
Its all red now! Merging.

PR #153, opened 2021-06-16
Task/rm 314 show unlock screen on opening and closing

jdaev:


lib/blocs/profile/profile_cubit.dart
```diff
@@ -35,17 +35,22 @@ class ProfileCubit extends Cubit<ProfileState> {
     final profile = await _profileDao.defaultProfile().getSingleOrNull();
 
     if (profile != null) {
-      if (profile.encryptedWallet.isEmpty) {
+      if (profile.isArConnect == 1) {
```
arielmelendez, line 38: Using the early return pattern makes it much easier to track happy-path code. I used to think that the debuggability benefits of a single point of return were worthwhile, but came to appreciate the enhanced readability of early returns, guard clauses, etc. Here, that could look like:
`
final profile = await _profileDao.defaultProfile().getSingleOrNull();
    
    // Profile unavailable - route to new profile screen
    if (profile == null) {
      emit(ProfilePromptAdd());
      return;
    }

    // json wallet present - route to login screen
    if (profile.isArConnect != 1) {
      emit(ProfilePromptLogIn());
      return;
    }

    // ArConnect extension missing - route to profile screen
    if (!arconnect.isExtensionPresent()) {
      emit(ProfilePromptAdd());
      return;
    }

    // ArConnect connected to expected wallet - route to login screen
    if (await arconnect.checkPermissions() &&
        profile.walletPublicKey == await arconnect.getPublicKey()) {
        emit(ProfilePromptLogIn());
        return;
    }

    // Unexpected ArConnect state - clean up and route to profile screen
    await _db.transaction(() async {
      for (final table in _db.allTables) {
        await _db.delete(table).go();
      }
    });
    emit(ProfilePromptAdd());
`

A few other things:
1) isArconnect is a boolean code smell. Consider using an enum to track the wallet type for the profile.
2) There's an opportunity here to unit test a pure function. You could create a function that takes as inputs a profile and an arconnect object and returns a ProfileState. If you make an abstract class to put in front of the arconnect object (perhaps using a mixin? Still learning the type system in Dart) that has all the functions we expect from (or simply wraps) the arconnect library, then you can create a stub arconnect interface for testing to give you the desired answers to the function calls to isExtensionPresent, checkPerissions, and getPublicKey. Let me know if you have more questions about this.
3) The _db connection code is pretty raw and low level. It would be nice to have that wrapped up in a DAO.

lib/blocs/profile_add/profile_add_cubit.dart
```diff
@@ -156,10 +156,12 @@ class ProfileAddCubit extends Cubit<ProfileAddState> {
       await _profileDao.addProfile(username, password, _wallet);
     } else {
       final walletAddress = await arconnect.getWalletAddress();
+      final walletPublicKey = await arconnect.getPublicKey();
```
arielmelendez, line 159: A thought... If the wallet address is just a SHA256 hash of the public key, do we really need both back from the arconnect lib? Handling both pieces of data separately feels brittle. It could be beneficial to have a new unit tested type (e.g. ARWalletInfo) that stores a public key internally and has a unit-tested accessor function for the wallet address. We might also want to do a head check somewhere that arconnect returns us a valid pubkey/address combination.

lib/models/daos/profile_dao.dart
```diff
@@ -56,7 +56,10 @@ class ProfileDao extends DatabaseAccessor<Database> with _$ProfileDaoMixin {
   /// Adds the specified profile and returns a profile key that was used to encrypt the user's wallet
   /// and can be used to encrypt the user's drive keys.
   Future<SecretKey> addProfile(
-      String username, String password, Wallet wallet) async {
+    String username,
```
arielmelendez, line 59: I realize this is just a formatting change, but some quick notes on this function. It's called "addProfile" but it takes profile inputs and then does work on them before actually doing the adding. Those operations include:
1) Encrypting the wallet
2) Builds a profile companion with the inputs
Consider factoring those pieces out separately to unit testable functions.

lib/models/daos/profile_dao.dart
```diff
@@ -66,21 +69,26 @@ class ProfileDao extends DatabaseAccessor<Database> with _$ProfileDaoMixin {
...
       nonce: profileSalt,
     );
-
     await into(profiles).insert(
       ProfilesCompanion.insert(
         id: await wallet.getAddress(),
         username: username,
         encryptedWallet: encryptedWallet.concatenation(nonce: false),
         keySalt: profileSalt,
+        isArConnect: 0,
```
arielmelendez, line 78: Same comment here about isArConnect being a code smell.

arielmelendez (changes requested):
Take a look at my comments and let me know whether you think any are addressable in this PR. Primarily I'm looking for some more tests to be added to verify your changes here.

lib/blocs/profile/profile_cubit.dart
jdaev: 1. Noted. isArConnect is actually an int with 0 and 1 since SQLite does not support bools. If I have to parse it as an enum I guess I could rename isArconnect to loginType, and have an enum with the values JSON and ArConnect and compare those.
2. Yeah, I get what you mean. If you want me to make an arconnect mock class, I can do it.
3. I'll see if I can create a clearDatabase function in driveDao and move that there.

lib/blocs/profile_add/profile_add_cubit.dart
jdaev: True, I just did it that way because they were handled individually for the JSON profiles. I can change it, but it's probably better done in another PR

lib/models/daos/profile_dao.dart
jdaev: I split those

lib/blocs/profile/profile_cubit.dart
jdaev: The DatabaseDao would be better in another PR in my opinion.

arielmelendez: >  isArConnect is actually an int with 0 and 1 since SQLite does not support bools.
We should not let the implementation details of a data store affect the interfaces that we provide in the classes that control our domain and business logic. Those implementation details should be dealt with in a very narrow and focused way in a DAO. Our business and domain logic should be optimized for correctness and readability while preserving the mantra that code should be "easy to use correctly, and hard to use incorrectly."

>  If you want me to make an arconnect mock class, I can do it.
What I'd like for us to be able to do is to "work with abstractions, not concretions". If we were to create a basic, testable, input output function like I described, we wouldn't want to pass a concrete type like "arconnect" in as a parameter. Instead, we'd want to pass through an abstract type that provides a contract for the ways in which we should be able to talk to that object, and then use the specific interface functions that we need in that function without caring what the underlying concrete implementation (e.g. arconnect class) actually is. Refer back to that refactoring for testability article that I shared earlier in the week (https://www.swiftbysundell.com/articles/refactoring-swift-code-for-testability/). That code was using a "Router" class internally which is not easily mockable. So they introduced a protocol (read: abstraction) called "CheckoutPageOpener" and then let the compiler know that Router already conforms to it. When ShoppingCart is constructed, its CheckoutPageOpener dependency is provided by whomever builds it (which is a separate concern) and can be anything - a stub/mock, a real Router, something else. It doesn't matter. As long as it conforms to the CheckoutPageOpener abstraction.

The idea here is that we might need an abstraction that helps us be able to tell that a wallet is available and to reset it if not. Imagine if we had an abstraction like "WalletHealthChecking" that has a function like "bool walletIsReady()" and that we create a Profile mixin that answers that question for different wallet types. Then the whole function above could collapse to something like (please excuse syntax):
`
ProfileState profileStateForProfile(Profile profile) {
  if (profile == null || !profile.walletIsReady()) {
    // Profile or wallet are unavailable - route to new profile screen
    return ProfilePromptAdd());
  }

  return ProfilePromptLogIn();
}

Future<void> promptToAuthenticate() async {
    final profile = await _profileDao.defaultProfile().getSingleOrNull();
    emit(profileStateForProfile(profile));
}
`
Arguably the resetting of the wallet should take place elsewhere (maybe any time we bring up the ProfilePromptAdd screen?). But anyway, the function now has way fewer conditions to test and all the implementation details now live down with classes that are better suited for handling them. There's a few more steps to achieve total testability, but I think you can see where I'm going with this. If not, let's sync up!

jdaev: Understood. But we should discuss this in detail and create another task for it. The profile you see is auto generated code by the moor library based on the SQL queries that are provided. I will need to do more research and experimentation to see how I could abstract that.

.github/workflows/pr.yaml
```diff
@@ -1,9 +1,9 @@
 name: pr check
 on:
   pull_request:
-    branches:
-      - dev
-
+    branches-ignore:
```
arielmelendez, line 4: Please merge the latest updates from `task/296-attach-arconnect-session-to-wallet-address` branch

arielmelendez (changes requested):
Please merge the latest updates from task/296-attach-arconnect-session-to-wallet-address branch

lib/blocs/profile/profile_cubit.dart
```diff
@@ -34,28 +35,42 @@ class ProfileCubit extends Cubit<ProfileState> {
...
+      for (final table in _db.allTables) {
+        await _db.delete(table).go();
+      }
+    });
     emit(ProfilePromptAdd());
   }
 
-  Future<void> unlockDefaultProfile(String password, bool isArconnect) async {
+  Future<void> unlockDefaultProfile(
+      String password, ProfileType profileType) async {
```
arielmelendez, line 73: Stray newline?

lib/blocs/profile_add/profile_add_cubit.dart
```diff
@@ -20,7 +21,8 @@ class ProfileAddCubit extends Cubit<ProfileAddState> {
   FormGroup form;
 
   Wallet _wallet;
-  bool isArconnect;
+  ProfileType profileType;
```
arielmelendez, line 24: Add underscores for private members

lib/blocs/profile_add/profile_add_cubit.dart
```diff
@@ -20,7 +21,8 @@ class ProfileAddCubit extends Cubit<ProfileAddState> {
...
+  String walletAddressOnLoad;
```
arielmelendez, line 25: Consider `_lastKnowWalletAddress`

lib/pages/profile_auth/components/profile_auth_fail_screen.dart
```diff
@@ -25,7 +25,7 @@ class ProfileAuthFailScreen extends StatelessWidget {
             ),
             const SizedBox(height: 32),
             Text(
-              'Please check arconnect permissions and try again',
+              'Sorry, login failed. Please try again.',
```
arielmelendez, line 28: Move this to resources.dat for localization purposes.

lib/pages/profile_auth/components/profile_auth_unlock_screen.dart
```diff
@@ -6,67 +6,80 @@ import 'package:flutter/material.dart';
...
-                        ),
-                        const SizedBox(height: 16),
-                        TextButton(
-                          child: Text(
-                            'Forget wallet and change profile',
-                            textAlign: TextAlign.center,
-                          ),
-                          onPressed: () =>
-                              context.read<ProfileCubit>().logoutProfile(),
+          builder: (context, state) {
```
arielmelendez, line 26: Move literal strings that are shown to users into resources.dat

lib/blocs/profile_unlock/profile_unlock_cubit.dart
```diff
@@ -38,7 +46,12 @@ class ProfileUnlockCubit extends Cubit<ProfileUnlockState> {
     if (form.invalid) {
       return;
     }
-
+    if (_profileType == ProfileType.ArConnect &&
```
arielmelendez, line 49: Consider switching on the profile type here in case we add more profile types in the future.

.github/workflows/pr.yaml
jdaev: [Done](https://github.com/ardriveapp/ardrive-web/pull/153/commits/76bcf0081682da400882fbcce1b17b3a5badb28f)

lib/blocs/profile/profile_cubit.dart
jdaev: It's just the vscode formatter.

lib/blocs/profile_add/profile_add_cubit.dart
jdaev: [Done](https://github.com/ardriveapp/ardrive-web/pull/153/commits/0dccc2ebc2881efa29996467275d7dffeee5d676)

lib/blocs/profile_add/profile_add_cubit.dart
jdaev: [Done ](https://github.com/ardriveapp/ardrive-web/pull/153/commits/0dccc2ebc2881efa29996467275d7dffeee5d676)

lib/pages/profile_auth/components/profile_auth_fail_screen.dart
jdaev: Will be done in [this ](https://ardrive.atlassian.net/browse/PE-20?atlOrigin=eyJpIjoiYjAyZDNmNTI0M2Y2NDZlZDlmNjdhYmQ4NmM5YTI2OTIiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)spike

lib/pages/profile_auth/components/profile_auth_unlock_screen.dart
jdaev: Will be done in [this ](https://ardrive.atlassian.net/browse/PE-20?atlOrigin=eyJpIjoiYjAyZDNmNTI0M2Y2NDZlZDlmNjdhYmQ4NmM5YTI2OTIiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)spike

lib/blocs/profile_unlock/profile_unlock_cubit.dart
jdaev: There are changes from 313 that need to be merged into this first. I can do this afterwards.

lib/pages/profile_auth/components/profile_auth_fail_screen.dart
arielmelendez: It's a story, not a spike, but ok thank you!

arielmelendez (approved):

PR #155, opened 2021-06-16
Task/rm 313 add wallet check to profile cubit

jdaev:
Adds a wallet check function to profile cubit that checks if the wallet public key has changed.
```
Future<void> checkForWalletMismatch() async {
    final profile = await _profileDao.defaultProfile().getSingleOrNull();

    if (profile.isArConnect == 1) {
      final currentPublicKey = await arconnect.getPublicKey();
      final savedPublicKey =
          (await _db.profileDao.defaultProfile().getSingleOrNull())
              .walletPublicKey;
      if (currentPublicKey != savedPublicKey) {
        await logoutProfile();
      }
    }
  }
```

This function is called before each transaction in the app, like so 
`    await _profileCubit.checkForWalletMismatch();
`

It has been added to UploadCubit, SyncCubit, FsEntryRenameCubit, FsEntryMoveCubit, FolderCreateCubit, DriveRenameCubit, and DriveCreateCubit. 

A different approach was tried and tested on this [branch](https://github.com/ardriveapp/ardrive-web/tree/task/RM-313-Check-Wallet-before-Transactions), but wasn't used because this is less verbose.

### DIFF
--- a/lib/blocs/drive_create/drive_create_cubit.dart
+++ b/lib/blocs/drive_create/drive_create_cubit.dart
@@ -6,7 +6,6 @@ import 'package:ardrive/services/services.dart';
 import 'package:bloc/bloc.dart';
 import 'package:equatable/equatable.dart';
 import 'package:meta/meta.dart';
-import 'package:pedantic/pedantic.dart';
 import 'package:reactive_forms/reactive_forms.dart';
 
 part 'drive_create_state.dart';
@@ -47,7 +46,7 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
       return;
     }
 
-    unawaited(_profileCubit.checkForWalletMismatch());
+    await _profileCubit.checkForWalletMismatch();
 
     final profile = _profileCubit.state as ProfileLoggedIn;
 
--- a/lib/blocs/drive_create/drive_create_cubit.dart
+++ b/lib/blocs/drive_create/drive_create_cubit.dart
@@ -6,6 +6,7 @@ import 'package:ardrive/services/services.dart';
 import 'package:bloc/bloc.dart';
 import 'package:equatable/equatable.dart';
 import 'package:meta/meta.dart';
+import 'package:pedantic/pedantic.dart';
 import 'package:reactive_forms/reactive_forms.dart';
 
 part 'drive_create_state.dart';
@@ -45,8 +46,11 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
     if (form.invalid) {
       return;
     }
+
+    unawaited(_profileCubit.checkForWalletMismatch());
+
     final profile = _profileCubit.state as ProfileLoggedIn;
-    final wallet = profile.wallet;
+
     final minimumWalletBalance = BigInt.from(10000000);
     if (profile.walletBalance <= minimumWalletBalance) {
       emit(DriveCreateZeroBalance());
--- a/lib/blocs/drive_create/drive_create_cubit.dart
+++ b/lib/blocs/drive_create/drive_create_cubit.dart
@@ -46,7 +46,9 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
       return;
     }
 
-    await _profileCubit.checkForWalletMismatch();
+    if (await _profileCubit.checkForWalletMismatch()) {
+      return;
+    }
 
     final profile = _profileCubit.state as ProfileLoggedIn;
 
--- a/lib/blocs/drive_create/drive_create_cubit.dart
+++ b/lib/blocs/drive_create/drive_create_cubit.dart
@@ -47,6 +47,7 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
     }
 
     if (await _profileCubit.checkForWalletMismatch()) {
+      emit(DriveCreateWalletMismatch());
       return;
     }
 
--- a/lib/blocs/drive_create/drive_create_cubit.dart
+++ b/lib/blocs/drive_create/drive_create_cubit.dart
@@ -46,7 +46,7 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
       return;
     }
 
-    if (await _profileCubit.checkForWalletMismatch()) {
+    if (await _profileCubit.logoutIfWalletMismatch()) {
       emit(DriveCreateWalletMismatch());
       return;
     }
--- a/lib/blocs/drive_create/drive_create_cubit.dart
+++ b/lib/blocs/drive_create/drive_create_cubit.dart
@@ -45,7 +45,13 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
     if (form.invalid) {
       return;
     }
-
+    final profile = _profileCubit.state as ProfileLoggedIn;
+    final wallet = profile.wallet;
+    final minimumWalletBalance = BigInt.from(10000000);
+    if (profile.walletBalance <= minimumWalletBalance) {
+      emit(DriveCreateZeroBalance());
+      return;
+    }
     emit(DriveCreateInProgress());
 
     try {
--- a/lib/blocs/drive_rename/drive_rename_cubit.dart
+++ b/lib/blocs/drive_rename/drive_rename_cubit.dart
@@ -64,7 +64,7 @@ class DriveRenameCubit extends Cubit<DriveRenameState> {
       final newName = form.control('name').value.toString().trim();
       final profile = _profileCubit.state as ProfileLoggedIn;
       final driveKey = await _driveDao.getDriveKey(driveId, profile.cipherKey);
-
+      await _profileCubit.checkForWalletMismatch();
       emit(DriveRenameInProgress());
 
       await _driveDao.transaction(() async {
--- a/lib/blocs/drive_rename/drive_rename_cubit.dart
+++ b/lib/blocs/drive_rename/drive_rename_cubit.dart
@@ -65,6 +65,7 @@ class DriveRenameCubit extends Cubit<DriveRenameState> {
       final profile = _profileCubit.state as ProfileLoggedIn;
       final driveKey = await _driveDao.getDriveKey(driveId, profile.cipherKey);
       if (await _profileCubit.checkForWalletMismatch()) {
+        emit(DriveRenameWalletMismatch());
         return;
       }
       emit(DriveRenameInProgress());
--- a/lib/blocs/drive_rename/drive_rename_cubit.dart
+++ b/lib/blocs/drive_rename/drive_rename_cubit.dart
@@ -64,7 +64,7 @@ class DriveRenameCubit extends Cubit<DriveRenameState> {
       final newName = form.control('name').value.toString().trim();
       final profile = _profileCubit.state as ProfileLoggedIn;
       final driveKey = await _driveDao.getDriveKey(driveId, profile.cipherKey);
-      if (await _profileCubit.checkForWalletMismatch()) {
+      if (await _profileCubit.logoutIfWalletMismatch()) {
         emit(DriveRenameWalletMismatch());
         return;
       }
--- a/lib/blocs/drive_rename/drive_rename_cubit.dart
+++ b/lib/blocs/drive_rename/drive_rename_cubit.dart
@@ -64,7 +64,9 @@ class DriveRenameCubit extends Cubit<DriveRenameState> {
       final newName = form.control('name').value.toString().trim();
       final profile = _profileCubit.state as ProfileLoggedIn;
       final driveKey = await _driveDao.getDriveKey(driveId, profile.cipherKey);
-      await _profileCubit.checkForWalletMismatch();
+      if (await _profileCubit.checkForWalletMismatch()) {
+        return;
+      }
       emit(DriveRenameInProgress());
 
       await _driveDao.transaction(() async {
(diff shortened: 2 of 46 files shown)


lib/blocs/drive_create/drive_create_cubit.dart
```diff
@@ -45,8 +46,11 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
     if (form.invalid) {
       return;
     }
+
+    unawaited(_profileCubit.checkForWalletMismatch());
```
arielmelendez, line 50: This does the following asynchronously:
- Gets the default profile
- Check the arconnect wallet state if necessary
- Gets the default wallet from the DB to get props from it
- Compares that to local state in the profile cubit
- May optionally kick off a profile logout which: disconnects arconnect -> emits new state -> deletes tables from the db -> prompts to authenticate

It then goes on to make various other state changes to the cubit synchronously. That means there's race conditions here, at a minimum, for the state management of this cubit. There's lots of other awaited activity on shared objects down below as well (read: likely more race conditions in the overall system).

When can't we await here?

arielmelendez (changes requested):
See my comment about race conditions.

lib/blocs/drive_rename/drive_rename_cubit.dart
```diff
@@ -64,7 +64,7 @@ class DriveRenameCubit extends Cubit<DriveRenameState> {
       final newName = form.control('name').value.toString().trim();
       final profile = _profileCubit.state as ProfileLoggedIn;
       final driveKey = await _driveDao.getDriveKey(driveId, profile.cipherKey);
-
+      await _profileCubit.checkForWalletMismatch();
```
arielmelendez, line 67: Do we have any kind of wait cursors or anything that we can throw up to gate user inputs when we're making these blocking calls?

lib/blocs/drive_create/drive_create_cubit.dart
jdaev: That was an oversight on my part. Thank you for catching that.

lib/blocs/drive_rename/drive_rename_cubit.dart
jdaev: As of now, no.

lib/blocs/fs_entry_move/fs_entry_move_state.dart
```diff
@@ -45,6 +45,10 @@ class FolderEntryMoveFailure extends FsEntryMoveState {
   FolderEntryMoveFailure() : super(isMovingFolder: true);
 }
 
+class FolderEntryMoveWalletMismatch extends FsEntryMoveState {
+  FolderEntryMoveWalletMismatch() : super(isMovingFolder: false);
```
arielmelendez, line 49: As discussed, this should be "true" even though we're not using it.

lib/blocs/fs_entry_rename/fs_entry_rename_state.dart
```diff
@@ -31,6 +31,10 @@ class FolderEntryRenameFailure extends FsEntryRenameState {
   FolderEntryRenameFailure() : super(isRenamingFolder: true);
 }
 
+class FolderEntryRenameWalletMismatch extends FsEntryRenameState {
+  FolderEntryRenameWalletMismatch() : super(isRenamingFolder: false);
```
arielmelendez, line 35: As discussed, this should be "true" even though we don't use the value.

lib/blocs/profile/profile_cubit.dart
```diff
@@ -69,6 +69,23 @@ class ProfileCubit extends Cubit<ProfileState> {
     emit(ProfilePromptAdd());
   }
 
+  Future<bool> checkForWalletMismatch() async {
```
arielmelendez, line 72: As discussed, rename to `logoutIfWalletMismatch`. Also add a docblock above to explain the cases in which this might happen AND what the return value means.

lib/blocs/profile/profile_cubit.dart
```diff
@@ -69,6 +69,24 @@ class ProfileCubit extends Cubit<ProfileState> {
     emit(ProfilePromptAdd());
   }
 
+  /// Returns true if there has been a wallet change, or permissions change on another tab.
```
arielmelendez, line 72: Consider:
`  /// Returns true if a logout flow is initiated as a result of a detected wallet or permissions change
`
It highlights that a logout flow takes place, the reasons why, and the meaning of the return value.

lib/blocs/profile/profile_cubit.dart
```diff
@@ -69,6 +69,24 @@ class ProfileCubit extends Cubit<ProfileState> {
...
+  Future<bool> logoutIfWalletMismatch() async {
+    final profile = await _profileDao.defaultProfile().getSingleOrNull();
+    var isMismatch = false;
+
+    if (profile != null && profile.profileType == ProfileType.ArConnect.index) {
```
arielmelendez, line 77: Should we force a logout flow if the profile is missing for some reason?

lib/blocs/profile/profile_cubit.dart
```diff
@@ -69,6 +69,24 @@ class ProfileCubit extends Cubit<ProfileState> {
...
+      final currentPublicKey = await arconnect.getPublicKey();
+      final savedPublicKey =
+          (await _db.profileDao.defaultProfile().getSingleOrNull())
+              .walletPublicKey;
```
arielmelendez, line 81: This could NPE if getSingleOrNull returns null, yes?

lib/blocs/profile_unlock/profile_unlock_cubit.dart
```diff
@@ -55,6 +63,41 @@ class ProfileUnlockCubit extends Cubit<ProfileUnlockState> {
...
+            checkDriveKey,
+          );
+
+          // If the private drive could not be decoded, the password is incorrect.
+          if (privateDrive == null) {
+            form
+                .control('password')
+                .setErrors({AppValidationMessage.passwordIncorrect: true});
+
+            form
```
arielmelendez, line 93: Is this copy-pasta?

lib/blocs/profile_unlock/profile_unlock_cubit.dart
```diff
@@ -55,6 +63,41 @@ class ProfileUnlockCubit extends Cubit<ProfileUnlockState> {
...
+          final checkDriveId = privateDriveTxs.first.getTag(EntityTag.driveId);
+          final signature = arconnect.getSignature;
+
+          final checkDriveKey = await deriveDriveKey(
+            signature,
+            checkDriveId,
+            password,
+          );
+
+          final privateDrive = await _arweave.getLatestDriveEntityWithId(
```
arielmelendez, line 82: In the case of a correct password, is this guaranteed to return a drive (e.g. in the case of a brand new user)?

lib/blocs/profile_unlock/profile_unlock_cubit.dart
```diff
@@ -55,6 +63,41 @@ class ProfileUnlockCubit extends Cubit<ProfileUnlockState> {
     final String password = form.control('password').value;
 
     try {
```
arielmelendez, line 65: There's a lot going on down below. Can you please add some comments to guide the reader through what they're about to consume and/or wrap some of what it's doing in other functions with sensible function names?

arielmelendez (changes requested):
See my comments above for requested changes.

lib/blocs/profile/profile_cubit.dart
jdaev: Can't think of any reason to. If the profile is missing it means that they aren't logged in in the first place.

lib/blocs/profile/profile_cubit.dart
jdaev: If the profile is null, which is checked before this. I don't know why I'm not using that same profile variable here, but have changed it now.

lib/blocs/profile/profile_cubit.dart
jdaev: Added

lib/blocs/fs_entry_rename/fs_entry_rename_state.dart
jdaev: Fixed

lib/blocs/fs_entry_move/fs_entry_move_state.dart
jdaev: Fixed

lib/blocs/profile/profile_cubit.dart
jdaev: Changed it to /// Returns true if a logout flow is initiated as a result of a detected wallet or permissions change

lib/blocs/profile_unlock/profile_unlock_cubit.dart
jdaev: What do you mean? It's the same process that's used for login.

lib/blocs/profile_unlock/profile_unlock_cubit.dart
jdaev: As long as the user has a private drive.

lib/blocs/profile_unlock/profile_unlock_cubit.dart
jdaev: Done. Added comments

lib/blocs/profile/profile_cubit.dart
arielmelendez: Since the use case for this function is to log people out during an authenticated request if anything looks fishy, I think we need to log them out in the case of a missing profile object.

lib/blocs/profile_unlock/profile_unlock_cubit.dart
jdaev: Simplified it [here](https://github.com/ardriveapp/ardrive-web/pull/155/commits/89b9cd7f034a5d2984117323f04a8e2845cc5dab)

lib/blocs/profile_unlock/profile_unlock_cubit.dart
```diff
@@ -39,6 +47,33 @@ class ProfileUnlockCubit extends Cubit<ProfileUnlockState> {
       emit(ProfileUnlockInitial(username: profile.username));
     }();
   }
+  // Validate the user's password by loading and decrypting a private drive.
+  Future<void> verifyPasswordArconnect(String password) async {
+    final profile = await _profileDao.defaultProfile().getSingle();
+    if (profile.profileType == ProfileType.ArConnect.index) {
+      _driveTxs = await _arweave.getUniqueUserDriveEntityTxs(await profile.id);
```
arielmelendez, line 54: Can we please wrap this functionality (retrieving private drives) in a function that lives in a Business Logic class? e.g. we can create an ArDriveBLL class which takes an ArweaveService as a dependency. It could have a function like `getPrivateDrivesForUserId`.

lib/blocs/profile/profile_cubit.dart
jdaev: Added [here ](https://github.com/ardriveapp/ardrive-web/pull/155/commits/ee4a4abeb3502188d2478bf8eda0044637f10c37)

lib/blocs/profile_unlock/profile_unlock_cubit.dart
jdaev: I agree, but I believe it would be better in another PR. We could look though the code to find such functions and abstract all of them instead of just doing it here for this case.

lib/blocs/profile_unlock/profile_unlock_cubit.dart
jdaev: Fixed [here](https://github.com/ardriveapp/ardrive-web/pull/155/commits/345df129e1a9c599816d0c2d5a7f6ad07e9cbb88)

lib/blocs/profile/profile_cubit.dart
```diff
@@ -69,6 +69,25 @@ class ProfileCubit extends Cubit<ProfileState> {
...
+    var isMismatch = false;
+
+    if (profile != null && profile.profileType == ProfileType.ArConnect.index) {
+      final currentPublicKey = await arconnect.getPublicKey();
+      final savedPublicKey = profile.walletPublicKey;
+      if (currentPublicKey != savedPublicKey) {
+        isMismatch = true;
+        await logoutProfile();
+      }
+    } else {
```
arielmelendez, line 84: Wouldn't this always log out non-arconnect wallets? I don't think that's what we want...

arielmelendez: Also, are there any considerations we should make for non-arconnect wallets here?

lib/blocs/profile_unlock/profile_unlock_cubit.dart
arielmelendez: Please apply the patch we worked on together today. Thanks!

lib/blocs/profile/profile_cubit.dart
jdaev: No. Only way to change json wallets is to log out, so it doesn't matter.

jdaev: Fixed [here ](https://github.com/ardriveapp/ardrive-web/pull/155/commits/c4cad29f3b6c2de7d1f088f4f7869410d118d486)

lib/blocs/profile_unlock/profile_unlock_cubit.dart
jdaev: Added [here](https://github.com/ardriveapp/ardrive-web/pull/155/commits/547f4420e76919b8603d31c4d4b60cfc65b554a0)

arielmelendez (approved):

PR #162, opened 2021-06-28
PE 20/move string literals to resource file

jdaev:
Sets up flutter localization and adds the string literals to arb files.


fedellen (dismissed):
Looks good and simple enough 👍

My only minor question is if we could make the usage slightly less verbose since we might be using it all over the place to localize in both the web app and mobile app. Is it possible with flutter/dart to shorten it from `AppLocalizations.of(context).myTextVariable` to something like a `localizedText(myTextVariable)` which handles the context in the background?

lib/main.dart
```diff
@@ -61,6 +63,15 @@ class _AppState extends State<App> {
...
             routeInformationParser: _routeInformationParser,
             routerDelegate: _routerDelegate,
+            localizationsDelegates: [
+              AppLocalizations.delegate,
+              GlobalMaterialLocalizations.delegate,
+              GlobalWidgetsLocalizations.delegate,
+            ],
+            supportedLocales: [
+              const Locale('en', ''), // English, no country code
+              const Locale('es', ''), // Spanish, no country code
```
arielmelendez, line 73: Should we have 'es' here if we don't yet have translations for it?

lib/l10n/app_en.arb
```diff
@@ -0,0 +1,87 @@
...
+  "howDoIGEtAR": "How do I get AR?",
+  "sync": "Sync",
+  "failedToSyncDrive": "Failed to sync drive contents.",
+  "tryAgain": "Try Again",
+  "driveRoot": "Drive Root",
+  "name": "Name",
+  "fileSize": "File Size",
+  "nothingToSeeHere": "There's nothing to see here.",
+  "nothingToSeeHereClickToAdd": "There's nothing to see here. Click \"new\" to add some files."
+
```
arielmelendez, line 86: rogue whitespace

lib/l10n/app_en.arb
```diff
@@ -0,0 +1,87 @@
+{
```
arielmelendez, line 1: This file is a great start. Where did you source all the strings from?

lib/main.dart
arielmelendez: I see now that we have the single translation for welcomeBack in an 'es' file. As long as everything falls back to 'en' in the absence of 'es' we should be fine.

arielmelendez: I confirmed for myself:
"The MaterialApp supportedLocales parameter limits locale changes. When the user changes the locale setting on their device, the app’s Localizations widget only follows suit if the new locale is a member of this list. If an exact match for the device locale isn’t found, then the first supported locale with a matching languageCode is used. If that fails, then the first element of the supportedLocales list is used."

Source: https://flutter.dev/docs/development/accessibility-and-localization/internationalization#specifying-the-apps-supportedlocales-parameter

arielmelendez (dismissed):
Approved with nit suggestions.

lib/l10n/app_en.arb
jdaev: I just did a search for all Text widgets and added the strings one by one.

lib/l10n/app_en.arb
jdaev: [Fixed](https://github.com/ardriveapp/ardrive-web/pull/162/commits/42a0a114e0e71b5a745dd3c549067508f5e2fdeb)

arielmelendez (approved):
LGTM

lib/l10n/app_es.arb
```diff
@@ -0,0 +1,3 @@
+{
+  "welcomeBack": "Bienvenido de nuevo"
```
agsuy, line 2: "welcomeBack": "Bienvenido de nuevo" should be "welcomeBack": "Bienvenido nuevamente"

agsuy (changes requested):
Please change the Spanish sample with the text I just provided.

agsuy (approved):

arielmelendez (approved):
LGTM

PR #163, opened 2021-07-01
PE 42/fix sync causing logout

jdaev:


lib/blocs/sync/sync_cubit.dart
```diff
@@ -45,6 +46,9 @@ class SyncCubit extends Cubit<SyncState> {
   }
 
   Future<void> startSync() async {
+    if (window.document.hidden) {
```
arielmelendez, line 49: So this would simply SKIP the sync. If we wanted to DEFER the sync, we'd need to:
a) be able to detect that the browser tab was reactivated
b) stash a state that a sync needs to happen when we defer it
c) when the tab reactivation callback is triggered, use the "needs sync" state to trigger the sync.
This depends on the requirements that we have for this feature adjustment. Let's sync up with the team in Slack and update the ticket to be

arielmelendez: Convo here: https://ardrive.slack.com/archives/C01J8FM0ZCY/p1625169376326800

arielmelendez (approved):

agsuy:
Wondering if we could reuse isCurrentProfileArConnect() somewhere as I am not sure where/how we are checking for ArConnect presence on the app (any doc on this?). Besides this I would start testing this right away.

PR #164, opened 2021-07-02
PE-45/Warn about user switching focus

jdaev:
Shows appropriate warning message based on wallet type.

### DIFF
--- a/lib/app_shell.dart
+++ b/lib/app_shell.dart
@@ -105,7 +105,7 @@ class _AppShellState extends State<AppShell> {
                                     );
                                   } else {
                                     return ProgressDialog(
-                                      title: 'Syncing, please wait',
+                                      title: 'Syncing... Please wait.',
                                     );
                                   }
                                 },
--- a/lib/app_shell.dart
+++ b/lib/app_shell.dart
@@ -90,8 +90,27 @@ class _AppShellState extends State<AppShell> {
                               color: Colors.black.withOpacity(0.5),
                             ),
                           ),
-                          ProgressDialog(
-                            title: 'Syncing, please wait',
+                          BlocBuilder<ProfileCubit, ProfileState>(
+                            builder: (context, state) {
+                              return FutureBuilder(
+                                future: context
+                                    .read<ProfileCubit>()
+                                    .isCurrentProfileArConnect(),
+                                builder: (BuildContext context,
+                                    AsyncSnapshot snapshot) {
+                                  if (snapshot.data ?? false) {
+                                    return ProgressDialog(
+                                      title:
+                                          'Syncing... Please remain on this tab.',
+                                    );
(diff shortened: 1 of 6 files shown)


lib/app_shell.dart
```diff
@@ -90,8 +90,27 @@ class _AppShellState extends State<AppShell> {
...
+                                builder: (BuildContext context,
+                                    AsyncSnapshot snapshot) {
+                                  if (snapshot.data ?? false) {
+                                    return ProgressDialog(
+                                      title:
+                                          'Syncing... Please remain on this tab.',
+                                    );
+                                  } else {
+                                    return ProgressDialog(
+                                      title: 'Syncing, please wait',
```
arielmelendez, line 108: Let's make the prompts consistent (I know this isn't in the requirements, but it should be). Let's make the copy here:
"Syncing... Please wait."

lib/pages/profile_auth/components/profile_auth_loading_screen.dart
```diff
@@ -4,12 +4,29 @@ import 'package:flutter/material.dart';
...
-        content: const Center(child: CircularProgressIndicator()),
+        content: Column(
+          crossAxisAlignment: CrossAxisAlignment.center,
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: [
+            if (isArConnect)
+              Padding(
+                padding: const EdgeInsets.all(8.0),
+                child: const Text(
+                  'Loading, please remain on this tab.',
```
arielmelendez, line 25: The copy from the requirements is:
"Loading... Please remain on this tab."

arielmelendez (changes requested):
A few copy updates need to be made.

lib/pages/profile_auth/components/profile_auth_loading_screen.dart
jdaev: [Corrected ](https://github.com/ardriveapp/ardrive-web/pull/164/commits/c9297c9d6e118e2983c398d49b53c35aff59a624)

lib/app_shell.dart
jdaev: [Corrected ](https://github.com/ardriveapp/ardrive-web/pull/164/commits/c9297c9d6e118e2983c398d49b53c35aff59a624)

arielmelendez (approved):

agsuy:
Wondering if we could reuse isCurrentProfileArConnect() somewhere as I am not sure where/how we are checking for ArConnect presence on the app (any doc on this?). Besides this I would start testing this right away.

arielmelendez (dismissed):

arielmelendez (dismissed):

agsuy (approved):
We now emit syncs to console.

Approved.

PR #165, opened 2021-07-06
task(ardrive-web): Bump version to 1.1.0 PE-48

jdaev:


web/index.html
```diff
@@ -35,7 +35,7 @@
   </script>
   <script src="pst.min.js"></script>
   <script defer src="sql-wasm.js"></script>
-  <script src="main.dart.js?version=6" type="application/javascript"></script>
+  <script src="main.dart.js?version=10" type="application/javascript"></script>
```
arielmelendez, line 38: Any risks here?

arielmelendez (approved):
Approved but please check my question about the dart version.

web/index.html
jdaev: That's not the dart version. It's the version for the web part of the app. Defining it here is necessary for Firefox to discard cached data.

PR #167, opened 2021-07-14
PE-56/Fix loading when there is no ArConnect login

jdaev:
PE-56


fedellen (approved):
Looks find from the code. `connect()` is now before the load as the commit intention states. We'll see if there any side effects from testing

agsuy (approved):

PR #169, opened 2021-07-16
Fix/PE-20 Remove Spanish Locale

jdaev:
Removes Spanish locale from supported locales files and removes app_es.arb


arielmelendez (approved):
LGTM

PR #175, opened 2021-07-21
fix(arconnect): Call disconnect only if ArConnect Login PE-79

jdaev:


lib/blocs/profile/profile_cubit.dart
```diff
@@ -150,10 +150,13 @@ class ProfileCubit extends Cubit<ProfileState> {
...
   /// Works even when the user is not authenticated.
   Future<void> logoutProfile() async {
-    try {
-      await arconnect.disconnect();
-    } catch (e) {
-      print(e);
+    final profile = await _profileDao.defaultProfile().getSingleOrNull();
+    if (profile != null && profile.profileType == ProfileType.ArConnect.index) {
+      try {
+        await arconnect.disconnect();
```
arielmelendez, line 156: It's somewhat surprising that calling arconnect.disconnect() ends up having an impact on the UI in the way that we're seeing in this ticket. What's the mechanism for that call resulting in us ending up on the ArConnect login screen?

jdaev: If you don't have any wallets or are logged out of ArConnect any call to the extension results in the login page.

arielmelendez: NUTTY. Thanks for the explanation!

arielmelendez (approved):
👍

agsuy (approved):

PR #177, opened 2021-07-22
PE-86/Resume sync after switching focus

jdaev:
Use document.visibilityState to detect if tab is hidden instead of document.hidden.

### DIFF
--- a/lib/blocs/sync/sync_cubit.dart
+++ b/lib/blocs/sync/sync_cubit.dart
@@ -48,11 +48,11 @@ class SyncCubit extends Cubit<SyncState> {
   Future<void> startSync() async {
     try {
       if (window.document.visibilityState != 'visible') {
-        print('tab hidded, skipping sync...');
+        print('Tab hidden, skipping sync...');
         emit(SyncIdle());
         return;
       }
-      print('syncing...');
+      print('Syncing...');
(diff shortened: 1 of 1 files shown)


jdaev:
> 
> 
> Do you need this in production as well?

Seems so. Don't know why, but if I use document.hidden it stops the sync when I switch back to the tab. Stops the stream entirely. It's strange because there isn't any error thrown either.

fedellen (changes requested):
See suggested change above, change 'tab hidded' to 'tab is hidden'

lib/blocs/sync/sync_cubit.dart
```diff
@@ -46,18 +46,18 @@ class SyncCubit extends Cubit<SyncState> {
...
-    print('syncing...');
-    if (await _profileCubit.logoutIfWalletMismatch()) {
-      emit(SyncWalletMismatch());
-      return;
-    }
-    emit(SyncInProgress());
-
     try {
+      if (window.document.visibilityState != 'visible') {
+        print('tab hidded, skipping sync...');
```
agsuy, line 51: typo here

jdaev (commented):
Fixed the typo

fedellen (approved):
Looks good!

arielmelendez (approved):
:+1:

PR #147, opened 2021-05-27
Task/296 attach arconnect session to wallet address

jdaev:
Shows a log out dialog when wallet is switched with arconnect.

PR #178, opened 2021-07-23
fix(ardrive-web): Show public drives without login PE-91

jdaev:


fedellen (dismissed):
Looks good!

fedellen (approved):

lib/pages/app_router_delegate.dart
```diff
@@ -67,11 +67,6 @@ class AppRouterDelegate extends RouterDelegate<AppRoutePath>
             signingIn = false;
             notifyListeners();
           }
-
-          if (state is ProfilePromptAdd) {
```
agsuy, line 71: Why are we deleting this?

jdaev: That was there to handle navigating to the correct profile page. It's been made redundant since the the correct profile page is now handled by profile_auth_page. I also checked everywhere the ProfilePromptAdd is emitted to make sure it doesn't interfere with something else. ProfilePromptAdd is emitted when the profile is null and also if the ArConnect extension is not present. So it is unlikely to cause any problems.

agsuy: So this wont interfere at all with ArConnect. Great news!

agsuy (approved):
Seems like a legacy piece of code was responsible for redirecting us to login page.
Everything else is just adding the "New button" and updating version.
Approved.

lib/pages/app_router_delegate.dart
jdaev: It was added on this commit to ensure that the profile data is cleared when ArConnect is disconnected
https://github.com/ardriveapp/ardrive-web/pull/147/commits/ec4781152ea1d60c11cca0546cd8070c1d5034ee

PR #181, opened 2021-07-27
Properly handle a missing parent folder

domob1812:
During sync, gracefully handle the situation that an entity's parent folder is missing.  This can happen temporarily while we are still retrieving data, for instance.  Without this change, the sync will abort and fail, even if the parent folder is actually there and would be added in later.

### DIFF
--- a/lib/blocs/sync/sync_cubit.dart
+++ b/lib/blocs/sync/sync_cubit.dart
@@ -435,10 +435,13 @@ class SyncCubit extends Cubit<SyncState> {
             .folderById(
                 driveId: driveId, folderId: treeRoot.folder.parentFolderId)
             .map((f) => f.path)
-            .getSingle();
+            .getSingleOrNull();
       }
 
-      await updateFolderTree(treeRoot, parentPath);
+      if (parentPath == null)
+        print('Missing parent folder: ' + treeRoot.folder.parentFolderId);
+      else
+        await updateFolderTree(treeRoot, parentPath);
     }
 
     // Update paths of files whose parent folders were not updated.


domob1812:
I have a private drive that actually runs into this issue (and works fine with the patch).  The issue there is that some folder entries failed to confirm initially while their content was confirmed, and I only later manually fixed it by recreating the missing folders (but later in the blockchain).  This drive also works fine with the CLI.

lib/blocs/sync/sync_cubit.dart
```diff
@@ -435,10 +435,13 @@ class SyncCubit extends Cubit<SyncState> {
             .folderById(
                 driveId: driveId, folderId: treeRoot.folder.parentFolderId)
             .map((f) => f.path)
-            .getSingle();
+            .getSingleOrNull();
       }
 
-      await updateFolderTree(treeRoot, parentPath);
```
arielmelendez, line 441: Was this throwing an exception before when we passed a null parentPath through? If not, then @jdaev will we need to run any kind of cleanup routine/data fix against the DB to excise folder entries with bad folder path values?

arielmelendez:
@domob1812 , thanks for this patch! We deeply appreciate contributions from the community! 🍻

lib/blocs/sync/sync_cubit.dart
domob1812: I'm not sure about this.  The exception that was causing the bug was thrown by `getSingle`.  I've not checked `updateFolderTree` to see if it would handle a `null` properly, perhaps that's the case.

What I did here seemed like the most straight-forward and least intrusive change.  But if you say that it is fine to pass `null` into `updateFolderTree`, then I'm happy to revert the second part of the change.

PR #194, opened 2021-08-19
PE 108/Migrate Bloc Test to 8.0.0

jdaev:
Migrates all the tests to bloc test and bloc version 8.0.0.
Note that some tests are incomplete and some did not work from the beginning, the scope of this PR was just to migrate the tests and get them to run.


fedellen:
Are the tests meant to be passing after this migration? I'm guessing not, but just double checking

jdaev:
> 
> 
> Are the tests meant to be passing after this migration? I'm guessing not, but just double checking

They're not. The original tests themselves were flawed and broken, one was also incomplete.

pubspec.lock
```diff
@@ -261,7 +268,7 @@ packages:
       name: file
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "6.1.2"
```
arielmelendez, line 264: Why did this roll back?

pubspec.lock
```diff
@@ -708,7 +718,7 @@ packages:
       name: process
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "4.2.3"
+    version: "4.2.1"
```
arielmelendez, line 721: Why did this roll back as well?

pubspec.lock
jdaev: Not entirely sure. Could be flutter version solving. Either way I'd suggest regenerating the lock file with the newest compatible dependency versions on 107.

pubspec.lock
jdaev: Same answer as above

arielmelendez (approved):
Looks ok minus the rollback in patch versions. If there's anything we can do about that, let's fix it. Otherwise LGTM.

PR #195, opened 2021-08-24
PE 175/Create ArConnect wallet subclass

jdaev:


lib/blocs/drive_create/drive_create_cubit.dart
```diff
@@ -65,12 +65,12 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
       final String drivePrivacy = form.control('privacy').value;
 
       final profile = _profileCubit.state as ProfileLoggedIn;
-      final walletAddress = await profile.getWalletAddress();
+      final walletAddress = await profile.wallet!.getAddress();
```
fedellen, line 68: Is asserting this as non-null always going to be safe here?

Played around in my IDE and we can type check it like this:

```dart
      // final profile = _profileCubit.state as ProfileLoggedIn;
      final wallet = profile.wallet;

      if (wallet == null) {
        return;
      }

      final walletAddress = await wallet.getAddress();
```

Also found that `final profile` was being defined twice in this part of the code

jdaev: Good catch. Thank you 👍🏽 .

lib/blocs/drive_create/drive_create_cubit.dart
```diff
@@ -46,13 +46,12 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
       return;
     }
 
+    final profile = _profileCubit.state as ProfileLoggedIn;
```
arielmelendez, line 49: Doesn't seem like this variable move was necessary.

PR #211, opened 2021-09-29
PE/454  Fix being unable to view drives without a session

jdaev:
The cause of this problem is the FolderEntry field in the FolderWithContents class being non nullable.
This field has to be nullable because there is often a delay in fetching the transactions and putting it all together.
Alternative approaches I tried to remove unwraps and nullables:

- Use watchSingleOrNull but instead of making the field nullable, wait 5 seconds and fetch again if the folder is null. This did not work since the stream doesn't update again.
- Create a wrapper type for returning the folder stream that can handle the null value, and also creating a different watchFolderStream method for attaching drives. Same problem as above where the stream would not update.


lib/blocs/fs_entry_move/fs_entry_move_cubit.dart
```diff
@@ -46,8 +46,8 @@ class FsEntryMoveCubit extends Cubit<FsEntryMoveState> {
 
   Future<void> loadParentFolder() async {
     final state = this.state as FsEntryMoveFolderLoadSuccess;
-    if (state.viewingFolder.folder.parentFolderId != null) {
-      return loadFolder(state.viewingFolder.folder.parentFolderId!);
+    if (state.viewingFolder.folder!.parentFolderId != null) {
```
arielmelendez, line 49: Please use ? rather than !

lib/blocs/fs_entry_move/fs_entry_move_cubit.dart
```diff
@@ -58,7 +58,7 @@ class FsEntryMoveCubit extends Cubit<FsEntryMoveState> {
         _driveDao.watchFolderContents(driveId, folderId: folderId).listen(
               (f) => emit(
                 FsEntryMoveFolderLoadSuccess(
-                    viewingRootFolder: f.folder.parentFolderId == null,
+                    viewingRootFolder: f.folder!.parentFolderId == null,
```
arielmelendez, line 61: Please use optional chaining ? rather than !

lib/models/daos/drive_dao/drive_dao.dart
```diff
@@ -221,7 +221,7 @@ class DriveDao extends DatabaseAccessor<Database> with _$DriveDaoMixin {
       subfolderQuery.watch(),
       filesQuery.watch(),
       (
-        FolderEntry folder,
+        FolderEntry? folder,
```
arielmelendez, line 224: May want to add a comment about why this could be optional

lib/pages/app_router_delegate.dart
```diff
@@ -120,7 +120,7 @@ class AppRouterDelegate extends RouterDelegate<AppRoutePath>
                       listener: (context, state) {
                         if (state is DriveDetailLoadSuccess) {
                           driveId = state.currentDrive.id;
-                          driveFolderId = state.currentFolder.folder.id;
+                          driveFolderId = state.currentFolder.folder?.id;
```
arielmelendez, line 123: May want to explain the meaning of a null folder ID here.

agsuy (approved):

PR #212, opened 2021-09-29
fix: grey screen on login

jdaev:


fedellen (approved):
Looks good and fixes the gray screen issue on my end

Cleans up that dangerous non-null assertion too (:

agsuy (approved):

PR #215, opened 2021-10-01
fix: filter drive entitites with owner PE-509

jdaev:
****


agsuy (approved):
Worked for public drive, pending edge cases next week

PR #216, opened 2021-10-01
fix: wallet balance parsing

jdaev:


agsuy (approved):
We show balance as '0' if its below expected threshold

PR #225, opened 2021-10-06
add tests job to github actions and repair tests PE-110

jdaev:
Adds tests job to github actions (PE-110). Also fixes the existing tests (PE-523)


agsuy:
Can not merge this until we fix Flutter Tests

jdaev:
PE-523 should have been a different PR. My mistake. If necessary I can separate these.

lib/services/crypto/keys.dart
```diff
@@ -39,7 +39,10 @@ Future<SecretKey> deriveDriveKey(
   return hkdf.deriveKey(
     secretKey: SecretKey(walletSignature),
     info: utf8.encode(password),
-    nonce: Uint8List(0),
+    // This was an empty Uint8List(0), but due to the deferring implmentations of web crypto and dart-vm
```
arielmelendez, line 42: Are the keys derived this way compatible with those derived from past and current versions of the CLI? I would be very surprised if so.

test/blocs/drive_create_cubit_test.dart
```diff
@@ -24,18 +24,16 @@ void main() {
     late DriveCreateCubit driveCreateCubit;
 
     const validDriveName = 'valid-drive-name';
+    const testGatewayURL = 'https://arweave.net';
```
arielmelendez, line 27: Does this need to connect to a real arweave gateway? Can these tests run without a network connection?

test/blocs/drive_create_cubit_test.dart
```diff
@@ -44,17 +42,18 @@ void main() {
...
       fillBytesWithSecureRandom(keyBytes);
-
       when(() => profileCubit.state).thenReturn(
         ProfileLoggedIn(
           username: 'Test',
           password: '123',
           wallet: wallet,
           walletAddress: walletAddress,
-          walletBalance: BigInt.one,
+          walletBalance: BigInt.from(20000000),
```
arielmelendez, line 51: Was this required because you started actually hitting the arweave network during tests (which we should not)?

test/utils/mocks.dart
```diff
@@ -29,3 +30,14 @@ class MockProfileCubit extends MockCubit<ProfileState> implements ProfileCubit {
...
+
+class MockProfileLoadDetails extends Mock implements ProfileLoadDetails {}
+
+class MockProfile extends Mock implements Profile {
+  @override
+  int get profileType => ProfileType.JSON.index;
+  @override
+  String get id => 'Test';
+  @override
+  String get username => 'Test';
```
arielmelendez, line 42: You might want to make these different values to reduce the chance of accidentally using them in the wrong spot and getting a false positive.

arielmelendez (changes requested):
Generally looks ok. Just please answer the few questions I posed and consider the updates I suggested. Thanks.

lib/services/crypto/keys.dart
jdaev: Yes, it's compatible tested it myself with private drives and files. However this is also the same nonce used in mine and Almog's mobile versions. Phil has confirmed it works fine on them.

test/blocs/drive_create_cubit_test.dart
jdaev: Yeah, it does. How should I change it? Add a dry run parameter?

jdaev: Actually, it wasn't required due to it hitting the network, but to get around the minimum balance limit we added before.

test/blocs/drive_create_cubit_test.dart
jdaev: It does. Fails without a network connection

arielmelendez: That's tech debt we should track for future cleanup. Our tests shouldn't rely on a live network connection.

arielmelendez (approved):

test/blocs/drive_create_cubit_test.dart
matibat: I'm not sure about the implementation, but I'd say we can accomplish this by mocking Arweave -the one that takes the gatewayUrl param-

test/models/daos/drive_dao_test.dart
```diff
@@ -26,125 +23,86 @@ void main() {
...
-
-    test('watchFolder() returns correct folder contents', () async {
-      var folderStream =
-          driveDao.watchFolderContents(driveId, folderPath: '').share();
+    test('watchFolder() with root path (' ') returns root folder', () async {
+      final folderStream =
+          driveDao.watchFolderContents(driveId, folderPath: '');
 
       await Future.wait([
         expectLater(folderStream.map((f) => f.folder!.id), emits(rootFolderId)),
```
matibat, line 46: Unrelated to the introduced changes, but: are we sure `f.folder` is not null?

matibat (commented):
Left one question & one comment

test/models/daos/drive_dao_test.dart
jdaev: Yes, folders must have an id. It's nullable because it derived from nullable tags.

jdaev:
Going to close this.
Split into https://github.com/ardriveapp/ardrive-web/pull/289 and https://github.com/ardriveapp/ardrive-web/pull/290

PR #229, opened 2021-10-15
PE 571/Single File Bundle Upload

jdaev:


web/js/tagparser.js
```diff
@@ -0,0 +1,52 @@
+var avro = window.avro;
+
+function serializeTags(tags) {
```
arielmelendez, line 3: Let's write some unit and/or integration tests for all these functions. Unit tests would be basic input/output. Integration tests might be serialize a set of tags, then deserialize them and ensure you get back the same starting set of tags.

jdaev: Added in arweave dart

test/services/tagparser_test.dart
```diff
@@ -0,0 +1,29 @@
+@TestOn('browser')
+import 'dart:html';
+
+import 'package:test/test.dart';
+
+void main() {
+  group('Serialize and Deserialze Tags With AVRO', () {
+    test('.split() splits the string on the delimiter', () {
+      print(window.toString());
```
arielmelendez, line 9: Why print?

PR #239, opened 2021-10-21
fix: Fix folder name overlap PE-22

jdaev:
UI tweaks to prevent folder and file names from overlapping with other columns of the data table.


agsuy (approved):
LGTM! 

Now the margin seems correct now.

PR #262, opened 2021-11-07
feature: csv export of file data for a drive PE-668

jdaev:
Exports the latest revisions of all files contained in a drive as a csv. Simply open the drive in the web app and click on the table icon next to the rename icon.

### DIFF
--- a/lib/blocs/data_export/data_export_cubit.dart
+++ b/lib/blocs/data_export/data_export_cubit.dart
@@ -1,7 +1,6 @@
 import 'dart:convert';
 
 import 'package:ardrive/models/models.dart';
-import 'package:ardrive/services/services.dart';
 import 'package:bloc/bloc.dart';
 import 'package:csv/csv.dart';
 import 'package:equatable/equatable.dart';
@@ -13,17 +12,15 @@ part 'data_export_state.dart';
 class DataExportCubit extends Cubit<DataExportState> {
   final String driveId;
   final DriveDao _driveDao;
-  final ArweaveService _arweave;
+  final String _gatewayURL;
 
   DataExportCubit({
     required this.driveId,
     required DriveDao driveDao,
-    required ArweaveService arweave,
+    required String gatewayURL,
   })  : _driveDao = driveDao,
-        _arweave = arweave,
-        super(DataExportStarting()) {
-    exportData();
-  }
+        _gatewayURL = gatewayURL,
+        super(DataExportInitial());
 
   Future<String> getFilesInDriveAsCSV(String driveId) async {
     final files = await _driveDao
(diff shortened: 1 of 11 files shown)


lib/pages/drive_detail/drive_detail_page.dart
```diff
@@ -2,6 +2,7 @@ import 'dart:math';
 
 import 'package:ardrive/blocs/blocs.dart';
 import 'package:ardrive/components/components.dart';
+import 'package:ardrive/components/csv_export_dialog.dart';
```
arielmelendez, line 5: Why is this needed here?

jdaev: The file that calls the dialog is a part of this file, so the import statement goes here.

arielmelendez (changes requested):

lib/models/queries/drive_queries.moor
```diff
@@ -108,6 +108,21 @@ filesInFolderAtPathWithRevisionTransactions ($order = '') AS FileWithLatestRevis
     WHERE driveId = :driveId AND path LIKE :path || '/%' AND path NOT LIKE :path || '/%/%'
     ORDER BY $order;
 
+filesInDriveithRevisionTransactions ($order = '') AS FileWithLatestRevisionTransactions:
```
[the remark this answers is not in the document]
jdaev: [Fixed](https://github.com/ardriveapp/ardrive-web/pull/262/commits/4e1c3fefa53787ffd52a6b0014858fad500d5e4e)

test/snapshots/data_export_snapshot.dart
```diff
@@ -0,0 +1,12 @@
+const String dataExportSnapshot =
```
arielmelendez, line 1: FYI there's a package mimicking Jest's snapshot testing capabilities for Flutter UIs calls `Fest`: https://pub.dev/documentation/fest/latest/
We can't use it here in the way that you can in Jest for non-UI components, but I thought you'd find it interesting.

lib/blocs/data_export/data_export_cubit.dart
```diff
@@ -55,14 +50,18 @@ class DataExportCubit extends Cubit<DataExportState> {
...
-        ..add(Uri.parse(
-                _arweave.client.api.gatewayUrl.origin + '/${file.dataTx.id}')
-            .toString());
+        ..add(file.lastModifiedDate.toString())
+        ..add(Uri.parse(_gatewayURL + '/${file.dataTx.id}').toString());
       export.add(fileContent);
     }
-    final csv = const ListToCsvConverter().convert(export);
-    final dataBytes = utf8.encode(csv) as Uint8List;
+    return const ListToCsvConverter().convert(export);
```
arielmelendez, line 57: This is a great first pass at this feature. Any idea how well it scales?

arielmelendez (approved):

lib/blocs/data_export/data_export_cubit.dart
jdaev: Well, it'll get slower the more files and folders there are, but still less than the time it takes to sync such a drive. I can improve it by adding multi-threading, but that's overkill at this point. A better improvement would be to add the exports to folders as well.

agsuy (approved):
Supports own and attached drives

PR #272, opened 2021-11-19
PE 571/Single File Bundle Upload

jdaev:
Uploads single file uploads as bundles. Packs the metadata tx and data tx of the upload as data items in a bundle.


test/services/tagparser_test.dart
```diff
@@ -0,0 +1,29 @@
+@TestOn('browser')
+import 'dart:html';
+
+import 'package:test/test.dart';
+
+void main() {
```
arielmelendez, line 6: remove

arielmelendez (changes requested):
cleanup

test/services/tagparser_test.dart
jdaev: [Done ](https://github.com/ardriveapp/ardrive-web/pull/272/commits/3eb59d9e2ae4344b47906ff9ef470de2ff59dc39)

arielmelendez (approved):

arielmelendez (approved):

PR #275, opened 2021-11-19
PE-572/Multi File Bundles

jdaev:
Packs the entityTx and dataItemTx in FileUploadHandles as DataItems into the BundleUploadHandle, then uploads that single bundle.


lib/blocs/upload/upload_cubit.dart
```diff
@@ -127,28 +131,30 @@ class UploadCubit extends Cubit<UploadState> {
...
-        .then((feePercentage) =>
-            // Workaround [BigInt] percentage division problems
-            // by first multiplying by the percentage * 100 and then dividing by 100.
-            uploadCost * BigInt.from(feePercentage * 100) ~/ BigInt.from(100))
-        .catchError((_) => BigInt.zero,
-            test: (err) => err is UnimplementedError);
-
-    final minimumPstTip = BigInt.from(10000000);
-    pstFee = pstFee > minimumPstTip ? pstFee : minimumPstTip;
+    final dataItemsCost = _dataItemUploadHandles.isNotEmpty
```
jdaev, line 145: We get the price directly from arweave for the combined size of all data items. However this may not be accurate since these data items themselves can be split across multiple bundles. Solution to this, move the packing of the data items here and calculate a more accurate tip, but keeping the packing of the data items to just before the upload is more efficient and gives the user more information on what file is being uploaded.

lib/blocs/upload/upload_handle.dart
```diff
@@ -0,0 +1,12 @@
+abstract class UploadHandle {
+  
+  BigInt? get cost;
+
+  /// The size of the file before it was encoded/encrypted for upload.
+  int? get size;
```
matibat, line 6: Could we make these non-nullable? Is there any impediment?

lib/blocs/upload/file_upload_handle.dart
```diff
@@ -1,58 +1,52 @@
...
-      return bundleTx!.reward;
-    } else {
-      return (entityTx as Transaction).reward + (dataTx as Transaction).reward;
-    }
+    return (entityTx as Transaction).reward + (dataTx as Transaction).reward;
   }
 
   /// The size of the file before it was encoded/encrypted for upload.
+  @override
   int? get size => entity.size;
```
matibat, line 17: Is `entity.size` nullable?

lib/blocs/upload/bundle_upload_handle.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:ardrive/blocs/upload/upload_handle.dart';
+import 'package:ardrive/services/services.dart';
+import 'package:arweave/arweave.dart';
+
+class BundleUploadHandle implements UploadHandle {
```
arielmelendez, line 5: What would you describe as the key utility for this class? It's not clear at first reading.

lib/blocs/upload/upload_cubit.dart
```diff
@@ -182,24 +188,110 @@ class UploadCubit extends Cubit<UploadState> {
...
       ),
     );
   }
 
+  Future<BigInt> getPSTFee(BigInt uploadCost) async {
+    var pstFee = await _pst
+        .getPstFeePercentage()
+        .then((feePercentage) =>
+            // Workaround [BigInt] percentage division problems
+            // by first multiplying by the percentage * 100 and then dividing by 100.
```
matibat, line 203: Let's put the reason here: `// Workaround to low precision in the product of a tiny number` or similar

lib/blocs/upload/bundle_upload_handle.dart
jdaev: It's mostly for showing upload progress in the UI

arielmelendez: Would an easier, functional implementation to use implementation just be a function that takes the necessary inputs and then gives back a stream of tuples of data like (bytesUploaded, totalBytes)?

jdaev: These classes are also now used to encapsulate relevant data, so I'm not sure if that approach would work.

lib/blocs/upload/upload_handle.dart
jdaev: We can if we provide 0 as the default. But it's here because the null value has meaning in this case.

lib/blocs/upload/file_upload_handle.dart
jdaev: Yes

lib/blocs/upload/upload_cubit.dart
jdaev: Not just because of low precision. More of a typing issue, dividing BigInt by Bigint gives a double which is not what we want. Will update this

lib/utils/bundles/next_fit_bundle_packer.dart
```diff
@@ -0,0 +1,41 @@
...
+        bundleItems = [];
+      }
+      bundleItems.add(item);
+    }
+    if (bundleItems.isNotEmpty) {
+      bundles.add(bundleItems);
+    }
+    return bundles;
+  }
+}
```
fedellen, line 41: Nice! This really encapsulates some simple underlying logic here. Seems to be the simplest implementation of packing for now, just checking to see if the next item fits which is 

I'll absolutely reference this over in Core for multi-file

lib/blocs/upload/upload_cubit.dart
```diff
@@ -184,24 +190,128 @@ class UploadCubit extends Cubit<UploadState> {
...
+            fileEntity.toRevisionCompanion(
+              performedAction: !conflictingFiles.containsKey(fileEntity.name)
+                  ? RevisionAction.create
+                  : RevisionAction.uploadNewVersion,
+            ),
+          );
+        }
+        final dataItems = <DataItem>[];
+        var dataItemTotalSize = 0;
+        for (var uploadHandle in uploadHandles) {
```
fedellen, line 278: Why are we looping into these twice instead of handling this within the same loop?

lib/blocs/upload/upload_cubit.dart
```diff
@@ -184,24 +190,128 @@ class UploadCubit extends Cubit<UploadState> {
...
+          await _driveDao.writeFileEntity(fileEntity, uploadHandle.path);
+          await _driveDao.insertFileRevision(
+            fileEntity.toRevisionCompanion(
+              performedAction: !conflictingFiles.containsKey(fileEntity.name)
+                  ? RevisionAction.create
+                  : RevisionAction.uploadNewVersion,
+            ),
+          );
+        }
+        final dataItems = <DataItem>[];
```
fedellen, line 276: This var doesn't seem to be used after being assigned, unless I'm missing something passed internally within `DataBundle()`

lib/blocs/upload/upload_cubit.dart
```diff
@@ -184,24 +190,128 @@ class UploadCubit extends Cubit<UploadState> {
...
+        var dataItemTotalSize = 0;
+        for (var uploadHandle in uploadHandles) {
+          final dataDataItem = (uploadHandle.dataTx as DataItem);
+          final entityDataItem = (uploadHandle.dataTx as DataItem);
+
+          assert(uploadHandle.entity.dataTxId == dataDataItem.id);
+
+          dataItemTotalSize += uploadHandle.entity.size!;
```
fedellen, line 284: The total size here won't account for the metadata data item size, if that is important below.

fedellen (approved):

lib/blocs/upload/upload_cubit.dart
jdaev: This is just for showing the progress in the UI. It's consistent with what we do for normal transactions

lib/blocs/upload/upload_cubit.dart
jdaev: Removed. Thanks for catching that

lib/blocs/upload/upload_cubit.dart
jdaev: Cleaned it up. Thanks for the review

lib/utils/bundles/next_fit_bundle_packer.dart
jdaev: Credits to Ariel for helping me with separating it out

lib/utils/bundles/sized_item.dart
```diff
@@ -0,0 +1,5 @@
+class SizedItem {
```
matibat, line 1: This is a new class, it's not instantiated at any place in the diff and we need the `size` getter to be implemented every time it's extended. Don't we want `SizedItem` to be `abstract`?

matibat (commented):
I left one single comment. It's looking really good!

lib/utils/bundles/sized_item.dart
jdaev: Good point. Will change it, thanks

lib/blocs/upload/multi_file_upload_handle.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:ardrive/blocs/upload/upload_handle.dart';
+import 'package:ardrive/entities/file_entity.dart';
+import 'package:ardrive/services/services.dart';
+import 'package:arweave/arweave.dart';
+
+class MultiFileUploadHandle implements UploadHandle {
+  final Transaction bundleTx;
+  List<FileEntity> fileEntities = [];
```
arielmelendez, line 8: Is this used for display purposes? It seems to have no internal use. Is there a reason it's mutable? Seems like it should be final.

lib/services/arweave/arweave_service.dart
```diff
@@ -40,6 +40,10 @@ class ArweaveService {
     _mempoolSize = await getMempoolAverage();
   }
 
+  Future<BigInt> getPrice({required int byteSize}) {
+    return client.api.get('/price/$byteSize').then((res) => BigInt.parse(res.body));
```
arielmelendez, line 44: Since we're dealing with bulk operations and will likely have many repeated calls to this function with the same byteSize value (e.g. metadata files), consider using a bounded LRU cache with TTL, e.g. https://pub.dev/packages/stash. One issue I ran into in the ArFSEntityCache PR is that when I'd reach for the same item back to back, the second fetch might happen while the first is in flight and before it has been cached. My workaround for that was to actually cache Promises so that multiple awaiters could wait on it and share the result as well. You could try something like that here as well. Another performance tweak you can make: since we know the new Arweave 2.5 pricing algorithm, you can just fetch the 1 chunk reward price (i.e. for 256 * 1024 bytes) and simply return that for any byte count less than it (but greater than 0, which itself is the special, base reward case that you should cache and handle separately). You can do this in a follow-up PR. But I'd recommend doing it soon.

lib/blocs/upload/upload_state.dart
```diff
@@ -52,6 +54,7 @@ class UploadReady extends UploadState {
   final bool uploadIsPublic;
 
   final List<FileUploadHandle> files;
+  final List<MultiFileUploadHandle> bundles;
```
arielmelendez, line 57: Is there ever a case where there would be files AND bundles? If not, could we make this class a generic and make this a List<T> where T is derived from UploadState<T extends UploadHandle>?

jdaev: Yeah. You can still upload files as V2 transactions if over the bundle size limit.

arielmelendez (approved):

lib/blocs/upload/multi_file_upload_handle.dart
jdaev: [Changed](https://github.com/ardriveapp/ardrive-web/pull/275/commits/b9efd60c9153891c9b0f466dad009a1f2948929f)

matibat (approved):
It looks good to me :clap:

arielmelendez (approved):

PR #279, opened 2021-11-30
task: remove tc agreement for returning users PE-734

jdaev:


lib/pages/profile_auth/components/profile_auth_add_screen.dart
```diff
@@ -89,42 +89,44 @@ class ProfileAuthAddScreen extends StatelessWidget {
...
                               ),
-                            ),
-                          ],
-                        ),
+                            ],
+                          ),
+                        ],
                         const SizedBox(height: 32),
                         SizedBox(
                           width: double.infinity,
```
matibat, line 132: I'm not sure to what is `SizedBox` related to, but is `double.infinity` safe here?

lib/pages/profile_auth/components/profile_auth_add_screen.dart
```diff
@@ -72,7 +72,7 @@ class ProfileAuthAddScreen extends StatelessWidget {
                               context.read<ProfileAddCubit>().submit(),
                           validationMessages: (_) => kValidationMessages,
                         ),
-                        if (!state.isExistingUser) ...{
+                        if (!state.isExistingUser) ...[
```
matibat, line 75: I'd guess this is an inline conditional, right? In the else case there's nothing to push to this `List` here I assume

matibat (commented):
Left a few questions there

lib/pages/profile_auth/components/profile_auth_add_screen.dart
jdaev: Yes. It's the value typically used for denoting MAX in flutter

lib/pages/profile_auth/components/profile_auth_add_screen.dart
jdaev: It's a language feature.
https://dart.dev/guides/language/language-tour#collection-operators

fedellen (approved):
👍

PR #281, opened 2021-12-02
fix: use correct app version in version tag PE-748

jdaev:


agsuy (dismissed):
LGTM!

arielmelendez (dismissed):

lib/services/arweave/arweave_service.dart
```diff
@@ -468,8 +469,11 @@ class ArweaveService {
   Future<Transaction> prepareDataBundleTx(
       DataBundle bundle, Wallet wallet) async {
     final bundleBlob = await bundle.asBlob();
+    final packageInfo = await PackageInfo.fromPlatform();
```
fedellen, line 472: Is there a reason why we can't handle this call  within `addApplicationTags`? It seems like all of the implementations similarly call it before adding the tags.

Is it to keep addApplicationTags as a synchronous call or some scope limitation?

fedellen (approved):
Left one minor question about reducing duplication of fromPlatform call, but its looking good!

lib/services/arweave/arweave_service.dart
jdaev: It's to keep it synchronous because addApplicationTags is usually called using the spread operator (tx..addApplicationTags())

PR #282, opened 2021-12-02
fix: correct private file size limit PE-746

jdaev:


lib/blocs/upload/upload_cubit.dart
```diff
@@ -43,7 +43,7 @@ class UploadCubit extends Cubit<UploadState> {
   Transaction? feeTx;
 
   final bundleSizeLimit = 503316480;
-
+  final privateFileSizeLimit = 104858;
```
fedellen, line 46: 102.4 KiB private file limit? Is this correct here, or is this not measured in bytes?

jdaev: Missed some zeros. Thanks for catching that.

fedellen (approved):

PR #292, opened 2021-12-09
PE 770/Create test for pst service

jdaev:


.github/workflows/deploy-dev.yaml
```diff
@@ -21,15 +19,19 @@ jobs:
...
+        run: |
           flutter config --enable-web
           flutter build web --profile
       - uses: JamesIves/github-pages-deploy-action@4.1.1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           branch: gh-pages
           folder: build/web
-          clean: true
+          clean: true
```
arielmelendez, line 37: @jdaev what plugin are you using that trims trailing newlines from the end of certain file types? It's... maddening 😆

jdaev: Name: YAML
Id: redhat.vscode-yaml
Description: YAML Language Support by Red Hat, with built-in Kubernetes syntax support
Version: 1.2.2
Publisher: Red Hat
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml

Should I disable formatting and push again?

.github/workflows/pr.yaml
```diff
@@ -24,8 +24,24 @@ jobs:
         run: |
           flutter config --enable-web
           flutter build web --profile
+      - uses: nanasess/setup-chromedriver@v1
+        with:
+          # Add latest version from https://chromedriver.chromium.org/downloads
+          chromedriver-version: '96.0.4664.45'
```
arielmelendez, line 30: What's the maintenance implication of this?

jdaev: We'd have to up it every new update. But this script isn't working at the moment. @agsuy is working on it.

jdaev: Best if we remove this script from here and follow up in another PR. This test does run locally though. 
You just need to start chromedriver on port 4444 and run the following command.
`flutter drive \
          --driver=test_driver/integration_test.dart \
          --target=integration_test/pst_service_test.dart \
          -d chrome`

PR #300, opened 2021-12-21
PE 809 Dont sync while user is interacting

jdaev:
Coverts the existing congestion warning wrapper into a user interaction wrapper that handles showing congestion warnings and sets the isPerformingAction flags while the action is being performed.


lib/components/upload_form.dart
```diff
@@ -16,31 +16,33 @@ Future<void> promptToUploadFile(
...
-  profleCubit.setOverlayOpen(true);
-  final selectedFiles = allowSelectMultiple
-      ? await file_selector.openFiles()
-      : [await file_selector.openFile()].where((file) => file != null)
-          as List<file_selector.XFile>;
-  profleCubit.setOverlayOpen(false);
-  if (selectedFiles.isEmpty) {
-    return;
-  }
+  profleCubit.performAction(
```
arielmelendez, line 19: I like the usability improvement of using 1 function to both launch the modal AND set the correct states, rather than the previous implementation which would have allowed callers to call one and forget to call the other. That said, `performAction` is too vague a name and doesn't really convey the intent that we have for the function, that being "issuing an uninterruptible modal experience over the app's primary UI". Could we rename the methods and variables around that concept and perhaps even limit the function type that is passed to the profileCubit even further? Right now, ANY workload can be passed to it, so it could be misused/abused in the future by a well-intentioned but context-missing actor.

jdaev: I can restrict it to only accept a dialog. Any suggestion for the name?

lib/pages/user_interaction_wrapper.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:ardrive/blocs/blocs.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+
+import 'congestion_warning_wrapper.dart';
+
+Future<void> performUserAction(
+  BuildContext context,
+  Function() showDialog, {
+  bool warnAboutCongestion = true,
```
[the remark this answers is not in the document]
jdaev: The ones inside the braces are named parameters that can be assigned a default value.
Congestion has emphasis because we show that before most dialogs. Any suggestion for improving this?

lib/components/upload_form.dart
matibat: Maybe...
 - Rename `performAction` to `performUninterruptibleAction` or `openUninterruptibleModal`,
 - the `action` argument could be typed to match "a callback that opens a modal" -if possible with the dart types-
 - `_isPerformingAction` coud be `_doNotInterrupt` or `_doNotSync`

lib/pages/user_interaction_wrapper.dart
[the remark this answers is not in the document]
matibat: So by default, it runs the congestion check and we can set it to false to avoid that behavior in some specific cases by using the named parameter `warnAboutCongestion`. I think that's fine

lib/pages/user_interaction_wrapper.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:ardrive/blocs/blocs.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+
+import 'congestion_warning_wrapper.dart';
+
+Future<void> showUninterruptibleDialog(
+  BuildContext context,
+  Future Function() showDialog, {
+  bool warnAboutCongestion = true,
```
arielmelendez, line 10: I think this function is too generic to be fed the responsibility of showing a congestion warning. I think you should create a different wrapper function that handles showing congestion warnings if necessary before creating the modal and use that one in the places where you would pass `warnAboutCongestion: true`.

lib/pages/user_interaction_wrapper.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:ardrive/blocs/blocs.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+
+import 'congestion_warning_wrapper.dart';
+
+Future<void> showUninterruptibleDialog(
```
arielmelendez, line 7: I'm not crazy about where we ended up with the name here. `Modal` is an accurate enough description to convey uninterruptibility. If we end up making this the defacto function responsible for showing a congestion-dependent modal, I'd name it something like `showCongestionDependentModalDialog` and update `showDialog` to `showModalDialog`.

lib/blocs/profile/profile_cubit.dart
```diff
@@ -21,7 +21,7 @@ class ProfileCubit extends Cubit<ProfileState> {
   final ProfileDao _profileDao;
   final Database _db;
 
-  bool _isOverlayOpen = false;
+  bool _doNotInterrupt = false;
```
arielmelendez, line 24: Along with the theme of `Modal` conveying "uninterruptibility" sufficiently, I'd say that an appropriate name here would be `_isModalPresented`.

lib/blocs/activity/activity_cubit.dart
```diff
@@ -0,0 +1,15 @@
...
+import 'package:equatable/equatable.dart';
+import 'package:meta/meta.dart';
+
+part 'activity_state.dart';
+
+class ActivityCubit extends Cubit<ActivityState> {
+  ActivityCubit() : super(ActivityNotRunning());
+
+  void performUninterruptableActivity(Function activity) async {
+    emit(ActivityInProgress());
```
arielmelendez, line 11: In the absence of something like a semaphore here to count the number of things in progress, can we please throw an error here if the cubit is already in progress when this function is called?

arielmelendez (approved):

PR #301, opened 2021-12-21
PE 810 Add lastSync var

jdaev:
Only sync after context switch if time elapsed since last sync is greater than 5 minutes.

### DIFF
--- a/lib/blocs/sync/sync_cubit.dart
+++ b/lib/blocs/sync/sync_cubit.dart
@@ -18,6 +18,7 @@ import '../blocs.dart';
 part 'sync_state.dart';
 
 const kRequiredTxConfirmationCount = 15;
+const kSyncTimerDuration = 5;
 
 /// The [SyncCubit] periodically syncs the user's owned and attached drives and their contents.
 /// It also checks the status of unconfirmed transactions made by revisions.
@@ -28,6 +29,7 @@ class SyncCubit extends Cubit<SyncState> {
   final Database _db;
 
   StreamSubscription? _syncSub;
+  DateTime? _lastSync;
 
   SyncCubit({
     required ProfileCubit profileCubit,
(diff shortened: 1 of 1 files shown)


lib/blocs/sync/sync_cubit.dart
```diff
@@ -46,16 +48,22 @@ class SyncCubit extends Cubit<SyncState> {
...
         .listen((_) {});
     startSync();
   }
 
   void restartSyncOnFocus() {
-    whenBrowserTabIsUnhidden(() => Future.delayed(Duration(seconds: 2))
-        .then((value) => createSyncStream()));
+    whenBrowserTabIsUnhidden(() {
+      if (_lastSync != null &&
+          DateTime.now().difference(_lastSync!).inMinutes <
```
arielmelendez, line 61: Surprising that Dart makes you force unwrap _lastSync after having just checked its nullness. Is it definitely necessary?

jdaev: Yeah, it's like that unless I separate it into two if conditions

lib/blocs/sync/sync_cubit.dart
```diff
@@ -46,16 +48,22 @@ class SyncCubit extends Cubit<SyncState> {
...
+              kSyncTimerDuration) {
+        return;
+      }
+      Future.delayed(Duration(seconds: 2)).then((value) => createSyncStream());
```
arielmelendez, line 65: I know this was here before but what's the significance of the 2 seconds? Sounds like a hack. And either way it should be documented here.

arielmelendez (approved):

lib/blocs/sync/sync_cubit.dart
jdaev: Just a slight delay so that it doesn't automatically sync after tab switch.

jdaev: It's an arbitrary value

PR #303, opened 2021-12-28
feat(ui): add table pagination PE-847

jdaev:
This PR adds pagination to the data table used for the desktop drive detail view. To meet the requirements for this ticket, it was necessary to copy and modify the source for the PaginatedDataTable flutter provides. It wasn't possible to do this in a more OOP way because I needed to take out a widget (Card) from the original hierarchy and add support for pages.

By default PaginatedDataTable doesn't have any concept of a page instead only showing a specific view of the list of rows. So, paging had to be added by calculating the current page and total page count from the current row, total rows and rows per page parameters from the PaginatedDataTableSource class.


lib/pages/drive_detail/components/drive_detail_data_table_source.dart
```diff
@@ -0,0 +1,232 @@
...
+        context: context,
+        folder: folder.folder,
+        onPressed: folder.onPressed,
+        index: index,
+        selected: folder.selected,
+      );
+    } else {
+      final file = files[folders.length > files.length
+          ? folders.length - index
+          : index - folders.length];
```
matibat, line 30: Given that at this point `index >= folders.length`, is there any issue by doing the following?
```suggestion
      int fileIndex = index - folders.length;
      final file = files[fileIndex];
```

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
```diff
@@ -0,0 +1,232 @@
...
+class DriveTableFile {
+  final FileWithLatestRevisionTransactions file;
+  final bool selected;
+  final Function onPressed;
+
+  DriveTableFile({
+    required this.file,
+    required this.selected,
+    required this.onPressed,
+  });
```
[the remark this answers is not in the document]
matibat: Q: Is it possible to organize this data in an interface instead of a class?

matibat (commented):
Nice changes overall! I left a couple of comments

matibat:
Also, would you take care of the merging conflicts with the base branch?

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -77,36 +79,74 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
...
+          var currentFolder;
+
+          var rowsPerPage;
+          var availableRowsPerPage;
+
+          if (folderContents.folder != null) {
+            currentFolder = folderContents;
+          }
+          final itemCount =
+              currentFolder.files.length + currentFolder.subfolders.length;
```
arielmelendez, line 97: Why isn't the linter warning about potentially uninitialized access here? `currentFolder` could be null if `folderContents.folder` is null. I'd avoid `var` without a type and uninitialized variables like the plague if possible.

arielmelendez: Using the type explicitly seems to get the static analyzer/compiler wise to the situation:
image

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -77,36 +79,74 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
...
-        if (state != null) {
-          emit(
-            state.copyWith(
+        } else {
+          final state = this.state is DriveDetailLoadSuccess
+              ? this.state as DriveDetailLoadSuccess
+              : null;
+          final profile = _profileCubit.state;
+
+          var currentFolder;
```
arielmelendez, line 88: I'd like to see as much of this region of code rewritten as a pure input/output, unit-testable/tested function as possible.

arielmelendez: Seems like:
inputs = item count, default rows per page
outputs = rowsPerPage, availableRowsPerPage

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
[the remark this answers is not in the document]
jdaev: Yeah, it's a constructor.
I could add an interface that abstracts away what's common with files and folders but I don't think it's really necessary, since files and folders are the only things we have.

lib/pages/drive_detail/components/custom_paginated_data_table.dart
```diff
@@ -0,0 +1,476 @@
...
+    for (var index = firstRowIndex; index < nextPageFirstRowIndex; index += 1) {
+      DataRow? row;
+      if (index < _rowCount || _rowCountApproximate) {
+        row = _rows.putIfAbsent(index, () => widget.source.getRow(index));
+        if (row == null && !haveProgressIndicator) {
+          row ??= _getProgressIndicatorRowFor(index);
+          haveProgressIndicator = true;
+        }
+      }
+      //row ??= _getBlankRowFor(index);
```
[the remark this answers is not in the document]
jdaev: [Removed](https://github.com/ardriveapp/ardrive-web/pull/303/commits/07a20d6e791cec419b4edf45b9a19260948f0e7d)

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -77,36 +79,74 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
...
+              availableRowsPerPage: availableRowsPerPage,
+            ));
+          }
         }
       },
     ).listen((_) {});
   }
 
+  void setRowsPerPage(int rowsPerPage) {
+    final state = this.state is DriveDetailLoadSuccess
```
arielmelendez, line 138: Saves you one line of code and a few branches:
```
    switch (state.runtimeType) {
      case DriveDetailLoadSuccess:
        final state = this.state as DriveDetailLoadSuccess;
        emit(
          state.copyWith(
            rowsPerPage: rowsPerPage,
          ),
        );
    }
```

lib/blocs/drive_detail/drive_detail_state.dart
```diff
@@ -26,12 +26,17 @@ class DriveDetailLoadSuccess extends DriveDetailState {
   /// Null if no file is selected.
   final Uri? selectedFilePreviewUrl;
 
+  final int rowsPerPage;
```
arielmelendez, line 29: I don't find these names meaningfully different at first glance. Either add comments explaining them, or update the names.

arielmelendez: Maybe `rowsPerPageChoices`? I'm not sure that's less confusing than `available`. I can see the naming trickiness here.

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
```diff
@@ -0,0 +1,232 @@
...
+  final List<DriveTableFile> files;
+  final BuildContext context;
+  DriveDetailDataTableSource(
+      {required this.folders, required this.files, required this.context});
+
+  @override
+  DataRow? getRow(int index) {
+    assert(index >= 0);
+
+    if (index >= (folders.length + files.length)) {
```
arielmelendez, line 14: If `(folders.length + files.length)` has a special meaning, consider making an accessor for it so the reader sees something like:
```
    if (index >= totalNumEntries()) {
```

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
```diff
@@ -0,0 +1,232 @@
...
+      return null;
+    }
+
+    if (index < folders.length) {
```
arielmelendez, line 18: Is this really `if (isFolderIndex(index)) {`?

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
```diff
@@ -0,0 +1,232 @@
...
+      final folder = folders[index];
+      return _buildFolderRow(
+        context: context,
+        folder: folder.folder,
+        onPressed: folder.onPressed,
+        index: index,
+        selected: folder.selected,
+      );
+    } else {
+      final file = files[folders.length > files.length
```
arielmelendez, line 28: Not super reader friendly here

arielmelendez (changes requested):
A bit more to review but taking a break for now.

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/303/commits/0185ac6325f84a2e4c31737d183a06cff4fc1d48)

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/303/commits/0185ac6325f84a2e4c31737d183a06cff4fc1d48)

lib/blocs/drive_detail/drive_detail_cubit.dart
jdaev: Thank you. [Updated](https://github.com/ardriveapp/ardrive-web/pull/303/commits/0185ac6325f84a2e4c31737d183a06cff4fc1d48)

lib/blocs/drive_detail/drive_detail_cubit.dart
jdaev: Thanks. [Updated it](https://github.com/ardriveapp/ardrive-web/pull/303/commits/0185ac6325f84a2e4c31737d183a06cff4fc1d48)

lib/blocs/drive_detail/drive_detail_state.dart
jdaev: rowsPerPage and availableRowsPerPage are whats used in the default flutter widgets, which is why I went with it.

lib/blocs/drive_detail/drive_detail_cubit.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/303/commits/e8b5e45405d2e1845ff5311ac3099ac9f77bd6a2)

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -77,36 +78,73 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
...
+              rowsPerPage: availableRowsPerPage.first,
+              availableRowsPerPage: availableRowsPerPage,
+            ));
+          }
         }
       },
     ).listen((_) {});
   }
 
+  List<int> calculateRowsPerPage(int totalEntries) {
```
arielmelendez, line 127: `calculateAvailbleRowsPerPage` might be a better name. And making it a static function rather than a member function and passing in the defaults would make it more unit testable.

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
arielmelendez: Then it might make sense to make that function. Your code is making a hard assumption that ordering in the list is based on folders, then files. This will break in the future if we allow users to sort differently. Then, in the future, you'd just change the logic of `isFolderIndex` to account for sort algorithm and this would "just work".

arielmelendez (dismissed):
Some missed opportunities for unit testing. Approving nonetheless.

fedellen (approved):

matibat (approved):
It looks good overall :+1:

matibat (dismissed):
LGTM

arielmelendez (dismissed):

arielmelendez (approved):

PR #307, opened 2022-01-03
PE 765/Add bundledIn to files

jdaev:
Adds a bundledIn field to file entities, file entries and file revisions.
UI now shows bundle tx id for files in the info side sheet. 
Also adds the bundled tx id to the file entity on db after the bundle is prepared and signed.


lib/pages/drive_detail/components/fs_entry_side_sheet.dart
```diff
@@ -201,6 +201,15 @@ class FsEntrySideSheet extends StatelessWidget {
                           ),
                         ),
                       ]),
+                      DataRow(cells: [
```
matibat, line 204: Shouldn't this field be hidden when there's no bundle tx (aka non-bundled entities)?
I think it will render a row with the clipboard icon, but it will copy `null` to the clipboard

matibat (commented):
Left one single comment there

lib/pages/drive_detail/components/fs_entry_side_sheet.dart
jdaev: [Corrected](https://github.com/ardriveapp/ardrive-web/pull/307/commits/e2b98885a976430df944ac047c493181001f88c5)

matibat (dismissed):
It looks good to me now

lib/blocs/upload/upload_cubit.dart
```diff
@@ -264,15 +264,38 @@ class UploadCubit extends Cubit<UploadState> {
...
         maxDataItemCount: maxFilesPerBundle,
       ).packItems(_dataItemUploadHandles.values.toList());
+
       //Create bundles
       for (var uploadHandles in bundleItems) {
+        final dataBundle = DataBundle(
+          items: uploadHandles
+              .map((e) => e.asDataItems().toList())
+              .reduce((value, element) => value += element)
+              .toList(),
```
arielmelendez, line 274: Is toList() redundant?

arielmelendez (changes requested):
Missing schema version update

lib/blocs/upload/upload_cubit.dart
jdaev: Nope. Reduce returns an Iterable which has to be converted to a List

lib/models/database/database.dart
```diff
@@ -25,7 +24,7 @@ class Database extends _$Database {
         },
         onUpgrade: (Migrator m, int from, int to) async {
           print('schema changed from $from to $to ');
-          if (from >= 1 && from <= 9) {
+          if (from >= 1 && from <= 10) {
```
matibat, line 27: If we want that number to be the previous version, then let's define a `prevSchemaVersion` so we reduce redundancy
_Please ensure the syntax I used is fine before committing, or edit in your IDE_
```suggestion
          int prevSchemaVersion = schemaVersion - 1;
          if (from >= 1 && from <= prevSchemaVersion) {
```

matibat: We can also use the strict less-than, so it never reaches the current version
```suggestion
          if (from >= 1 && from < schemaVersion) {
```

matibat (dismissed):
Left one minor suggestion. It seems nice to go!

lib/blocs/upload/upload_cubit.dart
arielmelendez: Hmm. When i tested locally the static analyzer/compiler didn't seem to care.

arielmelendez (approved):

lib/models/database/database.dart
jdaev: Will add this when I finalize the db version. There's multiple PRs right now that require it.

agsuy (approved):
Merge of 766 fix

PR #315, opened 2022-01-05
fix(arconnect): fix wallet switch not closing dialogs PE-867

jdaev:
Pops an extra context when activity is in progress.

Also using this PR to organize and cleanup imports.

### DIFF
--- a/lib/blocs/profile/profile_cubit.dart
+++ b/lib/blocs/profile/profile_cubit.dart
@@ -4,6 +4,7 @@ import 'package:ardrive/entities/profileTypes.dart';
 import 'package:ardrive/models/models.dart';
 import 'package:ardrive/services/arconnect/arconnect_wallet.dart';
 import 'package:ardrive/services/services.dart';
+import 'package:ardrive/utils/html/html_util.dart';
 import 'package:arweave/arweave.dart';
 import 'package:bloc/bloc.dart';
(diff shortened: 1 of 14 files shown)


lib/components/wallet_switch_dialog.dart
```diff
@@ -24,6 +23,9 @@ class WalletSwitchDialog extends StatelessWidget {
         actions: [
           TextButton(
             onPressed: () {
+              if (context.read<ActivityCubit>().state is ActivityInProgress) {
+                Navigator.pop(context);
```
matibat, line 27: Correct me if I'm wrong: the 1st pop is to close the `Wallet switch` modal after clicking `Logout`, the 2nd is for closing the activity in progress if any

lib/blocs/profile/profile_cubit.dart
```diff
@@ -106,6 +106,7 @@ class ProfileCubit extends Cubit<ProfileState> {
     final isMismatch = await checkIfWalletMismatch();
     if (isMismatch) {
       await logoutProfile();
+      triggerHTMLPageReload();
```
matibat, line 109: This line then triggers a reload after the user has clicked the `Logout` button, right?

matibat (approved):
Left a couple of questions, but it looks good to go

PR #316, opened 2022-01-05
PE 766 Add bundled in for drives

jdaev:
Adds a bundledIn field for drive and shows drive tx id, root folder tx id and bundle tx id in info sheet.

### DIFF
--- a/lib/blocs/fs_entry_info/fs_entry_info_cubit.dart
+++ b/lib/blocs/fs_entry_info/fs_entry_info_cubit.dart
@@ -52,16 +52,21 @@ class FsEntryInfoCubit extends Cubit<FsEntryInfoState> {
           );
     } else {
       _entrySubscription =
-          _driveDao.driveById(driveId: driveId).watchSingle().listen(
-                (d) => emit(
-                  FsEntryInfoSuccess<Drive>(
-                    name: d.name,
-                    lastUpdated: d.lastUpdated,
-                    dateCreated: d.dateCreated,
-                    entry: d,
-                  ),
-                ),
-              );
+          _driveDao.driveById(driveId: driveId).watchSingle().listen((d) async {
+        final rootFolder = await _driveDao
+            .latestFolderRevisionByFolderId(
+                folderId: d.rootFolderId, driveId: d.id)
+            .getSingle();
+        emit(
+          FsEntryDriveInfoSuccess(
+            name: d.name,
+            lastUpdated: d.lastUpdated,
+            dateCreated: d.dateCreated,
+            drive: d,
+            rootFolder: rootFolder,
+          ),
+        );
+      });
     }
   }
 
--- a/lib/blocs/fs_entry_info/fs_entry_info_cubit.dart
+++ b/lib/blocs/fs_entry_info/fs_entry_info_cubit.dart
@@ -51,22 +51,30 @@ class FsEntryInfoCubit extends Cubit<FsEntryInfoState> {
             ),
           );
     } else {
-      _entrySubscription =
-          _driveDao.driveById(driveId: driveId).watchSingle().listen((d) async {
-        final rootFolder = await _driveDao
-            .latestFolderRevisionByFolderId(
-                folderId: d.rootFolderId, driveId: d.id)
-            .getSingle();
-        emit(
-          FsEntryDriveInfoSuccess(
-            name: d.name,
-            lastUpdated: d.lastUpdated,
-            dateCreated: d.dateCreated,
-            drive: d,
-            rootFolder: rootFolder,
-          ),
-        );
-      });
+      _entrySubscription = _driveDao
+          .driveById(
+            driveId: driveId,
+          )
+          .watchSingle()
+          .listen(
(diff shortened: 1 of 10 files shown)


lib/blocs/fs_entry_info/fs_entry_info_cubit.dart
```diff
@@ -52,16 +52,21 @@ class FsEntryInfoCubit extends Cubit<FsEntryInfoState> {
...
-                  FsEntryInfoSuccess<Drive>(
-                    name: d.name,
-                    lastUpdated: d.lastUpdated,
-                    dateCreated: d.dateCreated,
-                    entry: d,
-                  ),
-                ),
-              );
+          _driveDao.driveById(driveId: driveId).watchSingle().listen((d) async {
+        final rootFolder = await _driveDao
```
matibat, line 56: The indentation is weird here

matibat (commented):
LGTM
Left one single comment regarding the indentation

lib/blocs/fs_entry_info/fs_entry_info_cubit.dart
jdaev: Updated

matibat (approved):

lib/pages/drive_detail/components/fs_entry_side_sheet.dart
```diff
@@ -161,16 +161,35 @@ class FsEntrySideSheet extends StatelessWidget {
...
                   rows: [
-                    if (infoState is FsEntryInfoSuccess<Drive>) ...{
+                    if (infoState is FsEntryDriveInfoSuccess) ...{
                       DataRow(cells: [
-                        DataCell(Text('Metadata Tx ID')),
+                        DataCell(Text('Drive Tx ID')),
                         DataCell(
                           CopyIconButton(
-                            tooltip: 'Copy Metadata Tx ID',
+                            tooltip: 'Copy Drive Tx ID',
```
arielmelendez, line 169: Why is this copy changing? There were no requirements in the ticket about this?

arielmelendez: I think I see why -  to disambiguate Drive Metadata Tx ID from Root Folder's. But there is nothing in the ticket about the need for this change. You should align with the team about a change like this before making it. My suggestion would be to use:
```
'Copy Drive Metadata Tx ID'
and
'Copy Root Folder Metadata Tx ID'
```

lib/blocs/fs_entry_info/fs_entry_info_state.dart
```diff
@@ -26,4 +26,32 @@ class FsEntryInfoSuccess<T> extends FsEntryInfoState {
...
 
+class FsEntryDriveInfoSuccess extends FsEntryInfoSuccess<Drive> {
+  @override
+  final String name;
+  @override
+  final DateTime lastUpdated;
+  @override
+  final DateTime dateCreated;
+
+  final Drive drive;
```
arielmelendez, line 37: Is it necessary to store a direct reference if the superclass is holding the `entry`?

lib/models/tables/drive_revisions.moor
```diff
@@ -15,6 +15,8 @@ CREATE TABLE drive_revisions (
 
     [action] TEXT NOT NULL,
 
+    bundledIn TEXT,
```
arielmelendez, line 18: Do we need another schema update?

arielmelendez (changes requested):
Some quick q's to answer.

lib/pages/drive_detail/components/fs_entry_side_sheet.dart
jdaev: Noted. Thank you

lib/blocs/fs_entry_info/fs_entry_info_state.dart
jdaev: Yeah. We need the drive there, or override entry if we want to pass a named parameter in the constructor.

lib/models/tables/drive_revisions.moor
jdaev: [Done](https://github.com/ardriveapp/ardrive-web/pull/316/commits/bb21021e4d9403259927eea861298a7a68d7129d)

arielmelendez: If we're going to keep forgetting to do this every time, then it may be time for some build automation that throws errors if changes are detected in these files without a schema update. This has happened in basically every DB update I've seen in the past few months. And/or a PR publishing checklist, though historically I haven't seen those implemented very successfully.

lib/blocs/fs_entry_info/fs_entry_info_state.dart
arielmelendez: image
Yea this seems to build fine for me. I don't see the need for the extra storage. If we need to access a property called drive seems like we could just add a getter that returns the entry.

arielmelendez (approved):

PR #319, opened 2022-01-11
PE-801 Rework Uploads

jdaev:
Reworks upload by reading file data from disk only inside bundle preparation and discarding any references that could cause excess memory consumption.

Suggest review order
- upload_cubit.dart
- file_upload_handle.dart
- data_item_upload_handle.dart
- bundle_upload_handle.dart
- arweave_service.dart
- the rest can be reviewed in any order


lib/blocs/upload/file_upload_handle.dart
```diff
@@ -1,17 +1,35 @@
...
   final FileEntity entity;
+  final XFile file;
   final String path;
-
+  final bool isPrivate;
+  final SecretKey? driveKey;
+  final SecretKey? fileKey;
   @override
   BigInt get cost {
     return (entityTx as Transaction).reward + (dataTx as Transaction).reward;
```
jdaev, line 26: Will fail for v2 files since entityTx is null at this point. TODO: Use size estimator here.

arielmelendez (changes requested):
Identified in pair review that V2 txs are broken

lib/utils/bundles/fake_tags.dart
```diff
@@ -0,0 +1,26 @@
...
+  Tag(EntityTag.cipher, Cipher.aes256),
+  Tag(
+    EntityTag.cipherIv,
+    'qwertyuiopasdfgh',
+  ),
+];
+
+final fakeApplicationTags = [
+  Tag(EntityTag.appName, 'ArDrive-Web'),
+  Tag(EntityTag.appVersion, '0.0.0'),
```
fedellen, line 15: I don't think we can assume that that this version will always have single digits. While its fine for now and the near future, we might bypass this to perhaps version `1.10.0` and not remember that this was hard coded here. Not a huge deal, it would only mean that we'd be under estimating by 1 byte

lib/services/arweave/arweave_service.dart
```diff
@@ -40,8 +49,23 @@ class ArweaveService {
...
+  }
+
+  BigInt calculateARPriceForByteSize({required int byteSize}) {
+    final chunks = bytesToChunks(byteSize);
+    if (_cachedPricePerChunk <= BigInt.zero) {
+      throw Error();
+    }
+    return _cachedPricePerChunk * BigInt.from(chunks);
+  }
+
```
fedellen, line 64: This does not seem to be accurate here. I wish it were this simple, it could save us a lot of pricing requests. First off, you need to factor in the base cost (`getPrice(byteSize: 0)`) and also the cost of one chunk (`getPrice(byteSize: 1)`) to determine the cost of each subsequent chunk ( oneChunk - basePrice = eachChunk )  . 

BUT there are additional variables that we haven't been able to solve here. And when we thought we had a solution, it proved to be false and the method eventually as under-funding rewards. Unfortunately we haven't been able to nail down an algorithm here after trying a couple times in Core. 

What we've done instead is cache prices on a chunk by chunk basis. Meaning we create and cache a mapping of: `[numberOfChunk]: winstonPrice`

So before each price request, we calculate how many chunks the provided byteCount would be. If we have a cached price for that number of chunks, we skip the price request and return that price. Otherwise we fetch a new price.

This saves on repeated price checks of the same chunk value. But with web as a potentially persisting program, I think we'll need some sort of cache invalidation after time to ensure that these values don't become stale.

jdaev: Thanks for that. I guess the way forward is to just use the price api like before.

arielmelendez (approved):

fedellen (approved):
🚀

PR #331, opened 2022-01-24
PE 919 fix ghost fixer

jdaev:
Fixes an issue where the bloc wouldn't stop emitting states.


matibat (approved):
Now we are emitting the `GhostFixerSuccess` only if the `try..catch` succeeds with no error. Looks good to me

PR #333, opened 2022-01-25
Release 1.5.0

jdaev:
- Implement fixer for missing parent folders
- Fix wrong last modified date on drag and drop uploads


matibat (approved):
The app version is correctly set as well as the schema version :+1:

PR #334, opened 2022-01-27
PE 910/ Pre push hook

jdaev:
This PR uses lefthook to add a pre push hook for checking database modifications.
It compares the diff, and checks if the database folder has been modified, and then stops the push if the database version has not been updated.


README.md
```diff
@@ -41,6 +41,9 @@ Finally, to start a development instance for web, run:
 flutter run -d Chrome
 ```
 
+Install lefthook for your platform from the intructions [here](https://github.com/evilmartians/lefthook/blob/master/docs/other.md).
```
arielmelendez, line 44: I'd put this at the TOP of this section.

lefthook/database_checker.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+if [ $(git diff @{u}...HEAD ./lib/models/tables | wc -l) -eq "0" ]; then
+    echo "  🟢 Database tables are unchanged."
+    exit 0
+else
+    previousSchemaPattern='([-][[:space:]]* int get schemaVersion =>[[:space:]]*[[:digit:]]+)'
```
arielmelendez, line 6:
```suggestion
    previousSchemaPattern='([-][[:space:]]*int get schemaVersion =>[[:space:]]*[[:digit:]]+)'
```

lefthook/database_checker.sh
```diff
@@ -0,0 +1,29 @@
...
+    currentSchemaPattern='([+][[:space:]]* int get schemaVersion =>[[:space:]]*[[:digit:]]+)'
```
arielmelendez, line 7:
```suggestion
    currentSchemaPattern='([+][[:space:]]*int get schemaVersion =>[[:space:]]*[[:digit:]]+)'
```

lefthook/database_checker.sh
```diff
@@ -0,0 +1,29 @@
...
+    previousVersion=${previousSchemaRegexMatch[${#previousSchemaRegexMatch[@]} - 1]}
+    # Force integer conversion
+    previousVersion=$(($previousVersion + 0))
+    [[ "${databaseSchemaDiff}" =~ $currentSchemaPattern ]]
+    currentSchemaRegexMatch=(${BASH_REMATCH[1]//[[:space:]]/ })
+    currentVersion=${currentSchemaRegexMatch[${#currentSchemaRegexMatch[@]} - 1]}
+    # Force integer conversion
+    currentVersion=$(($currentVersion + 0))
+
+    if [ $currentVersion -lt $previousVersion ]; then
```
arielmelendez, line 22: If something goes wrong with the regex match parsing and the integer conversion doesn't go well, could we end up with something like `0 -lt 0` and get a false positive from this script?

arielmelendez: We should really throw and error if we can't get a valid new schema version because it means that the schema version getter function has changed in a way we didn't anticipate. Perhaps the code below is already catching that case?

arielmelendez (changes requested):
I think the regexes can both have a space removed. Left suggested updates and a few questions.

lefthook/database_checker.sh
jdaev: Yeah, you were correct. I added a check for it [here](https://github.com/ardriveapp/ardrive-web/pull/334/commits/232bfe628a70e078ca90c841e360d9957f66fdf1)

arielmelendez (dismissed):

lefthook/database_checker.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+if [ $(git diff @{u}...HEAD ./lib/models/tables | wc -l) -eq "0" ]; then
+    echo "  🟢 Database tables are unchanged."
+    exit 0
+else
```
agsuy, line 5: What are you trying to achieve here with this else?

lefthook.yml
```diff
@@ -0,0 +1,6 @@
+pre-push:
+  parallel: true
```
agsuy, line 2: parallel with one command, while ok, takes no effect

lefthook/database_checker.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+if [ $(git diff @{u}...HEAD ./lib/models/tables | wc -l) -eq "0" ]; then
```
agsuy, line 2: Running this locally I get this outcome:

```
$ ./lefthook/database_checker.sh 
./lefthook/database_checker.sh: line 13: previousSchemaRegexMatch: bad array subscript
./lefthook/database_checker.sh: line 18: currentSchemaRegexMatch: bad array subscript
  🔴 Database schema parsing error. Quit
```

Seems like errors on:

`previousVersion=${previousSchemaRegexMatch[${#previousSchemaRegexMatch[@]} - 1]}`

and

`currentVersion=${currentSchemaRegexMatch[${#currentSchemaRegexMatch[@]} - 1]}`

agsuy (changes requested):
There are some concerns that script would not run as expected.

lefthook/database_checker.sh
jdaev: Yes

lefthook/database_checker.sh
agsuy: This was actually script logging errors to STDERR.

agsuy (approved):
Now we filter STDERR !
LGTM!

PR #338, opened 2022-01-31
refactor: don't create ghosts for attached drive PE-959

jdaev:
Checks whether the user owns the drive and if they don't skips ghost creation.

### DIFF
--- a/lib/blocs/sync/sync_cubit.dart
+++ b/lib/blocs/sync/sync_cubit.dart
@@ -202,10 +202,10 @@ class SyncCubit extends Cubit<SyncState> {
 
       // Dont create ghost folder if the ghost is a missing root folder
       // Or if the drive doesn't belong to the user
-      final isOwnDrive = drive.ownerAddress == ownerAddress;
+      final isReadOnlyDrive = drive.ownerAddress != ownerAddress;
       final isRootFolderGhost = drive.rootFolderId == ghostFolder.folderId;
 
-      if (!isOwnDrive && !isRootFolderGhost) {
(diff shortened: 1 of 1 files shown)


karlprieb (dismissed):
LGTM

lib/blocs/sync/sync_cubit.dart
```diff
@@ -198,7 +201,11 @@ class SyncCubit extends Cubit<SyncState> {
           await _driveDao.driveById(driveId: ghostFolder.driveId).getSingle();
 
       // Dont create ghost folder if the ghost is a missing root folder
-      if (drive.rootFolderId == ghostFolder.folderId) {
+      // Or if the drive doesn't belong to the user
+      final isReadOnlyDrive = drive.ownerAddress != ownerAddress;
+      final isRootFolderGhost = drive.rootFolderId == ghostFolder.folderId;
+
+      if (isReadOnlyDrive || isRootFolderGhost) {
```
arielmelendez, line 208: Seems ok. Is there anything we need to do to clean up existing DB entries that may have been added erroneously in released code? Additionally, have we considered what the "damage" might have been (if any) if users attempted to fix a ghost folder that wasn't in there drive?

arielmelendez (approved):
Left a question that should be answered before proceeding.

lib/blocs/sync/sync_cubit.dart
jdaev: We don't sync entities that don't belong to the user, so if someone fixes a folder on an attached drive, it would be visible only to them.
If we need to force clean then all we need is to update db schema version.

PR #340, opened 2022-02-01
fix: handle network errors for entity parse PE-969

jdaev:
This PR handles errors while fetching raw entity data from the arweave.net/txid endpoint during sync.


lib/services/arweave/arweave_service.dart
```diff
@@ -107,9 +109,9 @@ class ArweaveService {
     );
     final queryEdges = driveEntityHistoryQuery.data!.transactions.edges;
     final entityTxs = queryEdges.map((e) => e.node).toList();
-    final rawEntityData =
+    final responses =
         await Future.wait(entityTxs.map((e) => client.api.get(e.id)))
-            .then((rs) => rs.map((r) => r.bodyBytes).toList());
+            .then((rs) => rs.map((r) => r).toList());
```
karlprieb, line 114: Do we need this `map` here?

jdaev: [Removed](https://github.com/ardriveapp/ardrive-web/commit/be10557fa8c365308db34e845196bdb2cca487fd)

lib/services/arweave/arweave_service.dart
```diff
@@ -130,17 +132,25 @@ class ArweaveService {
       try {
         final entityType = transaction.getTag(EntityTag.entityType);
 
+        if (responses[i].statusCode != 200) {
+          throw Exception(
+            'Failed to fetch entity data for tx ${entityTxs[i].id}',
```
karlprieb, line 137:
```suggestion
            'Failed to fetch entity data for tx ${transcation.id}',
```
As we already have `entityTxs[i]` defined before I think that make sense to use it.

karlprieb: Similar suggestion can be applied to `responses[i]` that we use 2 times, it can increase readability, but it's up to you. I think it's readable enough the way it is.

karlprieb (approved):
LGTM

lib/services/arweave/arweave_service.dart
```diff
@@ -129,18 +131,25 @@ class ArweaveService {
 
       try {
         final entityType = transaction.getTag(EntityTag.entityType);
+        final entityResponse = responses[i];
+
+        if (entityResponse.statusCode != 200) {
+          throw EntityTransactionDataNetworkException();
```
arielmelendez, line 137: It could be nice to know exactly which request failed.

jdaev: [Added transaction id to the exception](https://github.com/ardriveapp/ardrive-web/pull/340/commits/12c905d046a349e09f916544f6692cd43c1cdf72)

arielmelendez (approved):

PR #343, opened 2022-02-03
PE 975/Fix wrong last page

jdaev:
Fixes issue where an empty page was shown at certain page counts.


arielmelendez (approved):
Approving, though I'd note that this PR probably wouldn't have been necessary if we'd factored out anything about page counting into a simple, unit tested object.

arielmelendez (approved):

matibat (approved):

PR #350, opened 2022-02-07
fix: add ghost fix button back PE-996

jdaev:
Adds back the Fix button for ghost folders.

### DIFF
--- a/lib/pages/drive_detail/components/drive_detail_data_table_source.dart
+++ b/lib/pages/drive_detail/components/drive_detail_data_table_source.dart
@@ -122,7 +122,15 @@ DataRow _buildFolderRow({
         ),
       ),
       DataCell(Text('-')),
-      DataCell(Text('-')),
+      folder.isGhost
+          ? DataCell(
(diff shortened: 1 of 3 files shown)


lib/pages/drive_detail/components/drive_detail_data_table_source.dart
```diff
@@ -122,7 +122,22 @@ DataRow _buildFolderRow({
         ),
       ),
       DataCell(Text('-')),
-      DataCell(Text('-')),
+      folder.isGhost
```
matibat, line 125: If it already handles the case of attached drives with ghost folders, then it is fine

matibat (approved):
It looks good to me

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
jdaev: Yeah, those aren't created in the first place.

arielmelendez (approved):

PR #364, opened 2022-02-16
chore: version up to 1.7.2 PE-1026

jdaev:
Also updates arweave-dart ref to tag v.2.0.1.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -47,8 +47,8 @@ packages:
     dependency: "direct main"
     description:
       path: "."
-      ref: ce5d2453300dd47077711a84979803939c8175dc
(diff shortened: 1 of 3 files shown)


pubspec.lock
```diff
@@ -47,8 +47,8 @@ packages:
     dependency: "direct main"
     description:
       path: "."
-      ref: ce5d2453300dd47077711a84979803939c8175dc
-      resolved-ref: ce5d2453300dd47077711a84979803939c8175dc
+      ref: "v.2.0.1"
```
arielmelendez, line 50: Um... why is there a dot after the v?

PR #367, opened 2022-02-18
PE 876/Generate private drive share links

jdaev:
This PR adds private drive sharing by passing the base64 encoded drive key as a query parameter in the shared link.


lib/utils/link_generators.dart
```diff
@@ -0,0 +1,23 @@
...
+import 'package:cryptography/cryptography.dart';
+import 'package:flutter/foundation.dart';
+
+Future<Uri> generateDriveShareLink({
+  required Drive drive,
+  SecretKey? driveKey,
+}) async {
+  // On web, link to the current origin the user is on.
+  // Elsewhere, link to app.ardrive.io.
+  final linkOrigin = kIsWeb ? Uri.base.origin : 'https://app.ardrive.io';
```
karlprieb, line 12: Can we create a constant somewhere for `https://app.ardrive.io` and use it here and on `file_share_cubit`?

jdaev: [Added](https://github.com/ardriveapp/ardrive-web/pull/367/commits/3b88ab05d040a2750e0fb06ab2401a946628220b)

lib/utils/link_generators.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:ardrive/models/models.dart';
+import 'package:arweave/utils.dart';
+import 'package:cryptography/cryptography.dart';
+import 'package:flutter/foundation.dart';
+
+Future<Uri> generateDriveShareLink({
```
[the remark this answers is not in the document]
jdaev: In this ticket an PR? That doesn't really have anything to do with the requirements on this one

lib/blocs/drive_share/drive_share_state.dart
```diff
@@ -26,3 +26,14 @@ class DriveShareLoadSuccess extends DriveShareState {
   @override
   List<Object> get props => [driveName, driveShareLink];
 }
+
```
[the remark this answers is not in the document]
jdaev: [Added](https://github.com/ardriveapp/ardrive-web/pull/367/commits/029b6db39496f08d923ce5b0305f1c96a4f7ebec)

lib/utils/link_generators.dart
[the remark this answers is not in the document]
karlprieb: It doesn't have anything with this ticket requirements, but I considered that as code maintenance as you are introducing better standards.

lib/blocs/drive_share/drive_share_cubit.dart
```diff
@@ -44,7 +44,7 @@ class DriveShareCubit extends Cubit<DriveShareState> {
 
     emit(
       DriveShareLoadSuccess(
-        driveName: drive.name,
+        drive: drive,
```
karlprieb, line 47: 👍

lib/blocs/file_share/file_share_cubit.dart
```diff
@@ -35,29 +35,25 @@ class FileShareCubit extends Cubit<FileShareState> {
...
+      fileKey: fileKey,
+    );
+
     emit(
       FileShareLoadSuccess(
         fileName: file.name,
-        fileShareLink: Uri.parse(fileShareLink),
-        isPublicFile: isPublicFile,
+        fileShareLink: fileShareLink,
+        isPublicFile: drive.isPrivate,
```
karlprieb, line 56: Shouldn't be the negation of `drive.isPrivate`?

lib/utils/link_generators.dart
```diff
@@ -0,0 +1,43 @@
...
+    final driveKeyBase64 = encodeBytesToBase64(await driveKey.extractBytes());
+
+    driveShareLink = driveShareLink + '&driveKey=$driveKeyBase64';
+  }
+  return Uri.parse(driveShareLink);
+}
+
+Future<Uri> generateFileShareLink({
+  required FileEntry file,
+  required Privacy drivePrivacy,
```
karlprieb, line 29: Not an issue, just a question: Why in some places we are passing `isPublic` or `isPrivate` and here we use `drivePrivacy`?

jdaev: Just trying to reduce the use of booleans where I can, especially where you have to pass it into a function.

lib/blocs/drive_share/drive_share_cubit.dart
```diff
@@ -7,34 +9,43 @@ part 'drive_share_state.dart';
...
-    // On web, link to the current origin the user is on.
-    // Elsewhere, link to app.ardrive.io.
-    final linkOrigin = kIsWeb ? Uri.base.origin : 'https://app.ardrive.io';
-    final driveName = drive.name;
-
-    final driveShareLink = '$linkOrigin/#/drives/${drive.id}?name=' +
-        Uri.encodeQueryComponent(driveName);
+    if (drive.isPrivate && !(_profileCubit.state is ProfileLoggedIn)) {
+      // Note: Should we even show the share drive link on attached private drive?
+      emit(DriveShareLoadFail(message: 'Please login to share private drive.'));
```
fedellen, line 33: Without a return here, doesn't this emit here get overwritten by the one below?

fedellen: Could eliminate the `late` here to make it easier to digest:

```dart
  if (drive.isPrivate && !(_profileCubit.state is ProfileLoggedIn)) {
      // Note: Should we even show the share drive link on attached private drive?
      emit(DriveShareLoadFail(message: 'Please login to share private drive.'));
      return;
    }

    final driveShareLink = await () async {
      if (drive.isPrivate) {
        final profileKey = (_profileCubit.state as ProfileLoggedIn).cipherKey;
        final driveKey = await _driveDao.getDriveKey(drive.id, profileKey);
        return generateDriveShareLink(
          drive: drive,
          driveKey: driveKey,
        );
      }

      return generateDriveShareLink(drive: drive);
    }();

    emit(
      // ...
    )
```

lib/entities/string_types.dart
```diff
@@ -1,3 +1,4 @@
 typedef FolderID = String;
 typedef FileID = String;
 typedef DriveID = String;
+typedef Privacy = String;
```
fedellen, line 4: Doesn't the DrivePrivacy type already provide this?

fedellen (dismissed):

lib/components/drive_share_dialog.dart
```diff
@@ -84,10 +85,12 @@ class _DriveShareDialogState extends State<DriveShareDialog> {
                   ),
                   const SizedBox(height: 16),
                   Text(
-                    'Anyone can access this drive using the link above.',
+                    'Anyone can access this ${state.drive.isPublic ? 'public' : 'private'} '
+                    'drive using the link above.',
```
matibat, line 89: Isn't there a need to use commas here?

matibat (dismissed):
LGTM

lib/blocs/drive_share/drive_share_cubit.dart
arielmelendez: `login` -> `log in`
login is a noun
log in is an imperative phrase

lib/components/drive_share_dialog.dart
jdaev: It's for formatting.

lib/blocs/drive_share/drive_share_cubit.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/367/commits/8d0b8a91e1b225e5bf51b3f4a0bd468603ca7ec0)

lib/blocs/drive_share/drive_share_cubit.dart
```diff
@@ -7,34 +9,46 @@ part 'drive_share_state.dart';
...
-    final drive = await _driveDao.driveById(driveId: driveId).getSingle();
-
-    // On web, link to the current origin the user is on.
-    // Elsewhere, link to app.ardrive.io.
-    final linkOrigin = kIsWeb ? Uri.base.origin : 'https://app.ardrive.io';
-    final driveName = drive.name;
-
-    final driveShareLink = '$linkOrigin/#/drives/${drive.id}?name=' +
-        Uri.encodeQueryComponent(driveName);
+    if (drive.isPrivate && !(_profileCubit.state is ProfileLoggedIn)) {
```
arielmelendez, line 31: Negation logic is tough on the reader. While I appreciate the value of an early return via guard clauses, I think the following is just easier to reason about:

```
    if (drive.isPrivate) {
      if (_profileCubit.state is ProfileLoggedIn) {
        final profileKey = (_profileCubit.state as ProfileLoggedIn).cipherKey;
        final driveKey = await _driveDao.getDriveKey(drive.id, profileKey);
        driveShareLink = await generateDriveShareLink(
          drive: drive,
          driveKey: driveKey,
        );
      } else {
        emit(
          DriveShareLoadFail(message: 'Please log in to share private drive.'),
        );
        return;
      }
    } else {
      driveShareLink = await generateDriveShareLink(drive: drive);
    }
```

jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/367/commits/4202d469896223116a414d7fdf87c03cdb4d0144)

lib/utils/link_generators.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:ardrive/entities/constants.dart';
+import 'package:ardrive/entities/string_types.dart';
+import 'package:ardrive/models/models.dart';
+import 'package:arweave/utils.dart';
+import 'package:cryptography/cryptography.dart';
+import 'package:flutter/foundation.dart';
+
+Future<Uri> generateDriveShareLink({
+  required Drive drive,
```
arielmelendez, line 9: So I'd like to point out some of the consequences of the design choices here.

The variables needed to generate a drive share link are:

1. a drive name (a string)
2. a drive ID (a string)
3. a drive key (a SecretKey)

The choice to provide a Drive object to satisfy 1 and 2 creates a dependency on an object. And what's more, that object is actually a database entity.

Consequence 1: This function has no value to any other application or library now without a database.

By providing the Drive object, we get lulled into thinking:
- that the drive name and drive id, since they're traveling together, must be cohesive and correct
- that we'll make a good decision about what to do about the optionally provided drive key since we'll have access to the drive object's privacy setting.

But these beliefs are an illusion because:
- there's no guarantee that the Drive object, as assembled, is "legitimate".
- there's no guarantee that the provided drive key is the right one for the provided drive object.

We don't have access to the classes necessary to prove either of those things, nor would it be palatable to introduce those here due to crushing number of dependencies and complexity it would introduce. That sort of validation could and should be handled before calls to this (potentially) simple, pure input/output function.

Consequence 2: More branching logic that needs to be tested to make sure that the function gives the correct outputs given various permutations of Drive objects and key nullability.

Consequence 3: The potential for the function to be used incorrectly and the need to throw errors (e.g. when it's a private drive but no drive key is provided.

I think you can control for all three of the consequences above, put the heavier business logic back where it belongs (e.g. cubits and/or supporting helper classes), and make this function easier to use correctly and harder to use incorrectly like so:

```
Uri generatePublicDriveShareLink({
  required final String driveName,
  required final String driveId,
}) {
  // On web, link to the current origin the user is on.
  // Elsewhere, link to app.ardrive.io.
  final linkOrigin = kIsWeb ? Uri.base.origin : linkOriginProduction;

  final driveShareLink = '$linkOrigin/#/drives/${driveId}?name=' +
      Uri.encodeQueryComponent(driveName);
  return Uri.parse(driveShareLink);
}

Future<Uri> generatePrivateDriveShareLink({
  required final String driveName,
  required final String driveId,
  required final SecretKey driveKey,
}) async {
  final driveKeyBase64 = encodeBytesToBase64(await driveKey.extractBytes());

  return Uri.parse(generatePublicDriveShareLink(
        driveName: driveName,
        driveId: driveId,
      ).toString() +
      '&driveKey=$driveKeyBase64');
}
```

test/utils/link_generators_test.dart
```diff
@@ -0,0 +1,135 @@
...
+          dateCreated: DateTime.now(),
+          lastUpdated: DateTime.now(),
+        );
+        testPrivateDriveKeyBase64 =
+            'X123YZAB-CD4e5fgHIjKlmN6O7pqrStuVwxYzaBcd8E';
+        testPrivateDriveKey =
+            SecretKey(decodeBase64ToBytes(testPrivateDriveKeyBase64));
+      });
+      test(
+          'generateDriveShareLink generates the correct link for a private drive',
```
arielmelendez, line 40: Love that you added tests for these (seemingly) simple functions! As I mentioned in the design decisions message above, since it's possible for your code as currently written to pass a private drive and no secret key, you need a test for that permutation. That said, that's probably an illegal permutation that we don't even want. So either:
a) Test for the current behavior you have coded up which is to return the equivalent of a public drive link (which is not really desirable since that link won't even work anywhere, so why let the app even generate them?)
b) Throw an error when that state is encountered and add a test for that.
c) Update to the public vs. private implementation that I suggested where the illegal states aren't necessarily possible since the business logic considerations around validity are hoisted up to the callers that have the appropriate tools to make good choices.

arielmelendez (changes requested):
Made a comment about the design choices for generateDriveShareLink (it's applicable to the file version as well). Whether or not you decide to make the changes I suggested, you'll need to update tests.

lib/utils/link_generators.dart
jdaev: Thanks for the insight, I made the mistake of assuming that I was just using an entity class instead of a database class without looking into it. 
[Updated](https://github.com/ardriveapp/ardrive-web/pull/367/commits/62b5a73987a0d5ca459eb0b1d62b980b01207f91)

test/utils/link_generators_test.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/367/commits/62b5a73987a0d5ca459eb0b1d62b980b01207f91)

jdaev:
Note: Do not merge before attach private drives is also approved. Sharing private drives is useless without it.

arielmelendez (approved):
Thanks for the updates!

fedellen (approved):
LGTM!

Though if this PR depends on another PR, I would expect the base branch to be the one it depends on rather than `dev`

jdaev:
> 

This one doesn't depend on anything, but Attach Private Drives depends on this

lib/blocs/file_share/file_share_cubit.dart
```diff
@@ -35,29 +36,43 @@ class FileShareCubit extends Cubit<FileShareState> {
...
+        fileShareLink = await generatePrivateFileShareLink(
+          fileId: file.id,
+          fileKey: fileKey,
+        );
+
+        break;
+      case DrivePrivacy.public:
+        fileShareLink = generatePublicFileShareLink(fileId: file.id);
+        break;
+      default:
```
matibat, line 68: This empty `default` could be handled at PE-1115

matibat (approved):
:+1:

PR #372, opened 2022-02-28
feat/parallelize chunk uploads PE 1056

jdaev:
Adds faster uploads


matibat (dismissed):
Updates the arweave dep

arielmelendez (changes requested):
We're going to close this PR in favor of JY's updates yes?

jdaev:
@arielmelendez Yes. I left this open in case the deployed link was handed out to users.

PR #383, opened 2022-03-07
chore: update index.html version number

javdhu:
Updates the index.html version number from 57 to 58 on staging.
This tells the browser to stop serving the cached page and update the cache.


agsuy (approved):
Build version up.

PR #385, opened 2022-03-07
Pe-1003 skip file conflicts on upload

thiagocarvalhodev:
## Overview
When a user uploads files to a folder, a check is made to see if any existing file names conflict with the files being uploaded.  The user only has the option to replace the existing files and overwrite them with the new data.

Basically, it was added the `SKIP` button capability - it will ignore the conflicting files; And rename the `CONTINUE` button to `REPLACE` - this one still with the same behavior.

## Tests
This PR intends to implement the tests, specifically UploadCubit. To do so, I've needed to refactor some code:

#### UploadPlanUtils:
I moved the function `xfilesToUploadPlan` from `UploadCubit` to another class which was a hard function to mock. I called it UploadPlanUtils but I'm wondering about a better name.

#### UploadCubit:
I've changed the initialization of the object. Once we couldn't wait for the object initialization, it made weird behavior in the tests, like this:

image

Instead, I move the object initialization to the `startUploadPreparation` function. To notify the page that the initialization was made, we could check by listening to the cubit and waiting for the `UploadCubitInitialized` state. It was implemented in the pages that use the UploadCubit.

### DIFF
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -2,25 +2,19 @@ import 'dart:async';
 import 'dart:math' as math;
 
 import 'package:ardrive/blocs/upload/cost_estimate.dart';
-import 'package:ardrive/blocs/upload/data_item_upload_handle.dart';
 import 'package:ardrive/blocs/upload/upload_plan.dart';
-import 'package:ardrive/entities/entities.dart';
 import 'package:ardrive/models/models.dart';
 import 'package:ardrive/services/services.dart';
 import 'package:ardrive/utils/upload_plan_utils.dart';
-import 'package:arweave/arweave.dart';
 import 'package:bloc/bloc.dart';
-import 'package:cryptography/cryptography.dart';
 import 'package:equatable/equatable.dart';
 import 'package:file_selector/file_selector.dart';
 import 'package:meta/meta.dart';
-import 'package:mime/mime.dart';
 import 'package:pedantic/pedantic.dart';
 import 'package:uuid/uuid.dart';
 
 import '../blocs.dart';
 import 'enums/conflicting_files_actions.dart';
-import 'file_upload_handle.dart';
 
 part 'upload_state.dart';
 
@@ -221,68 +215,6 @@ class UploadCubit extends Cubit<UploadState> {
     unawaited(_profileCubit.refreshBalance());
 
     emit(UploadComplete());
-  }
-
-  Future<UploadPlan> xfilesToUploadPlan({
-    required List<XFile> files,
-    required SecretKey cipherKey,
-    required Wallet wallet,
-  }) async {
-    final _dataItemUploadHandles = <String, DataItemUploadHandle>{};
-    final _v2FileUploadHandles = <String, FileUploadHandle>{};
-    for (var file in files) {
-      final fileName = file.name;
-      final filePath = '${_targetFolder.path}/$fileName';
-      final fileSize = await file.length();
-      final fileEntity = FileEntity(
-        driveId: _targetDrive.id,
-        name: fileName,
-        size: fileSize,
-        lastModifiedDate: await file.lastModified(),
-        parentFolderId: _targetFolder.id,
-        dataContentType: lookupMimeType(fileName) ?? 'application/octet-stream',
-      );
-
-      // If this file conflicts with one that already exists in the target folder reuse the id of the conflicting file.
-      fileEntity.id = conflictingFiles[fileName] ?? _uuid.v4();
-
-      final private = _targetDrive.isPrivate;
-      final driveKey = private
-          ? await _driveDao.getDriveKey(_targetDrive.id, cipherKey)
-          : null;
-      final fileKey =
-          private ? await deriveFileKey(driveKey!, fileEntity.id!) : null;
-
-      final revisionAction = !conflictingFiles.containsKey(file.name)
-          ? RevisionAction.create
-          : RevisionAction.uploadNewVersion;
-
-      if (fileSize < bundleSizeLimit) {
-        _dataItemUploadHandles[fileEntity.id!] = DataItemUploadHandle(
-          entity: fileEntity,
-          path: filePath,
-          file: file,
-          driveKey: driveKey,
-          fileKey: fileKey,
-          arweave: _arweave,
-          wallet: wallet,
-          revisionAction: revisionAction,
-        );
-      } else {
-        _v2FileUploadHandles[fileEntity.id!] = FileUploadHandle(
-          entity: fileEntity,
-          path: filePath,
-          file: file,
-          driveKey: driveKey,
-          fileKey: fileKey,
-          revisionAction: revisionAction,
-        );
-      }
-    }
-    return UploadPlan.create(
-      v2FileUploadHandles: _v2FileUploadHandles,
-      dataItemUploadHandles: _dataItemUploadHandles,
-    );
   }
 
   void _removeConflictingFiles() {
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -56,13 +56,13 @@ class UploadCubit extends Cubit<UploadState> {
         _uploadPlanUtils = uploadPlanUtils,
         super(UploadPreparationInProgress());
 
-  Future<void> initializeCubit() async {
+  Future<void> getDriveInformation() async {
     _targetDrive = await _driveDao.driveById(driveId: driveId).getSingle();
     _targetFolder = await _driveDao
         .folderById(driveId: driveId, folderId: folderId)
         .getSingle();
 
-    emit(UploadCubitInitialized());
+    emit(UploadPreparationInitialized());
   }
 
   /// Tries to find a files that conflict with the files in the target folder.
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -18,6 +18,7 @@ import 'package:pedantic/pedantic.dart';
 import 'package:uuid/uuid.dart';
 
 import '../blocs.dart';
+import 'enums/conflicting_files_actions.dart';
 import 'file_upload_handle.dart';
 
 part 'upload_state.dart';
@@ -93,13 +94,16 @@ class UploadCubit extends Cubit<UploadState> {
 
     if (conflictingFiles.isNotEmpty) {
       emit(UploadFileConflict(
+          isAllFilesConflicting: conflictingFiles.length == files.length,
           conflictingFileNames: conflictingFiles.keys.toList()));
     } else {
       await prepareUploadPlanAndCostEstimates();
     }
   }
 
-  Future<void> prepareUploadPlanAndCostEstimates() async {
+  /// If `conflictingFileAction` is null, means that had no conflict.
+  Future<void> prepareUploadPlanAndCostEstimates(
+      {ConflictingFileActions? conflictingFileAction}) async {
     final profile = _profileCubit.state as ProfileLoggedIn;
 
     if (await _profileCubit.checkIfWalletMismatch()) {
@@ -114,6 +118,11 @@ class UploadCubit extends Cubit<UploadState> {
     );
     final sizeLimit =
         _targetDrive.isPrivate ? privateFileSizeLimit : publicFileSizeLimit;
+
+    if (conflictingFileAction == ConflictingFileActions.Skip) {
+      _removeConflictingFiles();
+    }
+
     final tooLargeFiles = [
       for (final file in files)
         if (await file.length() > sizeLimit) file.name
@@ -267,6 +276,10 @@ class UploadCubit extends Cubit<UploadState> {
       v2FileUploadHandles: _v2FileUploadHandles,
       dataItemUploadHandles: _dataItemUploadHandles,
     );
+  }
+
+  void _removeConflictingFiles() {
+    files.removeWhere((element) => conflictingFiles.containsKey(element.name));
   }
 
   @override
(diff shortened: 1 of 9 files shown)


lib/blocs/upload/upload_cubit.dart
```diff
@@ -45,27 +40,29 @@ class UploadCubit extends Cubit<UploadState> {
...
-      _targetFolder = await _driveDao
-          .folderById(driveId: driveId, folderId: folderId)
-          .getSingle();
-
-      unawaited(checkConflictingFiles());
-    }();
+        _uploadPlanUtils = uploadPlanUtils,
+        super(UploadPreparationInProgress());
+
+  Future<void> initializeCubit() async {
```
jdaev, line 59: Consider calling this from inside the constructor. Otherwise we'd have to think of some way to force the user to call the initializeCubit function, maybe make the constructor private and have a static initialization function that returns the Cubit.

lib/blocs/upload/upload_cubit.dart
```diff
@@ -45,27 +40,29 @@ class UploadCubit extends Cubit<UploadState> {
...
-  })  : _profileCubit = profileCubit,
+  UploadCubit(
+      {required this.driveId,
+      required this.folderId,
+      required this.files,
+      required ProfileCubit profileCubit,
+      required DriveDao driveDao,
+      required ArweaveService arweave,
+      required PstService pst,
+      required UploadPlanUtils uploadPlanUtils})
```
jdaev, line 51: You could achieve what you want by moving the xFileToUploadPlan function by making it static and keep it in UploadCubit, but pass all the dependencies it needs as arguments.
You can also just create and use the UploadPlanUtils class from inside the UploadCubit without the user ever having to pass it in, the dependencies you need to create it are in UploadCubit anyway.

lib/blocs/upload/upload_cubit.dart
```diff
@@ -45,27 +40,29 @@ class UploadCubit extends Cubit<UploadState> {
...
+        _uploadPlanUtils = uploadPlanUtils,
+        super(UploadPreparationInProgress());
+
+  Future<void> initializeCubit() async {
+    _targetDrive = await _driveDao.driveById(driveId: driveId).getSingle();
+    _targetFolder = await _driveDao
+        .folderById(driveId: driveId, folderId: folderId)
+        .getSingle();
+
+    emit(UploadCubitInitialized());
```
jdaev, line 65: UploadPreparationInitial would be a better name here. It will match what we have in other cubits and would be in line with the states we have already defined.

lib/blocs/upload/upload_cubit.dart
```diff
@@ -93,13 +90,16 @@ class UploadCubit extends Cubit<UploadState> {
...
     if (conflictingFiles.isNotEmpty) {
       emit(UploadFileConflict(
+          isAllFilesConflicting: conflictingFiles.length == files.length,
           conflictingFileNames: conflictingFiles.keys.toList()));
     } else {
       await prepareUploadPlanAndCostEstimates();
     }
   }
 
-  Future<void> prepareUploadPlanAndCostEstimates() async {
```
jdaev, line 102: Could be better to have another function above this that resolves the conflicts and calls prepareUploadPlanAndCostEstimates itself.

lib/blocs/upload/upload_cubit.dart
thiagocarvalhodev: Thanks, Javed. I think that the static function solution wouldn't work. This function becomes very hard to test, especially because of this file:
https://github.com/ardriveapp/arweave-dart/blob/v2.0.1/lib/src/utils/implementations/bundle_tag_parser_ffi.dart

We could do the second option, but we need to pass this object to mock and test by dependency injection.  

One solution is to pass an object by dependency injection, and if the user didn't pass, initialize it on the constructor. 

What do you think?

jdaev: Why would the static solution not work? Bundle tag parser will only throw when you try to sign. If you don't call startUpload it should be fine.

jdaev: Disregard the static comment, having a static sync constructor wouldn't work with blocbuilder or bloclisters, but you can just create the UploadUtils class inside initializeCubit.

lib/components/upload_form.dart
```diff
@@ -46,12 +51,15 @@ class UploadForm extends StatelessWidget {
...
             Navigator.pop(context);
+          } else if (state is UploadCubitInitialized) {
+            await context.read<UploadCubit>().checkConflictingFiles();
           }
           if (state is UploadWalletMismatch) {
             Navigator.pop(context);
             await context.read<ProfileCubit>().logoutProfile();
           }
         },
+        // buildWhen: (_, currentState) => currentState is UploadCubitInitialized,
```
thiagocarvalhodev, line 62: I will remove that.

lib/blocs/upload/upload_cubit.dart
thiagocarvalhodev: I mean that it will not work to test. The function `serializeTags`, at this file above, is called when we try to create a UploadPlan, and throw and the test fails. Also, we need to set up a lot of functionalities to make that test.

jdaev: Ah, your're right. It's called during price estimation. I think it should be fine to sub the price estimation calls and return some value since it's not relevant to the tests you're trying to do. 
I'll leave it to @karlprieb and @arielmelendez to weigh in

lib/blocs/upload/upload_cubit.dart
thiagocarvalhodev: @jdaev, I think that one another approach for that is to initialize the cubit in its creation. I know that it needs more complexity but becomes easier to test.

For example:

-> Open the Upload UI
-> Instantiate the UploadCubit and just after its initialization, call the `initializeCubit`. 

```dart
BlocProvider(
  create: UploadCubit(
     arguments...
  ).. initializeCubit(),
);
```

An example:
https://github.com/felangel/bloc/blob/master/examples/flutter_shopping_cart/lib/app.dart

thiagocarvalhodev: I'll also move the body of initializeCubit function to the checkConflictingFiles. It is the first point that UploadCubit verifies.

It will be:

```dart
BlocProvider(
  create: UploadCubit(
     arguments...
  ).. checkConflictingFiles(),
);
```

thiagocarvalhodev: I'll also rename the initializeCubit to getDriveInformation.

It becomes:

```dart
BlocProvider(
  create: UploadCubit(
     arguments...
  )..getDriveInformation(),
);
```

lib/components/upload_form.dart
thiagocarvalhodev: [resolved](https://github.com/ardriveapp/ardrive-web/pull/385/commits/2f240240d2cfcdc1846a6bb1b823e709addcc148)

lib/blocs/upload/upload_cubit.dart
thiagocarvalhodev: [Changed](https://github.com/ardriveapp/ardrive-web/pull/385/commits/4c2e370c55d867c6a27b0301622688b7d6fef5fb)

lib/blocs/upload/upload_cubit.dart
```diff
@@ -2,40 +2,35 @@ import 'dart:async';
...
 
 import '../blocs.dart';
-import 'file_upload_handle.dart';
+import 'enums/conflicting_files_actions.dart';
 
 part 'upload_state.dart';
 
 final privateFileSizeLimit = 104857600;
-final publicFileSizeLimit = 1.25 * math.pow(10, 9) as int;
+final publicFileSizeLimit = 1.25 * math.pow(10, 9);
```
karlprieb, line 21: I know that it's not trivial, and maybe this case will never happen, but this calc can return a float if you really want
If you have more than 8 decimal places, you should get a float.
Possibly we can use `floor` to guarantee that. I don't think performance will suffer.

test/blocs/upload_cubit_test.dart
```diff
@@ -0,0 +1,342 @@
...
+      when(() => mockProfileCubit!.isCurrentProfileArConnect())
+          .thenAnswer((i) => Future.value(false));
+      when(() => mockPst.getPSTFee(BigInt.zero))
+          .thenAnswer((invocation) => Future.value(BigInt.zero));
+      when(() => mockArweave.getArUsdConversionRate())
+          .thenAnswer((invocation) => Future.value(10));
+      setDumbUploadPlan();
+    });
+    blocTest<UploadCubit, UploadState>(
+        'Should emit UploadFileConflict with correctly file names and'
```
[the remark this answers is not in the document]
thiagocarvalhodev: Thanks, Karl!

lib/blocs/upload/upload_cubit.dart
karlprieb: In fact we don't care if it's an int or double. `publicFileSizeLimit` will be assigned to `sizeLimit` https://github.com/ardriveapp/ardrive-web/blob/0b47c6f7f604fd38b41a3814be88e094af3de303/lib/blocs/upload/upload_cubit.dart#L123 and we use `sizeLimit` to compare with `file.length()`. It really doesn't matter if it's an int or double as Dart can compare int and double, for example `10 < 10.1`.

lib/blocs/upload/upload_state.dart
```diff
@@ -27,8 +29,11 @@ class UploadSigningInProgress extends UploadState {
 
 class UploadFileConflict extends UploadState {
   final List<String> conflictingFileNames;
+  final bool isAllFilesConflicting;
 
-  UploadFileConflict({required this.conflictingFileNames});
+  UploadFileConflict(
+      {required this.conflictingFileNames,
+      required this.isAllFilesConflicting});
```
jdaev, line 36: Add trailing comma

lib/components/upload_form.dart
```diff
@@ -25,14 +27,17 @@ Future<void> promptToUploadFile(
       context: context,
       builder: (_) => BlocProvider<UploadCubit>(
         create: (context) => UploadCubit(
+          uploadPlanUtils: UploadPlanUtils(
+              arweave: context.read<ArweaveService>(),
+              driveDao: context.read<DriveDao>()),
```
jdaev, line 32: Add trailing comma after <DriveDao>()

lib/pages/drive_detail/components/drive_file_drop_zone.dart
```diff
@@ -111,14 +112,17 @@ class _DriveFileDropZoneState extends State<DriveFileDropZone> {
           context: context,
           builder: (_) => BlocProvider<UploadCubit>(
             create: (context) => UploadCubit(
+              uploadPlanUtils: UploadPlanUtils(
+                  arweave: context.read<ArweaveService>(),
+                  driveDao: context.read<DriveDao>()),
```
jdaev, line 117: Please add trailing a comma here.

lib/blocs/upload/upload_cubit.dart
jdaev: Looks good. Thanks for the hard work 👍🏽

lib/pages/drive_detail/components/drive_file_drop_zone.dart
thiagocarvalhodev: [Resolved](https://github.com/ardriveapp/ardrive-web/pull/385/commits/8639ff18a1de6f4bd441c60121c8ade196704ce3)

lib/components/upload_form.dart
thiagocarvalhodev: [Resolved](https://github.com/ardriveapp/ardrive-web/pull/385/commits/8639ff18a1de6f4bd441c60121c8ade196704ce3)

lib/blocs/upload/upload_state.dart
jdaev: > I still don't understand why does Dart behave that way with indentation and commas lol

https://docs.flutter.dev/development/tools/formatting#using-trailing-commas

lib/blocs/upload/upload_cubit.dart
matibat: Hmmm where do these numbers come from? Could we have more context on what `1.25`, `10` and `9` stand for? :monocle_face:

matibat (commented):
It's looking good! :sunglasses:

lib/blocs/upload/upload_state.dart
matibat: Got ya! A nice implication of this is reducing the changes diff: As there's a trailing comma, when we are adding more items to the list we don't have to add that comma and the diff is only an extra line. I kind of like that

lib/components/upload_form.dart
```diff
@@ -80,12 +88,21 @@ class UploadForm extends StatelessWidget {
                   onPressed: () => Navigator.of(context).pop(false),
                   child: Text('CANCEL'),
                 ),
+                if (!state.isAllFilesConflicting)
+                  ElevatedButton(
+                    onPressed: () => context
+                        .read<UploadCubit>()
+                        .prepareUploadPlanAndCostEstimates(
+                            conflictingFileAction: ConflictingFileActions.Skip),
+                    child: Text('SKIP'),
```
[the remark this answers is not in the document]
thiagocarvalhodev: Thanks, Mati!

lib/blocs/upload/upload_state.dart
karlprieb: We should add this rule to our TS projects :)

karlprieb (dismissed):

fedellen (dismissed):
Nice bloc test examples here!

karlprieb (approved):

fedellen (approved):

matibat (approved):
:sunglasses:

karlprieb (approved):

PR #387, opened 2022-03-08
fix(drive_detail): fix folder info metadata tx not being shown PE-1093

jdaev:
Corrects the type used in the folder info state.


matibat (approved):
:+1:

PR #388, opened 2022-03-08
fix(drive_detail): fix changed state not updating data table PE-1084

jdaev:
Adds equatable bust to the drive detail load state for the drive table object key to compare equality.

A widget will only re-render if the key is different.

### DIFF
--- a/lib/pages/drive_detail/components/drive_detail_data_table.dart
+++ b/lib/pages/drive_detail/components/drive_detail_data_table.dart
@@ -2,6 +2,7 @@ part of '../drive_detail_page.dart';
 
 Widget _buildDataTable(BuildContext context, DriveDetailLoadSuccess state) {
   return DriveDataTable(
+    key: ObjectKey(state),
     driveDetailState: state,
     context: context,
   );
--- a/lib/blocs/drive_detail/drive_detail_state.dart
+++ b/lib/blocs/drive_detail/drive_detail_state.dart
@@ -27,6 +27,7 @@ class DriveDetailLoadSuccess extends DriveDetailState {
 
   final int rowsPerPage;
   final List<int> availableRowsPerPage;
+  final int _equatableBust = DateTime.now().millisecondsSinceEpoch;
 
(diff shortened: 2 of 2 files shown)


lib/pages/drive_detail/components/drive_detail_data_table.dart
```diff
@@ -2,6 +2,7 @@ part of '../drive_detail_page.dart';
 
 Widget _buildDataTable(BuildContext context, DriveDetailLoadSuccess state) {
   return DriveDataTable(
+    key: ObjectKey(state),
```
thiagocarvalhodev, line 5: Could we compare by using just the _equatableBust with a ValueKey?

jdaev: Yeah, it's also an option.

thiagocarvalhodev: Great! I think that it will be less expensive.

matibat (dismissed):
:+1:

karlprieb (approved):

lib/pages/drive_detail/components/fs_entry_side_sheet.dart
```diff
@@ -210,7 +210,7 @@ class FsEntrySideSheet extends StatelessWidget {
           builder: (context, state) {
             if (state is FsEntryActivitySuccess) {
               if (state.revisions.isNotEmpty) {
-                final revision = state.revisions.last;
+                final revision = state.revisions.first;
```
fedellen, line 213: This looks like its from a different ticket

Is this correct here?

jdaev: Seems like I accidentally merged another branch into this one. Shouldn't matter now since the same commits are on the integration branch

fedellen: Sounds good, just confirming that `first` is the correct syntax here. It seems like it is

PR #393, opened 2022-03-10
chore: point dev to default arlocal endpoint

jdaev:


jdaev:
> Was this just to bump the firebase preview?
> 
> We can probably close this identical PR/branch now: #347

Just me testing things. Still getting problems with arlocal, so closing this until open another PR and have a ticket to track it.

PR #394, opened 2022-03-10
integration branch for ui fixes

jdaev:
Integration branch of UI fixes for easier testing.

Contains:
PE-1084
PE-1093
PE-905
PE-1100

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -37,7 +37,7 @@
   <script defer src="avsc.min.js"></script>
   <script defer src="js/tagparser.js"></script>
   <script defer src="sql-wasm.js"></script>
-  <script src="main.dart.js?version=59" type="application/javascript"></script>
+  <script src="main.dart.js?version=60" type="application/javascript"></script>
   
 </body>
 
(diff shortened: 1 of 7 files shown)


fedellen (approved):
Nice bug fixes here!

web/index.html
```diff
@@ -37,7 +37,7 @@
   <script defer src="avsc.min.js"></script>
   <script defer src="js/tagparser.js"></script>
   <script defer src="sql-wasm.js"></script>
-  <script src="main.dart.js?version=59" type="application/javascript"></script>
```
karlprieb, line 40: We can keep 59... it was bumped yesterday. Currently, `master` is at 58.

jdaev: Isn't every commit to dev supposed to up this now?

karlprieb: If we need to update this version on each staging deploy make sense to not bump when release a new version and do on each PR to dev. Otherwise we could just bump that on each prod release

karlprieb: We will keep updating webapp version on `dev` after prod release, but this asset version should be updated on each PR to `dev`.

karlprieb (approved):

PR #395, opened 2022-03-11
chore: update pubspec PE-1103

jdaev:
Updates the app version and arweave-dart.


fedellen (approved):
Good to go as long as v1.8.1 is what we're going with here

thiagocarvalhodev (approved):

PR #400, opened 2022-03-14
refactor: add tableKey property to CustomDataTable PE-1109

jdaev:
Uses a key only in the CustomDataTable widget to avoid scroll and page reset. Removes all other keys from the drive data table.


matibat (approved):
:+1:

PR #401, opened 2022-03-15
PE-1105: Disallow empty manifest

fedellen:
This PR :
- Disables create manifest button on drives that have no file entities
- Disables folders that have no file entities during folder selection
- Adjusts folder selection state's modal title and button text

### DIFF
--- a/lib/blocs/drive_detail/drive_detail_state.dart
+++ b/lib/blocs/drive_detail/drive_detail_state.dart
@@ -11,6 +11,7 @@ class DriveDetailLoadInProgress extends DriveDetailState {}
 class DriveDetailLoadSuccess extends DriveDetailState {
   final Drive currentDrive;
   final bool hasWritePermissions;
+  final bool driveIsEmpty;
 
   final FolderWithContents folderInView;
 
@@ -40,6 +41,7 @@ class DriveDetailLoadSuccess extends DriveDetailState {
     this.maybeSelectedItem,
     this.showSelectedItemDetails = false,
     this.selectedFilePreviewUrl,
+    required this.driveIsEmpty,
   });
 
   DriveDetailLoadSuccess copyWith({
@@ -53,6 +55,7 @@ class DriveDetailLoadSuccess extends DriveDetailState {
     Uri? selectedFilePreviewUrl,
     int? rowsPerPage,
     List<int>? availableRowsPerPage,
+    bool? driveIsEmpty,
   }) =>
       DriveDetailLoadSuccess(
         currentDrive: currentDrive ?? this.currentDrive,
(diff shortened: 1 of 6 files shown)


lib/blocs/drive_detail/drive_detail_state.dart
```diff
@@ -67,6 +70,7 @@ class DriveDetailLoadSuccess extends DriveDetailState {
             selectedFilePreviewUrl ?? this.selectedFilePreviewUrl,
         availableRowsPerPage: availableRowsPerPage ?? this.availableRowsPerPage,
         rowsPerPage: rowsPerPage ?? this.rowsPerPage,
+        driveIsEmpty: driveIsEmpty ?? this.driveIsEmpty,
```
karlprieb, line 73: I don't know if I understand that.
You're passing `driveIsEmpty` from params or `this.driveIsEmpty` that, I believe, is the same data as we assign that early, what am I missing?

karlprieb (approved):

lib/components/create_manifest_form.dart
```diff
@@ -280,16 +281,28 @@ class CreateManifestForm extends StatelessWidget {
...
+          /// Returns false if a folder does not contain any file entities
+          bool isFolderEmpty(FolderID folderId) {
+            final folderNode =
+                readCubitContext.rootFolderNode.searchForFolder(folderId);
+
+            if (folderNode == null) {
+              return true;
+            }
+
+            return folderNode.getRecursiveFileCount() == 0;
```
thiagocarvalhodev, line 293: This was implemented before at this file: [create_manifest_cubit.dart](https://github.com/ardriveapp/ardrive-web/pull/401/files#diff-ba2152939d6eaf5045f9885785a7fd4fbc7688053d078440543258784f934e77). Would be better if we have this function on its entity or in some helper class. What do you think?

thiagocarvalhodev: it could be just:
```dart
folderNode.isEmpty()
```

lib/components/create_manifest_form.dart
```diff
@@ -280,16 +281,28 @@ class CreateManifestForm extends StatelessWidget {
         }
 
         if (state is CreateManifestFolderLoadSuccess) {
+          /// Returns false if a folder does not contain any file entities
+          bool isFolderEmpty(FolderID folderId) {
```
thiagocarvalhodev, line 285: It's more common to declare functions outside of other functions' scope. Is better at the class level as a private method.

lib/components/create_manifest_form.dart
fedellen: Added here: 0a4a53fa

lib/components/create_manifest_form.dart
fedellen: Added here: 0a4a53fa

thiagocarvalhodev (approved):

PR #404, opened 2022-03-15
Release 1.9.0 PE-1113

karlprieb:
Release Notes:
* Adds support for sharing and attaching private drives.
* Fix scroll reset when selecting a file or folder.


matibat (approved):
:+1:

PR #405, opened 2022-03-15
fix(file revision): Pass dataContentType on rename and move actions PE-1110

fedellen:
This PR preserves the `dataContentType` field on files when performing rename / move actions


jdaev (approved):
Looks good 👍🏽

karlprieb (approved):

PR #407, opened 2022-03-16
fix(ux and grammar fix at conflicting files): Rename modal label and …

thiagocarvalhodev:
- Change texts at conflicting files modal;
- Change button orders: from CANCEL -> SKIP -> REPLACE to SKIP -> CANCEL -> REPLACE
- Change buttons from ElevatedButton to TextButton (to remove the background)


fedellen (approved):
✅

karlprieb (approved):

PR #409, opened 2022-03-16
Pe 1121 fix buttons alignment

thiagocarvalhodev:
Align the buttons SKIP - CANCEL - REPLACE.

The TextButton width depends directly on the text inside. So, to align we need to pass a fixed width. 

Before:
image

Now:
image

Thanks, @agsuy


matibat (approved):
Looks good 👍🏼

karlprieb (approved):

PR #414, opened 2022-03-16
PE-663: Neutral Spanish translations

matibat:
**Notes**

- `app_en.arb` and `app_en_US.arb` have the same content
- `app_es.arb` and `app_es_ES.arb` have the same content

PR #416, opened 2022-03-18
PE-1130: Spanish of Spain locale `es_ES`

matibat:
We decided to separate this specific locale to move faster with the neutral Spanish at #414


matibat:
Outdated, and we decided to do not go for it atm

PR #426, opened 2022-03-29
PE-1154: Add the source_text to the ARB entries

matibat:


matibat:
We are discarding this ticket as the "source_text" has to be maintained at all files when the original translation changes

PR #435, opened 2022-04-05
PE-1125: TextPartitions (translations with different styles)

matibat:
- Implements the helper class `TextPartitions`, which is able to take a phrase and some segments of it, then return the strings that composes the phrase, index-based. 
- Then there is the `splitTranslationsWithMultipleStyles` method that depends on `TextPartitions`, and uses it to map the strings in the arguments to the specified widgets.
- The way of using it is to call `splitTranslationsWithMultipleStyles` passing the whole phrase as first argument, a widget factory for the default style as second argument, and a mapping of segments to widget factory as third argument. It will return an array of widgets that we can be directly used in the widget's build method.

```dart
splitTranslationsWithMultipleStyles(
    appLocalizations.aggreeToTerms_body,
    (String text) => Text(text),
    {
        [appLocalizations.aggreeToTerms_link]: (text) =>  TextButton(
            onPress: () => print('Open the terms and conditions'),
            child: Text(text),
        ),
    },
)

/// For the locale `en`, it will return
/// [
///     Text('I agree to the '),
///     TextButton(
///           onPress: () => print('Open the terms and conditions'),
///           child: Text('ArDrive terms of service and privacy policy'),
///       ),
/// ]


/// For the locale `ja`, it will return
/// [
///     Text('利用規約と '),
///     TextButton(
///           onPress: () => print('Open the terms and conditions'),
///           child: Text('ArDrive の プライバシーポリシー'),
///       ),
///     Text('に同意します'),
/// ]
```


lib/utils/text_partitions.dart
```diff
@@ -0,0 +1,45 @@
...
+    if (index > amount) {
+      throw Exception('Index overflow');
+    }
+    final sortedIndexes = indexes;
+    final start = sortedIndexes.elementAt(index);
+    final end = sortedIndexes.elementAt(index + 1);
+    return wholeText.substring(start, end);
+  }
+
+  void setSegment(String segment) {
```
matibat, line 34: I'd actually like to accept a RegExp instead of a String, or maybe to have an `setSegmentRegExp`.
Just thinking on possible future features we could need (e.g. to match what's after or before certain word).

matibat (commented):
We do need unit tests for right-to-left languages as well to ensure that the returned widgets preserve the order on those cases.

matibat:
> We do need unit tests for right-to-left languages as well to ensure that the returned widgets preserve the order on those cases.

I actually think we don't. Given that flutter already have support for `Directionality` that can be globally set making all inline elements to follow that rule.

lib/utils/text_partitions.dart
matibat: It is easy to apply, but we don't yet need the feature. So I won't apply it.

lib/pages/profile_auth/components/profile_auth_add_screen.dart
```diff
@@ -108,23 +109,26 @@ class ProfileAuthAddScreen extends StatelessWidget {
...
-                                        TextSpan(
-                                          // TODO replace at PE-1125
-                                          text: appLocalizationsOf(context)
-                                              .aggreeToTerms_link,
-                                          style: TextStyle(
-                                            decoration:
-                                                TextDecoration.underline,
-                                          ),
-                                        ),
-                                        TextSpan(text: '.'),
```
matibat, line 126: The reason why there's a change in all of the ARB files is because the point was previously hardcoded.

lib/pages/profile_auth/components/profile_auth_prompt_wallet_screen.dart
```diff
@@ -19,18 +20,21 @@ class ProfileAuthPromptWalletScreen extends StatelessWidget {
...
+                style: Theme.of(context).textTheme.headline6,
+              ),
+              parts: {
+                appLocalizationsOf(context).welcomeTo_main: (text) => Text(
+                      text,
+                      textAlign: TextAlign.center,
+                      style: Theme.of(context).textTheme.headline5,
+                    ),
+              },
+              separator: const SizedBox(height: 32),
```
matibat, line 37: Note we are using here a separator widget to preserve the same alignment.

lib/utils/split_localizations.dart
```diff
@@ -0,0 +1,37 @@
...
+    final segment = partitions.getSegment(segmentIndex);
+    final customMapper = parts[segment];
+    if (customMapper == null) {
+      /// It's not a custom part, use default mapper
+      mappedParts.add(defaultMapper(segment));
+    } else {
+      /// It's a given part, use specified mapper
+      mappedParts.add(customMapper(segment));
+    }
+    if (separator != null && segmentIndex + 1 < numSegments) {
```
matibat, line 31: If the `separator` was provided, then it's placed BETWEEN the returned elements. Without separator it returns `N` items, but when specified it returns `N + N-1` items.

test/utils/split_localizations.dart
```diff
@@ -0,0 +1,53 @@
...
+        parts: {
+          'ArDrive terms of service and privacy policy': anchorFactory,
+        },
+      );
+      expect(widgetsEN[0] is Text, true);
+      expect(widgetsEN[1] is TextButton, true);
+      expect(widgetsEN.length, 2);
+    });
+
+    test('puts separators between each part if specified', () {
```
matibat, line 37: Added the test for the separator feature.

matibat: Without separator it returns `N` items, but when specified it returns `N + N-1` items (`N-1` stands for the extra separator elements).

lib/utils/split_localizations.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:ardrive/utils/text_partitions.dart';
+
+typedef WidgetFactory<T> = T Function(String text);
+
+List<T> splitTranslationsWithMultipleStyles<T>({
+  required String originalText,
+  required WidgetFactory defaultMapper,
+  required Map<String, WidgetFactory> parts,
+  T? separator,
+}) {
```
matibat, line 10: I made the positional arguments here to be named arguments.

lib/utils/split_localizations.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:ardrive/utils/text_partitions.dart';
+
+typedef WidgetFactory<T> = T Function(String text);
+
+List<T> splitTranslationsWithMultipleStyles<T>({
```
matibat, line 5: Being the method generic gives us the opportunity to don't use `Widget`s only (such as in the case of `InlineSpan`s)

lib/utils/text_partitions.dart
```diff
@@ -0,0 +1,45 @@
...
+  final String wholeText;
+
+  TextPartitions({
+    required this.wholeText,
+  }) {
+    if (wholeText == '') {
+      throw Exception('Text is empty');
+    }
+    _indexes.add(0);
+    _indexes.add(wholeText.length);
```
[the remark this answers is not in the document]
matibat: Well...
- The `wholeText` is a String.
- Strings always have a number attribute: `.length` (i.e. not allowed to be null or undefined).
- In the conditional above we throw if the text is empty (e.i. when `wholeText.length === 0`).
  - It's because `_indexes` is a Set, meaning that if you add `0` twice, there will be a single `0` item on it, and we do need two at least.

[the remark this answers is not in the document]
karlprieb: I was expecting to have `wholeText.length ?? 0` instead of adding `0` and after `wholeText.length`, but you mean make sense...
So why do we set `0` first?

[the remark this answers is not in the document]
matibat: My attempt was to implement the concept of Partitions (the same used at calculus to apply the Riemann Integral) but applied to Strings instead of Real Numbers.
That is: Given a segment `[a, b]` within the domain (the indexes of `wholeText` in this case), a partition of that segment is a Set of finite points within the domain that contains `a` and `b` on it, which may include zero or more points of `(a, b)`. In this case, `a` is `0`, and `b` is `wholeText.length`.

Just to clarify conventions:
- `[a, b]` is a closed segment (includes `a` and `b`), 
- `(a, b)` is an open segment (excludes `a` and `b`).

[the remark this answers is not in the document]
karlprieb: Got it! Thank you for the explanation!

karlprieb (approved):

PR #443, opened 2022-04-11
PE-1297: Release v1.12.1

karlprieb:
**Release notes:**
- The app will no longer allow sibling files and folders to have conflicting names
- Adds internationalization support for Neutral Spanish Translations


matibat (approved):
The version is correct; the release notes look good to me!

PR #451, opened 2022-04-15
PE 1178/implement folder upload

jdaev:
Adds support for uploading folders.

Suggested Review Order:
- upload_form.dart
- upload_cubit.dart
Follow the upload flow in upload_cubit.dart while reviewing added utility functions.
- New File Classes in the blocs/upload/models folder
- Name changes and import changes that result from the above classes in the remaining files


lib/blocs/upload/upload_handles/file_data_item_upload_handle.dart
```diff
@@ -146,9 +146,11 @@ class DataItemUploadHandle implements UploadHandle, DataItemHandle {
...
   @override
-  Future<List<DataItem>> createDataItemsFromFileHandle() async {
+  Future<List<DataItem>> getDataItems() async {
     final dataItems = await prepareAndSignDataItems();
-    // Remove file data references
     return dataItems;
   }
+
+  @override
+  int get dataItemCount => 2;
```
thiagocarvalhodev, line 155: Would be better to provide a comment or name the constant `2` for something with more context

lib/blocs/upload/upload_handles/folder_data_item_upload_handle.dart
```diff
@@ -0,0 +1,97 @@
...
+
+  FolderDataItemUploadHandle({
+    required this.folder,
+    required this.arweave,
+    required this.wallet,
+    required this.targetDriveId,
+    this.driveKey,
+  });
+
+  Future<void> prepareAndsignFolderDataItem() async {
```
thiagocarvalhodev, line 51:
```suggestion
  Future<void> prepareAndSignFolderDataItem() async {
```

lib/blocs/upload/upload_handles/folder_data_item_upload_handle.dart
```diff
@@ -0,0 +1,97 @@
...
+      await driveDao.insertFolderRevision(
+        folderEntity.toRevisionCompanion(
+          performedAction: RevisionAction.create,
+        ),
+      );
+    });
+  }
+
+  @override
+  int get dataItemCount => 1;
```
thiagocarvalhodev, line 90: The same for the previous comment

lib/components/app_drawer/app_drawer.dart
```diff
@@ -221,6 +221,18 @@ class AppDrawer extends StatelessWidget {
...
+                          PopupMenuItem(
+                            enabled: state.hasWritePermissions,
+                            value: (context) => promptToUploadFolder(
+                              context,
+                              driveId: state.currentDrive.id,
+                              folderId: state.folderInView.folder.id,
+                            ),
+                            child: ListTile(
+                              enabled: state.hasWritePermissions,
+                              title: Text('Upload Folder'),
```
thiagocarvalhodev, line 233: @matibat , internationalization

lib/components/upload_form.dart
```diff
@@ -18,33 +20,97 @@ Future<void> promptToUploadFile(
...
+        barrierDismissible: false,
       ),
-      barrierDismissible: false,
-    ),
-  );
+    );
+  });
+}
+
+Future<void> promptToUploadFolder(
```
thiagocarvalhodev, line 66: The logic is almost the same of `promptToUploadFile`. Would be better to split this logic into another function

lib/blocs/upload/models/io_file.dart
```diff
@@ -0,0 +1,45 @@
...
+    this.parentFolderId,
+  ) : super(
+          name: file.name,
+          path: file.path,
+          lastModifiedDate: lastModifiedDate,
+          size: size,
+          parentFolderId: parentFolderId,
+        );
+
+  static Future<IOFile> fromXFile(XFile file, String parentFolderId) async {
```
thiagocarvalhodev, line 31: Why don't we use the factory constructor?

lib/blocs/upload/models/io_file.dart
```diff
@@ -0,0 +1,45 @@
...
+
+class IOFile extends UploadFile {
+  final XFile file;
+  @override
+  final DateTime lastModifiedDate;
+  @override
+  final int size;
+
+  @override
+  final String parentFolderId;
```
thiagocarvalhodev, line 16: We are overriding these variables but we are passing exactly the same for the super class. Is there a reason for that?

lib/blocs/upload/models/web_file.dart
```diff
@@ -0,0 +1,27 @@
+import 'dart:html';
+import 'dart:typed_data';
+
+import 'package:ardrive/blocs/upload/models/upload_file.dart';
+
+class WebFile extends UploadFile {
+  final File file;
+  @override
+  final String parentFolderId;
+  WebFile(this.file, this.parentFolderId)
```
thiagocarvalhodev, line 10: Why do we override the `parentFolderId`? As we are passing the `parentFolderId` to the super class, I don't see to override it. What about creating the `parentFolderId` only as a parameter?

thiagocarvalhodev (commented):
Very good work, @jdaev ! Just left a few comments.

lib/blocs/upload/models/io_file.dart
jdaev: This is the factory pattern. If you're wondering why I'm not using the factory keyword here it's because we can't have async constructors in dart.

lib/blocs/upload/models/io_file.dart
jdaev: Because UploadFile is an abstract base class that's being extended, not a superclass. It doesn't have an instance.

lib/blocs/upload/models/web_file.dart
jdaev: See above

lib/blocs/upload/upload_handles/folder_data_item_upload_handle.dart
jdaev: Thanks for catching that 👍🏽

lib/blocs/upload/upload_handles/folder_data_item_upload_handle.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/451/commits/7c3cd5f1eaed3ec928b6c083813ba87015bf958d)

lib/blocs/upload/upload_handles/file_data_item_upload_handle.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/451/commits/7c3cd5f1eaed3ec928b6c083813ba87015bf958d)

lib/blocs/upload/upload_handles/folder_data_item_upload_handle.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/451/commits/44cfae8061b2e48b67efc79a4eb114b3fd547795)

lib/components/app_drawer/app_drawer.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/451/commits/c19b6efca9128b4fc4ace7aca9d07cf1152016a5)

lib/blocs/upload/models/io_file.dart
thiagocarvalhodev: Ohh I see! Thanks!

thiagocarvalhodev (dismissed):
LGTM

karlprieb (dismissed):

lib/blocs/upload/models/upload_plan.dart
```diff
@@ -0,0 +1,69 @@
...
+    }
+    return bundle;
+  }
+
+  Future<void> createBundleHandlesFromDataItemHandles({
+    Map<String, FileDataItemUploadHandle> fileDataItemUploadHandles = const {},
+    Map<String, FolderDataItemUploadHandle> folderDataItemUploadHandles =
+        const {},
+  }) async {
+    // NOTE: Using maxFilesPerBundle since FileUploadHandles have 2 data items
```
fedellen, line 47: This comment assumption should probably be updated here as folder upload handles would have 1 data item

lib/blocs/upload/upload_cubit.dart
```diff
@@ -135,6 +144,68 @@ class UploadCubit extends Cubit<UploadState> {
...
+          .getSingleOrNull();
+      if (existingFolderId != null) {
+        folder.id = existingFolderId;
+        foldersToSkip.add(folder);
+      }
+      if (existingFileId != null) {
+        conflictingFolders.add(folder.name);
+      }
+      folder.parentFolderId =
+          folders[folder.parentFolderPath]?.id ?? _targetFolder.id;
```
fedellen, line 189: Don't we already check this value above? I think we could use `parentFolderId`, very small change though

It seems like we should just grab and preserve that folder as a nullable variable `final parentFolder = folders[folder.parentFolderPath]`. Then we wouldn't need to constantly check it for every operation here

lib/blocs/upload/upload_cubit.dart
```diff
@@ -135,6 +144,68 @@ class UploadCubit extends Cubit<UploadState> {
...
+      folder.path = folder.parentFolderPath.isNotEmpty
+          ? '${_targetFolder.path}/${folder.parentFolderPath}/${folder.name}'
+          : '${_targetFolder.path}/${folder.name}';
+    }
+    final filesToUpload = <WebFile>[];
+    files.forEach((file) {
+      final fileFolder = (file.path.split('/')..removeLast()).join('/');
```
fedellen, line 196: We should move this into a function with a name that describes this string manipulation or have a comment quickly describing the "why" here.

This is very short and concise, but we're essentially modifying the path three times on the same line of code.

lib/blocs/upload/upload_cubit.dart
```diff
@@ -135,6 +144,68 @@ class UploadCubit extends Cubit<UploadState> {
...
+      filesToUpload.add(
+        WebFile(
+          file.file,
+          folders[fileFolder]?.id ?? _targetFolder.id,
+        ),
+      );
+    });
+    folders.removeWhere((key, value) => foldersToSkip.contains(value));
+    foldersToUpload.addAll(folders);
+    return filesToUpload;
```
fedellen, line 206: I don't have full context here, but its a little awkward that we're returning the `filesToUpload` directly but the `foldersToUpload` are accumulating via a public class member

fedellen (approved):

matibat (commented):
Created a PR for the Spanish translation. I didn't review the whole PR yet

fedellen (approved):

PR #453, opened 2022-04-19
feat: add drive and file sizes PE-1134

jdaev:
Adds folder and drive sizes to the info sheet.


thiagocarvalhodev (dismissed):

thiagocarvalhodev (approved):

karlprieb (approved):

lib/l10n/app_en.arb
```diff
@@ -291,6 +291,10 @@
   "@fileSize": {
     "description": "The size of a file"
   },
+  "size": "Size",
```
karlprieb, line 294: @jdaev @matibat can we add spanish translation for this key?

karlprieb (approved):

PR #454, opened 2022-04-20
PE 1343 -Split getNewEntitiesForDrive function into smaller functions

thiagocarvalhodev:
Split the getNewEntitiesForDrive into smaller functions to change the sync algorithm order.

This function was split in two functions:

- getAllTransactionsFromDrive
- createDriveEntityHistoryFromTransactions

Most of the logic was reused.

### DIFF
--- a/lib/services/arweave/arweave_service.dart
+++ b/lib/services/arweave/arweave_service.dart
@@ -97,7 +97,148 @@ class ArweaveService {
     return query.data?.transaction;
   }
 
+  Future<List<DriveEntityHistory$Query$TransactionConnection$TransactionEdge>>
+      getAllTransactionsFromDrive(
+    String driveId, {
+    String? after,
+    int? lastBlockHeight,
+  }) async {
+    final transactionsEdges =
+        <DriveEntityHistory$Query$TransactionConnection$TransactionEdge>[];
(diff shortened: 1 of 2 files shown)


lib/services/arweave/arweave_service.dart
```diff
@@ -97,7 +99,150 @@ class ArweaveService {
...
+          ),
+        ),
+      );
+
+      transactionsEdges
+          .addAll(driveEntityHistoryQuery.data!.transactions.edges);
+
+      cursor = transactionsEdges.last.cursor;
+
+      if (driveEntityHistoryQuery.data!.transactions.edges.length <
```
karlprieb, line 130: we could use `hasNextPage` from the response to make it more readable

thiagocarvalhodev: Which object has this property?

karlprieb: the response from gql

thiagocarvalhodev: [bb68286](https://github.com/ardriveapp/ardrive-web/pull/454/commits/bb68286b091b40a560c6f4eb4408c9d96f4173c2)

PR #455, opened 2022-04-21
feat: get gatewayurl from cookie

jdaev:


lib/services/config/config_service.dart
```diff
@@ -10,10 +11,16 @@ class ConfigService {
 
   Future<AppConfig> getConfig() async {
     if (_config == null) {
-      final environment = kReleaseMode ? 'prod' : 'dev';
+      final environment = (kReleaseMode ? 'prod' : 'dev');
```
karlprieb, line 14: Why wrapping in parentheses?

jdaev: Leftover from something I tried. Will remove

lib/services/config/config_service.dart
```diff
@@ -10,10 +11,16 @@ class ConfigService {
...
       final configContent =
           await rootBundle.loadString('assets/config/$environment.json');
       _config = AppConfig.fromJson(json.decode(configContent));
+
+      // For development, set the gateway from cookie
+      final gatewayCookie = cookie.get('arweaveGatewayUrl');
```
karlprieb, line 20: That way we set `defaultArweaveGatewayUrl` using the json file and if we have the cookie set we set it again with the value from the cookie right?

Why not check if the cookie exist, if so set `defaultArweaveGatewayUrl` if not read and parse the json file?

jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/455/commits/e717172278ec8a77778f88a26b406f2e4e007d23)

karlprieb (approved):

fedellen (approved):

PR #456, opened 2022-04-21
PE-1342 Change the _syncDrive function from recursive to iterative

thiagocarvalhodev:
### Changes
- Change the function _syncDrive from recursive to iterative
- Paginate the process after getting the transactions with pages of 200 entities per time

### DIFF
--- a/lib/blocs/sync/sync_cubit.dart
+++ b/lib/blocs/sync/sync_cubit.dart
@@ -43,6 +43,7 @@ class SyncCubit extends Cubit<SyncState> {
   StreamSubscription? _arconnectSyncSub;
 
   DateTime? _lastSync;
+  late DateTime _initSync;
 
   SyncCubit({
     required ProfileCubit profileCubit,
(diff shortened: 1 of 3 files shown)


lib/blocs/sync/sync_cubit.dart
```diff
@@ -186,6 +188,7 @@ class SyncCubit extends Cubit<SyncState> {
       addError(err);
     }
     _lastSync = DateTime.now();
+    print(_initSync.difference(_lastSync!).inMilliseconds);
```
karlprieb, line 191: Do we want to log that?

karlprieb, line 191: Do we want to log that?

thiagocarvalhodev: For test purpose

karlprieb (approved):

jdaev (approved):

PR #459, opened 2022-04-21
PE-1345 Implement sync progress counting at SyncCubit

thiagocarvalhodev:


lib/services/arweave/arweave_service.dart
```diff
@@ -97,7 +99,150 @@ class ArweaveService {
...
+        ),
+      );
+
+      transactionsEdges
+          .addAll(driveEntityHistoryQuery.data!.transactions.edges);
+
+      cursor =
+          transactionsEdges.isNotEmpty ? transactionsEdges.last.cursor : null;
+
+      if (driveEntityHistoryQuery.data!.transactions.edges.length <
```
karlprieb, line 130: `hasNextPage`?

thiagocarvalhodev: Needs to merge with PE-1342.

PR #460, opened 2022-04-21
feat: add filesize util PE-1354

jdaev:
Adds filesize to string util that takes byte size as input and returns a string.


lib/utils/filesize.dart
```diff
@@ -0,0 +1,64 @@
+import 'dart:math';
+
+/// A method returns a human readable string representing a file _size
+String filesize(dynamic size, [int round = 2]) {
+  /** 
+   * [size] can be passed as number or as string
```
karlprieb, line 6: I think you can explain better that `size` is bytes

lib/pages/drive_detail/drive_detail_page.dart
```diff
@@ -27,6 +26,7 @@ import 'package:timeago/timeago.dart';
 import 'package:url_launcher/url_launcher.dart';
 
 import '../../../utils/app_localizations_wrapper.dart';
+import '../../../utils/filesize.dart';
```
karlprieb, line 29: why not `package:ardrive/utils/filesize.dart`?

jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/460/commits/b171249380d3bbadc8207664e599c9f949dfc938)

lib/utils/filesize.dart
jdaev: [Updated](https://github.com/ardriveapp/ardrive-web/pull/460/commits/b171249380d3bbadc8207664e599c9f949dfc938)

karlprieb (approved):

fedellen (approved):
🔢

PR #466, opened 2022-04-27
feat(internationalization es): adds keys for Spanish PE-1368

matibat:


lib/l10n/app_es.arb
```diff
@@ -323,6 +323,10 @@
   "@uploadDragAndDrop": {
     "description": "The action of uploading a file"
   },
+  "itemPreviewEmphasized": "VISTA PREVIA",
+  "@itemPreviewEmphasized": {
+    "description": "Item previews"
```
agsuy, line 328: Item preview, singular

agsuy (changes requested):
Left one comment

lib/l10n/app_es.arb
matibat: [4e1ab25](https://github.com/ardriveapp/ardrive-web/pull/466/commits/4e1ab25dd79ef34d1801eb933e3115a208ab1e31)

PR #471, opened 2022-04-29
feat: Display a sync progress indicator on web PE-1342

thiagocarvalhodev:
### Sync Progress Indicator
When syncing drives on ArDrive Web a modal with a progress animation is displayed. Since the sync process can take a long time it would be interesting to display the progress of the sync process, with this addition the user will be able to understand how much of the sync process is already done and how much it is still left to finish.

### Measuring the progress.
The progress is calculated by the following approach: the sync progress will be divided in two phases for each drive. The first one is for `fetching` transactions, let’s call `fetchPhase` and the second one to get the metadata for each entity and save them on database, called `getPhase`. The fetch phase will correspond 10% of the entire process and the left 90% is the get phase.

### Most significant code changes
- SyncCubit
- ArweaveService


lib/services/arweave/graphql/queries/DriveEntityHistory.graphql
```diff
@@ -9,6 +9,9 @@ query DriveEntityHistory($driveId: String!, $after: String, $lastBlockHeight: In
     after: $after,
     block: {min: $lastBlockHeight}
   ) {
+    pageInfo {
+      hasNextPage
+    }
```
fedellen, line 14: ❤️

lib/blocs/sync/sync_cubit.dart
```diff
@@ -168,23 +227,41 @@ class SyncCubit extends Cubit<SyncState> {
...
+          syncProgressController.add(syncProgress);
+        }
+        _syncProgress = _syncProgress.copyWith(
+            drivesSynced: _syncProgress.drivesSynced + 1);
+        syncProgressController.add(_syncProgress);
+      }));
+
+      print(
+          'Syncing drives finished.\nDrives quantity: ${_syncProgress.drivesCount}\n'
+          'The total progress was ${(_syncProgress.progress * 100).roundToDouble()}');
```
fedellen, line 247: There is a significant amount of logging added here. Are we bringing this through to production? 

I think its pretty useful. If we don't keep it, we might want a verbose logging cookie or something

lib/services/arweave/arweave_service.dart
```diff
@@ -96,7 +98,142 @@ class ArweaveService {
...
+    return DriveEntityHistory(
+      queryEdges.isNotEmpty ? queryEdges.last.cursor : null,
+      blockHistory.isNotEmpty ? blockHistory.last.blockHeight : lastBlockHeight,
+      blockHistory,
+    );
+  }
+
   /// Gets the entity history for a particular drive starting from the specified block height.
+  @Deprecated(
+      'Should use getAllTransactionsFromDrive and createDriveEntityHistoryFromTransactions to acomplish the same result.')
```
fedellen, line 236: I don't think we need to deprecate this. If its not being used in the app, just remove the code. It will be saved in git forever (:

lib/components/progress_dialog.dart
```diff
@@ -12,19 +12,45 @@ Future<void> showProgressDialog(BuildContext context, String title) =>
 
 class ProgressDialog extends StatelessWidget {
   final String title;
+  final Widget? percentageDetails;
+  final Widget? progressDescription;
+  final Widget? progressBar;
```
fedellen, line 17: Why are these all optional fields? Are they null while determining the data?

lib/blocs/sync/sync_cubit.dart
```diff
@@ -168,23 +227,41 @@ class SyncCubit extends Cubit<SyncState> {
...
+
       await createGhosts(ownerAddress: ownerAddress);
-      emit(SyncEmpty());
+
+      /// In order to have a smooth transition at the end.
+      await Future.delayed(Duration(milliseconds: 1000));
```
fedellen, line 252: Is this like a `sleep(1000)` in dart? e.g: this line of code just stops for 1 second

fedellen (dismissed):
Very good improvements here, LGTM!

A lot of the code seems to have been moved around rather than starting from scratch. We could work on separating some of this logic out into testable algorithms

But the UX improvement of the sync is huge, so we should get this moving forward to our users!

lib/blocs/sync/sync_cubit.dart
thiagocarvalhodev: Yep! Without it we can get 95%.. 99% and not touch 100% before close the modal. What do you think?

lib/components/progress_dialog.dart
thiagocarvalhodev: This component is used by others pages, like change folder name, move folder, etc... so, to don't change it in other pages, I made these widgets optionals. If we don't pass them, like in the move folder modal, this one will not change.

lib/blocs/sync/sync_cubit.dart
thiagocarvalhodev: Yep! I talked with @karlprieb and we have chosen to go with those logs.

lib/services/arweave/arweave_service.dart
thiagocarvalhodev: As this ticket has reached QA, I'll create a ticket to fix that!

fedellen (approved):

fedellen (approved):

PR #476, opened 2022-05-03
fix: Allow attaching drives on empty wallets without drives

thiagocarvalhodev:
If the user has no funds the button, `NEW` will be disabled and he won't be capable of attaching drives public drives. 

As no fund is required to do that, the button `NEW` will be enabled but the dropdown's options that are required to have funds will be disabled and will show a helpful message.

The disabled items:

- Create Drive 
- Create Folder
- Create Manifest
- Upload Files
- Upload Folder

### DIFF
--- a/lib/components/app_drawer/app_drawer.dart
+++ b/lib/components/app_drawer/app_drawer.dart
@@ -274,6 +274,7 @@ class AppDrawer extends StatelessWidget {
   PopupMenuEntry<Function> _buildNewFolderItem(
       context, DriveDetailLoadSuccess state, bool hasMinimumWalletBalance) {
     return _buildMenuItemTile(
+      context: context,
       isEnabled: state.hasWritePermissions && hasMinimumWalletBalance,
       itemTitle: appLocalizationsOf(context).newFolder,
       message: hasMinimumWalletBalance
@@ -290,6 +291,7 @@ class AppDrawer extends StatelessWidget {
   PopupMenuEntry<Function> _buildUploadFileItem(
       context, DriveDetailLoadSuccess state, bool hasMinimumWalletBalance) {
     return _buildMenuItemTile(
+      context: context,
       isEnabled: state.hasWritePermissions && hasMinimumWalletBalance,
       message: hasMinimumWalletBalance
           ? null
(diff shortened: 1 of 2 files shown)


karlprieb (approved):
Great clean up here 👍

lib/components/app_drawer/app_drawer.dart
```diff
@@ -179,201 +179,227 @@ class AppDrawer extends StatelessWidget {
...
+              child: Padding(
+                padding:
+                    const EdgeInsets.symmetric(horizontal: 8, vertical: 16),
+                child: BlocBuilder<DriveDetailCubit, DriveDetailState>(
+                  builder: (context, state) => PopupMenuButton<Function>(
+                    onSelected: (callback) => callback(context),
+                    itemBuilder: (context) => [
+                      if (state is DriveDetailLoadSuccess) ...{
+                        _buildNewFolderItem(context, state,
+                            profile.walletBalance >= minimumWalletBalance),
```
fedellen, line 198: We could extract this boolean to a constant above to avoid the repetition here

lib/components/app_drawer/app_drawer.dart
```diff
@@ -179,201 +179,227 @@ class AppDrawer extends StatelessWidget {
...
+                            profile.walletBalance >= minimumWalletBalance)
+                      },
+                    ],
+                    child: _buildNewButton(context),
+                  ),
+                ),
+              ),
+            ),
+          ),
+          if (profile.walletBalance < minimumWalletBalance) ...{
```
fedellen, line 223: We could use that boolean if created here as well, with `!hasMinBalance`

fedellen (approved):
Looking good, thanks for those improvements!

lib/components/app_drawer/app_drawer.dart
thiagocarvalhodev: Good point, @fedellen . We also don't have a high level class to represent our users' data, just the `ProfileLoggedIn`. I think that would be great to have a class to abstract the user in our system.

We could create something like `User` or just `Profile` .

lib/components/app_drawer/app_drawer.dart
thiagocarvalhodev: [Solved](https://github.com/ardriveapp/ardrive-web/commit/a97df5171e174d04666bbf3c5b9cb5ea4c87a7ec)

fedellen (approved):

PR #481, opened 2022-05-04
fix: When create another Drive, change the Drive view to this one PE-1421

thiagocarvalhodev:
### Problem
While you’re creating a new Drive and there is another created drive , and finishing the creation of the drive. The drive selected won’t be the newly created one.

### Fix
When create another Drive, change the Drive view to this one


fedellen (approved):
Nice! This is a long-standing issue

PR #484, opened 2022-05-06
fix(drag and drop): Dont use blob as path for DB entry PE-1445

fedellen:


karlprieb (approved):
LGTM

lib/utils/upload_plan_utils.dart
```diff
@@ -35,7 +35,12 @@ class UploadPlanUtils {
         private ? await driveDao.getDriveKey(targetDrive.id, cipherKey) : null;
     for (var file in files) {
       final fileName = file.name;
-      final filePath = '${targetFolder.path}/${file.path}';
+
+      // If path is a blob from drag and drop, use file name. Else use the path field from folder upload
+      final pathToUse =
+          file.path.split(':')[0] == 'blob' ? fileName : file.path;
+
+      final filePath = '${targetFolder.path}/$pathToUse';
```
fedellen, line 43:
```suggestion
      final filePath = '${targetFolder.path}/${file.getIdentifier()}';
```

PR #486, opened 2022-05-09
feat(tips on v2): Add community tip directly to v2 data transaction PE-1453

fedellen:
This PR adds the community tip to v2 data transactions and consolidates handling the community tip to the `PstService` class.

### DIFF
--- a/lib/blocs/upload/cost_estimate.dart
+++ b/lib/blocs/upload/cost_estimate.dart
@@ -1,14 +1,10 @@
 import 'package:ardrive/blocs/upload/models/upload_plan.dart';
 import 'package:ardrive/blocs/upload/upload_handles/bundle_upload_handle.dart';
 import 'package:ardrive/blocs/upload/upload_handles/file_v2_upload_handle.dart';
-import 'package:ardrive/entities/entity.dart';
 import 'package:ardrive/services/arweave/arweave.dart';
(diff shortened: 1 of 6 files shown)


jdaev (approved):
👍🏽

thiagocarvalhodev (approved):

lib/blocs/upload/cost_estimate.dart
```diff
@@ -52,60 +45,26 @@ class CostEstimate {
         arweaveService: arweaveService);
 
     final bundlePstFee = await pstService.getPSTFee(dataItemsCost);
+    final v2FilesPstFee = v2FilesUploadCost <= BigInt.zero
```
arielmelendez, line 48: Less than zero should arguably be an error.

arielmelendez: No need to change this though. I think it's good defense even if we expect something to throw upstream.

arielmelendez: Actually wouldn't it be cleaner just just let `await pstService.getPSTFee(v2FilesUploadCost);` handle this logic?

fedellen: This can be 0 when all files are bundled. It is just for estimating totalPrice

arielmelendez (approved):
Looks good. Issued one minor suggestion.

PR #487, opened 2022-05-09
Filter specific types of images to preview

thiagocarvalhodev:
Quoting [Flutter docs](https://api.flutter.dev/flutter/widgets/Image-class.html):

```
The following image formats are supported: JPEG, PNG, GIF, Animated GIF, WebP, Animated WebP, BMP, and WBMP.
We should filter by these type of images instead of just mime-types that starts with image/
```

According to [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml#image) we should filter the following mime-types:
- image/jpeg
- image/png
- image/gif
- image/webp
- image/bmp
- image/vnd.wap.wbmp


fedellen (approved):
👍

karlprieb (approved):

PR #498, opened 2022-05-16
Fix: Add retries on sync GraphQL queries context

thiagocarvalhodev:
Added retries for:

- Get all transactions;
- Get entities metadata;
- Get transaction confirmations.

Additional information about retry function:
```dart 
Duration delayFactor = const Duration(milliseconds: 200),
double randomizationFactor = 0.25,
Duration maxDelay = const Duration(seconds: 30),
int maxAttempts = 8,

### DIFF
--- a/lib/blocs/sync/sync_cubit.dart
+++ b/lib/blocs/sync/sync_cubit.dart
@@ -215,7 +215,6 @@ class SyncCubit extends Cubit<SyncState> {
         //
         // It also adds the encryption keys onto the drive models which isn't touched by the
         // later system.
-        //
         final userDriveEntities = await _arweave.getUniqueUserDriveEntities(
             profile.wallet, profile.password);
 
@@ -249,8 +248,8 @@ class SyncCubit extends Cubit<SyncState> {
             ),
             currentBlockheight: currentBlockHeight,
           ).handleError((error, stackTrace) {
-            print('Error syncing drive with id ${drive.id}');
-            print(onError.toString() + stackTrace.toString());
+            print(
+                'Error syncing drive with id ${drive.id}. Skipping sync to this one.\nException: ${onError.toString()}\nStackTrace: ${stackTrace.toString()}');
             addError(error!);
           }));
 
@@ -270,15 +269,19 @@ class SyncCubit extends Cubit<SyncState> {
       await createGhosts(ownerAddress: ownerAddress);
 
       /// In order to have a smooth transition at the end.
-      await Future.delayed(Duration(milliseconds: 1000));
+      await Future.delayed(const Duration(milliseconds: 1000));
 
       await Future.wait([
         if (profile is ProfileLoggedIn) _profileCubit.refreshBalance(),
         _updateTransactionStatuses(),
       ]);
-    } catch (err) {
+    } catch (err, stacktrace) {
+      if (err is FormatException) {
+        print('FormatException');
+      }
+      print(
+          'An error occurs while sync.\nError: ${err.toString()}\nStacktrace: ${stacktrace.toString()}');
       addError(err);
-      print('An error occurs while sync. Error: ' + err.toString());
     }
     _lastSync = DateTime.now();
     print('The sync process took: '
@@ -296,7 +299,6 @@ class SyncCubit extends Cubit<SyncState> {
 
   Future<void> createGhosts({String? ownerAddress}) async {
     //Finalize missing parent list
-
     for (final ghostFolder in ghostFolders.values) {
       final folderExists = (await _driveDao
               .folderById(
@@ -1131,11 +1133,14 @@ class SyncCubit extends Cubit<SyncState> {
 
   @override
   void onError(Object error, StackTrace stackTrace) {
+    print(
+        'An error occurs while sync.\nError: ${error.toString()}\nStacktrace${stackTrace.toString()}');
+    print('Emiting SyncFailure state');
     emit(SyncFailure(error: error, stackTrace: stackTrace));
-    super.onError(error, stackTrace);
-    emit(SyncIdle());
 
-    print('Failed to sync: $error $stackTrace');
+    print('Emiting SyncIdle state');
+    emit(SyncIdle());
+    super.onError(error, stackTrace);
   }
 
   @override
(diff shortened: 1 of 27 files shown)

### CODE REVIEW

File: lib/blocs/sync/sync_cubit.dart [hunk 1]
```diff
@@ -249,8 +248,8 @@ class SyncCubit extends Cubit<SyncState> {
             ),
             currentBlockheight: currentBlockHeight,
           ).handleError((error, stackTrace) {
-            print('Error syncing drive with id ${drive.id}');
-            print(onError.toString() + stackTrace.toString());
+            print(
+                'Error syncing drive with id ${drive.id}. Skipping sync to this one.\nException: ${onError.toString()}\nStackTrace: ${stackTrace.toString()}');
```
fedellen, line 252:
```suggestion
                'Error syncing drive with id ${drive.id}. Skipping sync on this drive.\nException: ${onError.toString()}\nStackTrace: ${stackTrace.toString()}');
```

lib/blocs/sync/sync_cubit.dart
```diff
@@ -270,13 +269,18 @@ class SyncCubit extends Cubit<SyncState> {
...
       ]);
-    } catch (err) {
+    } catch (err, stacktrace) {
+      if (err is FormatException) {
+        print('FormatException');
+      }
+      print(
+          'An error occurs while sync.\nError: ${err.toString()}\nStacktrace: ${stacktrace.toString()}');
       addError(err);
       print('An error occurs while sync. Error: ' + err.toString());
```
fedellen, line 285:
```suggestion
          'An error has occurred while syncing.\nError: ${err.toString()}\nStacktrace: ${stacktrace.toString()}');
      addError(err);
      print('An error has occurred while syncing. Error: ' + err.toString());
```

fedellen: Is this printing the same log twice here?

lib/blocs/sync/sync_cubit.dart
```diff
@@ -1131,11 +1134,14 @@ class SyncCubit extends Cubit<SyncState> {
 
   @override
   void onError(Object error, StackTrace stackTrace) {
+    print(
+        'An error occurs while sync.\nError: ${error.toString()}\nStacktrace${stackTrace.toString()}');
```
fedellen, line 1138: Same grammer suggestion as above

fedellen: Consider a `_printSyncError(Object error, StackTrace stackTrace)` method if this is repeated elsewhere

karlprieb:
What are the defaults for timeout and max retries of the retry package that we started to use?

lib/blocs/sync/sync_cubit.dart
thiagocarvalhodev: Yes! Good catch!

lib/services/arweave/arweave_service.dart
```diff
@@ -145,7 +155,14 @@ class ArweaveService {
...
       entityTxs.map((e) async {
-        return client.api.get(e.id);
+        return retry(
+          () => client.api.get(e.id),
+          onRetry: (exception) {
+            print(
+              'Retrying for get ${e.id} metadata on exception ${exception.toString()}',
+            );
+          },
+        );
```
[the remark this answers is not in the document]
thiagocarvalhodev: Yes. As we aren't handling each exception in a proper way, I guess printing the exception with `exception.toString();` should be enough. What do you think?

thiagocarvalhodev:
> What are the defaults for timeout and max retries of the retry package that we started to use?

Duration delayFactor = const Duration(milliseconds: 200),
double randomizationFactor = 0.25,
Duration maxDelay = const Duration(seconds: 30),
int maxAttempts = 8,

lib/services/arweave/arweave_service.dart
```diff
@@ -25,12 +27,15 @@ class ArweaveService {
...
       : _gql = ArtemisClient('${client.api.gatewayUrl.origin}/graphql') {
     unawaited(initializeMempoolStream());
+    graphQLRetry = GraphQLRetry(_gql);
   }
 
   int bytesToChunks(int bytes) {
     return (bytes / byteCountPerChunk).ceil();
   }
 
+  late GraphQLRetry graphQLRetry;
```
fedellen, line 37: Why is this instantiated as a `late` instead of `final` like the `_gql` member? They seem to be similarly required and handled in the constructor

lib/services/arweave/arweave_service.dart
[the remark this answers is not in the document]
fedellen: It depends on the shape of `exception` there. We are probably good

lib/utils/graphql_retry.dart
```diff
@@ -0,0 +1,39 @@
...
+      return queryResponse;
+    } catch (e) {
+      late Object exception;
+      if (e.toString().contains('FormatException')) {
+        exception = const FormatException('Returned data is not a valid JSON.');
+      } else {
+        exception = e;
+      }
+      print(
+          'Fatal error while querying ${query.operationName}. Exceed the number of retries. Exception ${exception.toString()}');
```
fedellen, line 35: grammer nit: `Exceeded the number of retries.` or `Number of retries has been exceeded.`

fedellen (approved):
Left a couple non-blocking nits, LGTM!

lib/services/arweave/arweave_service.dart
thiagocarvalhodev: To be final we should initialize on a declaration or through the constructor. As the `graphQLRetry` needs `_gql` we, and `_gql` is initialized on the constructor (not through dependency injection) we can not initialize both through the constructor.

thiagocarvalhodev: Totally agree with change it to privade.

lib/utils/graphql_retry.dart
thiagocarvalhodev: [758f1c0](https://github.com/ardriveapp/ardrive-web/pull/498/commits/758f1c0ebfb88f5ed4790d78ff42aa862b98d3ae)

thiagocarvalhodev: Thanks, @fedellen !

PR #499, opened 2022-05-17
feat: use sandboxed domain to fetch tx PE-1440

jdaev:
Generated and uses sandboxed domains to get tx data.


lib/services/arweave/arweave_service.dart
```diff
@@ -116,7 +116,7 @@ class ArweaveService {
     final queryEdges = driveEntityHistoryQuery.data!.transactions.edges;
     final entityTxs = queryEdges.map((e) => e.node).toList();
     final responses = await Future.wait(
-      entityTxs.map((e) => client.api.get(e.id)),
+      entityTxs.map((e) => client.api.getSandboxedTx(e.id)),
```
fedellen, line 119: We discussed possibly having a fallback for in case the sandbox algorithm fails to produce a viable subdomain

jdaev: In my experience the gateway redirects automatically if the sandbox doesn't match

thiagocarvalhodev (dismissed):
LGTM

matibat (dismissed):
LGTM

PR #502, opened 2022-05-18
Fix: Handle negative numbers getting the current block height

thiagocarvalhodev:
UI Crashes and enters an infinite loop on negative blockheights.

For unknown reasons if the query to Arweave for the latest blockheight fails, we could end with a negative blockheight to be used during sync, which crashes the Sync progress UI and leads to an infinite loop.

### DIFF
--- a/lib/blocs/sync/sync_cubit.dart
+++ b/lib/blocs/sync/sync_cubit.dart
@@ -17,6 +17,7 @@ import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:meta/meta.dart';
 import 'package:moor/moor.dart';
+import 'package:retry/retry.dart';
 
 part 'sync_state.dart';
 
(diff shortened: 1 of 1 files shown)


lib/blocs/sync/sync_cubit.dart
```diff
@@ -295,6 +303,16 @@ class SyncCubit extends Cubit<SyncState> {
     }
   }
 
+  Future<int> getCurrentBlockHeight() async {
+    final currentBlockHeight = await arweave.getCurrentBlockHeight();
+
+    if (currentBlockHeight < 0) {
+      throw Exception(
+          'The current block height is negative. It should be equal or greater than 0.');
```
fedellen, line 311: We could probably provide more type safety than this. What would happen if a gateway returned a block height as a float? e.g `685213.23`

Also we should include the currentBlockHeight in the exception for more debugging information

thiagocarvalhodev: >  Also we should include the currentBlockHeight in the exception for more debugging information

[34b2d03](https://github.com/ardriveapp/ardrive-web/pull/502/commits/34b2d03f0c57253231052eb8de1c2afc062cc61a)

fedellen (approved):

lib/blocs/sync/sync_cubit.dart
```diff
@@ -295,6 +303,16 @@ class SyncCubit extends Cubit<SyncState> {
     }
   }
 
+  Future<int> getCurrentBlockHeight() async {
+    final currentBlockHeight = await arweave.getCurrentBlockHeight();
+
+    if (currentBlockHeight < 0) {
+      throw Exception(
+          'The current block height $currentBlockHeight is negative. It should be equal or greater than 0.');
```
matibat, line 311:
```suggestion
    if (Number.isInteger(currentBlockHeight) && currentBlockHeight < 0) {
      throw Exception(
          'The current block height $currentBlockHeight should be an integer equal or greater than 0.');
```

matibat (approved):
We could check if the number is integer as well. But you are good as it is

PR #506, opened 2022-05-19
Fix: Skip files above limit and upload remaining ones.

thiagocarvalhodev:
Skip files above limit and upload remaining ones.

### DIFF
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -209,7 +209,7 @@ class UploadCubit extends Cubit<UploadState> {
 
   /// If `conflictingFileAction` is null, means that had no conflict.
   Future<void> prepareUploadPlanAndCostEstimates({
-    ConflictingFileActions? conflictingFileAction,
+    UploadActions? uploadAction,
   }) async {
     final profile = _profileCubit.state as ProfileLoggedIn;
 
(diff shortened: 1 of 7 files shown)


lib/blocs/upload/upload_cubit.dart
```diff
@@ -226,21 +226,24 @@ class UploadCubit extends Cubit<UploadState> {
...
-        isPrivate: _targetDrive.isPrivate,
-      ));
-      return;
+    } else if (uploadAction == UploadActions.SkipBigFiles) {
+      _removeBigFiles();
+    } else {
+      final tooLargeFiles = [
+        for (final file in files)
+          if (file.size > sizeLimit) file.name
+      ];
```
karlprieb, line 237: Can we use a filter like `where` here?

thiagocarvalhodev: We can. But I think that the current way is more readble than 
files.where((element) => element.size > sizeLimit).map((e) => e.name);
What do you think?

karlprieb: I still think it's more readable, but maybe just because I'm not used to Dart syntax :)
Let's keep your implementation as it will just iterate over the list once.

karlprieb (approved):

lib/blocs/upload/upload_cubit.dart
thiagocarvalhodev: Sure!

matibat (approved):
LGTM

lib/blocs/upload/upload_state.dart
```diff
@@ -52,9 +52,12 @@ class UploadFolderNameConflict extends UploadFileConflict {
 
 class UploadFileTooLarge extends UploadState {
   final List<String> tooLargeFileNames;
+  final bool hasFilesToUpload;
   final bool isPrivate;
   UploadFileTooLarge(
-      {required this.tooLargeFileNames, required this.isPrivate});
+      {required this.tooLargeFileNames,
+      required this.isPrivate,
+      required this.hasFilesToUpload});
```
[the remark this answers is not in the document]
thiagocarvalhodev: [39c97f5](https://github.com/ardriveapp/ardrive-web/commit/39c97f535544f0e0a92833db0cb1b0b10a5d30b8)

matibat (approved):

PR #511, opened 2022-05-23
chore(arweave-dart) Change ref to dev

thiagocarvalhodev:


agsuy (approved):
Uses latest arweave-dart

PR #531, opened 2022-05-31
feat: check for files above limit first and then check for conflicts

thiagocarvalhodev:
Changes the order of validation from:

**name conflicting → check file size** 

to

**check file size → name conflicting**

### DIFF
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -17,7 +17,7 @@ import 'enums/conflicting_files_actions.dart';
 
 part 'upload_state.dart';
 
-final privateFileSizeLimit = 104857600;
+const privateFileSizeLimit = 104857600;
 final publicFileSizeLimit = 1.25 * math.pow(10, 9);
 final filesNamesToExclude = ['.DS_Store'];
 
@@ -70,11 +70,18 @@ class UploadCubit extends Cubit<UploadState> {
     emit(UploadPreparationInitialized());
   }
 
+  Future<void> checkConflicts() async {
+    if (uploadFolders) {
+      return await checkConflictingFolders();
+    }
+
+    await checkConflictingFiles();
+  }
+
   /// Tries to find a files that conflict with the files in the target folder.
   ///
   /// If there's one, prompt the user to upload the file as a version of the existing one.
   /// If there isn't one, prepare to upload the file.
-
   Future<void> checkConflictingFolders() async {
     emit(UploadPreparationInProgress());
     if (uploadFolders) {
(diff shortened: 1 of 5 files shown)


karlprieb (dismissed):

lib/components/upload_form.dart
```diff
@@ -219,10 +221,10 @@ class UploadForm extends StatelessWidget {
                 ),
                 if (state.hasFilesToUpload)
                   TextButton(
-                    onPressed: () => context
-                        .read<UploadCubit>()
-                        .prepareUploadPlanAndCostEstimates(
-                            uploadAction: UploadActions.SkipBigFiles),
+                    onPressed: () {
+                      context.read<UploadCubit>().removeBigFiles();
+                      context.read<UploadCubit>().checkConflicts();
```
fedellen, line 226: This second call is async, is it okay to not await here?

test/blocs/upload_cubit_test.dart
```diff
@@ -374,9 +374,10 @@ void main() {
             tNoConflictingFiles..add(tTooLargeFile));
       },
       act: (cubit) async {
+        // TODO(@thiagocarvalhodev): Review
         await cubit.startUploadPreparation();
         await cubit.prepareUploadPlanAndCostEstimates(
-            uploadAction: UploadActions.SkipBigFiles);
+            uploadAction: UploadActions.Skip);
```
fedellen, line 380: We didn't really get any test coverage over the new logic here. UploadReady will have the same behavior as before right?

lib/blocs/upload/upload_cubit.dart
```diff
@@ -191,7 +211,7 @@ class UploadCubit extends Cubit<UploadState> {
           : '${_targetFolder.path}/${folder.name}';
     }
     final filesToUpload = <WebFile>[];
-    files.forEach((file) {
+    for (var file in files) {
```
fedellen, line 214: Is this a performance increase for Dart?

fedellen (approved):
Left some non-blocking questions. Looking great!

lib/components/upload_form.dart
thiagocarvalhodev: It's okay! As this function changes the states of this page, we don't need to wait as at every state change this page will be affected by the side effects.

test/blocs/upload_cubit_test.dart
thiagocarvalhodev: Yeah. Those tests should be refactored. To test those functions today is currently not possible due to the upload_cubit depends of other classes that import `dart:html` directly, so the tests not compile. If you are able to compile, please let me know!

lib/blocs/upload/upload_cubit.dart
thiagocarvalhodev: It is not, it is just to fix the linter rule: `Avoid using forEach with a function literal.` 

But I agree, it should be a final variable as we aren't changing it!

PR #532, opened 2022-05-31
PE 1621/select other drive after detach

jdaev:
Fixes contents of detached drive remaining on the drive view.
Also fixes the uncaught console error.

### DIFF
--- a/lib/blocs/drives/drives_cubit.dart
+++ b/lib/blocs/drives/drives_cubit.dart
@@ -92,15 +92,16 @@ class DrivesCubit extends Cubit<DrivesState> {
               : null;
       if (firstOrNullDrive != null) {
         emit(state.copyWith(selectedDriveId: firstOrNullDrive));
+        return;
       }
     }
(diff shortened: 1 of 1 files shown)


lib/blocs/drives/drives_cubit.dart
```diff
@@ -92,15 +92,16 @@ class DrivesCubit extends Cubit<DrivesState> {
...
       }
     }
     emit(DrivesLoadedWithNoDrivesFound(canCreateNewDrive: canCreateNewDrive));
   }
 
   Future<void> detachDrive(DriveID driveId) async {
-    await _driveDao.deleteDriveById(driveId: driveId);
     _resetDriveSelection(driveId);
-    emit(state);
+    await Future.delayed(const Duration(seconds: 1));
```
thiagocarvalhodev, line 103: Why this delay?

thiagocarvalhodev: Thanks for the explanation, Javed! Do you think that we can get any race conditions?

karlprieb (approved):

PR #536, opened 2022-06-04
refactor: remove mempool stream PE-1579

jdaev:
Removes cached mempool stream.

### DIFF
--- a/lib/pages/congestion_warning_wrapper.dart
+++ b/lib/pages/congestion_warning_wrapper.dart
@@ -10,7 +10,7 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 Future<void> showCongestionDependentModalDialog(
     BuildContext context, Function() showAppDialog) async {
   final warnAboutCongestion =
-      await context.read<ArweaveService>().getCachedMempoolSize() >
+      await context.read<ArweaveService>().getMempoolSizeFromArweave() >
           mempoolWarningSizeLimit;
   return await showModalDialog(context, () async {
     if (warnAboutCongestion) {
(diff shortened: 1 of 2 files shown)


fedellen (approved):
👍

fedellen:
Whats the status here? We should push this. It'll be a nice performance boost

jdaev:
Sorry about that. I did this on a weekend and forgot about it.

karlprieb (approved):

matibat (approved):

thiagocarvalhodev (dismissed):

jdaev:
🦖 👍🏽

lib/pages/congestion_warning_wrapper.dart
```diff
@@ -9,9 +9,17 @@ import 'package:flutter_bloc/flutter_bloc.dart';
...
-          mempoolWarningSizeLimit;
+  late bool warnAboutCongestion;
+
+  try {
+    final mempoolSize =
+        await context.read<ArweaveService>().getMempoolSizeFromArweave();
+    
+   warnAboutCongestion = mempoolSize > mempoolWarningSizeLimit;
+  } catch (e) {
+    warnAboutCongestion = false;
```
matibat, line 20: Hmmm so now we set it to false when the request fails? Shouldn't we retry? :opionion:

matibat (approved):
Actually I think we are not changing the behavior. That prev comment shouldn't be a blocker

lib/pages/congestion_warning_wrapper.dart
thiagocarvalhodev: We could, but It is not critical. Today we have the GraphQLRetry class to perform retries. When implemented a similar class to all requests, we can address it for sure.

matibat: Yeah, you are good to go!. Thanks @thiagocarvalhodev 😎

PR #537, opened 2022-06-06
Release v1.20.0

thiagocarvalhodev:
#### Release notes
- Logged in user are now able to detach an attached drive


matibat (approved):
LGTM!

PR #542, opened 2022-06-13
PE-1685: Adds missing strings to the Mandarin ARB file

matibat:


matibat:
The PR is displaying commits that are already present in the base (PE-1463). Maybe because I merged them here before merging them to the base

matibat:
Replaced by #543

PR #543, opened 2022-06-13
feat(localization zh): adds the missing strings from English PE-1685

matibat:


lib/l10n/app_zh.arb
```diff
@@ -1112,7 +1142,7 @@
     "description": "Show that needs funds to create a Drive"
   },
   "insufficientFundsForCreateAFolder": "你没有足够的余额来创建文件夹. 请到账户的充值页面进行充值.",
-  "@insufficientFundsForCreateADrive": {
+  "@insufficientFundsForCreateAFolder": {
```
fedellen, line 1145: Was this a concern in any other arb files?

fedellen (approved):

lib/l10n/app_zh.arb
matibat: This was fixed in a PR of @thiagocarvalhodev (already present at dev, both for English and Spanish)

PR #544, opened 2022-06-14
PE-1664: Persist the Feedback Survey State

matibat:


test/utils/key_value_store_test.dart
```diff
@@ -0,0 +1,47 @@
+import 'package:ardrive/utils/key_value_store.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('KeyValueStore class setup', () {
+    test('throws if not setted up', () async {
```
fedellen, line 7: nit:
```suggestion
  group('KeyValueStore class', () {
    test('throws if used before the setup method is called', () async {
```

test/utils/key_value_store_test.dart
```diff
@@ -0,0 +1,47 @@
...
+        currentValue = store.getBool('isItTrue');
+        expect(currentValue, true);
+      });
+    });
+
+    group('remove method', () {
+      test('returns true when sucessfully removed', () async {
+        final success = await store.remove('isItTrue');
+        expect(success, true);
+      });
```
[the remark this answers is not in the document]
fedellen: Or is that not whats expected. It seems to be false by default? This suite of tests doesn't make it clear how we're meant to interact here

test/blocs/feedback_survey_cubit_test.dart
```diff
@@ -53,12 +54,37 @@ void main() {
...
     );
   });
+
+  group('FeedbackSurveyCubit preferences', () {
+    final store = KeyValueStore();
+    late FeedbackSurveyCubit feedbackCubit;
+
+    setUp(() async {
+      registerFallbackValue(SyncStatefake());
+      registerFallbackValue(ProfileStatefake());
```
fedellen, line 70: camelCase nit:
```suggestion
      registerFallbackValue(SyncStateFake());
      registerFallbackValue(ProfileStateFake());
```

needs to be changed elsewhere though

lib/utils/key_value_store.dart
```diff
@@ -0,0 +1,28 @@
...
+  SharedPreferences get _preferences {
+    if (_prefs == null) {
+      throw Exception('You have to call setup() first!');
+    }
+    return _prefs!;
+  }
+
+  Future<void> setup({SharedPreferences? instance}) async {
+    _prefs = _prefs ?? instance ?? await SharedPreferences.getInstance();
+  }
```
fedellen, line 15: Shouldn't this be on constructor? Or is the sharedPreferences library not loaded up yet?

lib/blocs/feedback_survey/feedback_survey_cubit.dart
```diff
@@ -1,33 +1,45 @@
...
-    hasAlreadyBeenOpened = true;
   }
 
   void closeRemindMe() {
     emit(FeedbackSurveyRemindMe(isOpen: false));
   }
 
-  void leaveFeedback() {
+  Future<void> leaveFeedback() async {
+    await _store.setup();
```
fedellen, line 31: Do we need to `setup()` store on every call?

fedellen (commented):
Glad we got some cross platform basic key value persistence going here

Left a few questions, mostly about the wrapper class and its tests. Also, we could probably change the title here to `PE-1664: Persist the Feedback Survey State` or similar. `Read & write the pref` is a little obscure

Looking good!

lib/utils/key_value_store.dart
matibat: Hmm I cannot await in the constructor :c

test/utils/key_value_store_test.dart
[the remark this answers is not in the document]
matibat: Yes! We do expect the key to be false after removed
d034c63ad0e6c502750290084de1d8319ec64344

test/blocs/feedback_survey_cubit_test.dart
matibat: Nice! 3976169048c805b8fcbd24f05a87c04b8a0ad094

lib/blocs/feedback_survey/feedback_survey_cubit.dart
matibat: The method takes effect only the first time it's called. So it's safe to call it from all of those places. Again, my issue here is that I cannot setup the lib API in the constructor because it's a Future

lib/utils/key_value_store.dart
fedellen: True. Maybe setup should just be internalized then. We don't really need to expose it. 

Another approach could be to have a non-async constructor that takes a SharedPreference for that part of functionality (if we're using it), then just have `setup` become `_setup` that gets called on every interaction.

test/utils/key_value_store_test.dart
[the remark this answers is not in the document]
fedellen: Looks like you accidentally pushed some description undos there.

That still doesn't seem complete there. If the default is false, we would want to turn the boolean true, assert that it is true, then `remove()` it, then assert that it is false.

Also the description should be updated once expanded

lib/utils/key_value_store.dart
fedellen: In TypeScript we usually handle async constructors with a static method like `MyClass.from()` or similar. @thiagocarvalhodev how would you normally handle this in Dart?

karlprieb: I like @fedellen approach for that. You can use this stack overflow answer as base for your implementation https://stackoverflow.com/a/59304510

lib/utils/key_value_store.dart
```diff
@@ -0,0 +1,28 @@
...
+
+  Future<bool> putBool(String key, bool value) {
+    return _preferences.setBool(key, value);
+  }
+
+  bool getBool(String key) {
+    return _preferences.getBool(key) ?? false;
```
karlprieb, line 22: should we return `false` when the `key` doesn't exist? I don't think we should do that.

lib/blocs/feedback_survey/feedback_survey_cubit.dart
karlprieb: Can't we have a private method that call it from the constructor?

test/utils/key_value_store_test.dart
fedellen: This one got unresolved in a follow up commit

fedellen: But this will change if we move the setup implementation to the static constructor

lib/utils/key_value_store.dart
matibat: Well, I tend to agree... At some point we MIGHT have a feature requiring us to check for true vs false vs null

matibat: The feat: 4e93ae762da5d40967f5a22a9a0d9b4b9357fc3b
The test: 09595a8b4b4fa4e8f81560c469170bc116876fdf

lib/utils/key_value_store.dart
thiagocarvalhodev: I would recommend creating a static function and also a private constructor to avoid call this constructor without making the setup first.


```dart
class KeyValueStore {
  KeyValueStore._create();
   
   static Future<KeyValueStore> getInstance() {
     if(_prefs == null)
       _prefs = await SharedPreferences.getInstance();
     return KeyValueStore._create();
   } 
   
   SharedPreferences? _prefs;
}
```

And you can use it like that

```dart
final store = await KeyValueStore.getInstance();

 // do something with store
 ```

File: lib/utils/key_value_store.dart [hunk 7]
```diff
@@ -0,0 +1,28 @@
+import 'package:shared_preferences/shared_preferences.dart';
+
+class KeyValueStore {
```
thiagocarvalhodev, line 3: Should not this class be an abstract class? We are doing an abstraction here for the shared preferences but for every class that has it as a dependency they'll be strongly coupled with `shared_preferences` package which is not our goal.

I recommend changing this class for an abstract class and or adding a interface for it.

For instance:

```dart
abstract class KeyValueStore {  
   Future<bool> putBool(String key, bool value);
   bool? getBool(String key);
   Future<bool> remove(String key);
 }
 
 class LocalKeyValueStore implements KeyValueStore {
   // implement methods with `shared_preferences`
 }
 ```
 
 In this way, we can only define the interface on our project and use different implementation as needed.

File: test/test_utils/fakes.dart [hunk 8]
```diff
@@ -1,8 +1,8 @@
...
-
-class SyncStatefake extends Fake implements SyncState {}
-
-class ProfileStatefake extends Fake implements ProfileState {}
-
-class DrivesStatefake extends Fake implements DrivesState {}
+import 'package:ardrive/blocs/blocs.dart';
+import 'package:mocktail/mocktail.dart';
+
+class SyncStateFake extends Fake implements SyncState {}
```
thiagocarvalhodev, line 4: Very good catch!

lib/components/file_share_dialog.dart
```diff
@@ -99,9 +99,9 @@ class _FileShareDialogState extends State<FileShareDialog> {
           actions: [
             if (state is FileShareLoadSuccess)
               ElevatedButton(
-                onPressed: () {
+                onPressed: () async {
                   Navigator.pop(context);
-                  context.read<FeedbackSurveyCubit>().openRemindMe();
+                  await context.read<FeedbackSurveyCubit>().openRemindMe();
```
thiagocarvalhodev, line 104: Don't need the `await` keyword. This await will just prevent from keep following the current flow inside the function `onPressed`. As we aren't doing anything after, we call safely remove that

lib/components/feedback_survey.dart
```diff
@@ -114,9 +114,9 @@ class FeedbackSurveyModal extends StatelessWidget {
                                 style: ElevatedButton.styleFrom(
                                   minimumSize: const Size.fromHeight(50), // NEW
                                 ),
-                                onPressed: () {
+                                onPressed: () async {
                                   launchSurveyURL();
-                                  context
+                                  await context
```
thiagocarvalhodev, line 119: Same of the previous comment.

lib/components/feedback_survey.dart
```diff
@@ -101,8 +101,8 @@ class FeedbackSurveyModal extends StatelessWidget {
                                     minimumSize:
                                         const Size.fromHeight(50), // NEW
                                   ),
-                                  onPressed: () {
-                                    context
+                                  onPressed: () async {
+                                    await context
```
thiagocarvalhodev, line 105: Same of the previous comment.

lib/components/drive_share_dialog.dart
```diff
@@ -98,9 +98,9 @@ class _DriveShareDialogState extends State<DriveShareDialog> {
           actions: [
             if (state is DriveShareLoadSuccess)
               ElevatedButton(
-                onPressed: () {
+                onPressed: () async {
                   Navigator.pop(context);
-                  context.read<FeedbackSurveyCubit>().openRemindMe();
+                  await context.read<FeedbackSurveyCubit>().openRemindMe();
```
thiagocarvalhodev, line 103: Same of the previous comment.

lib/components/create_manifest_form.dart
```diff
@@ -53,7 +53,7 @@ class CreateManifestForm extends StatelessWidget {
             state is CreateManifestPrivacyMismatch) {
           Navigator.pop(context);
           Navigator.pop(context);
-          context.read<FeedbackSurveyCubit>().openRemindMe();
+          await context.read<FeedbackSurveyCubit>().openRemindMe();
```
thiagocarvalhodev, line 56: Same of the previous comment.

lib/blocs/feedback_survey/feedback_survey_cubit.dart
thiagocarvalhodev: With my suggestion [here](https://github.com/ardriveapp/ardrive-web/pull/544/files#r897961195), you can do it like this:

```dart
 Future<void> leaveFeedback() async {
     final store = KeyValueStore.getInstance();
     await store(dontRemindMeAgainKey, true);
     emit(FeedbackSurveyDontRemindMe(isOpen: false));
   }
```
But as we are passing the `_store` on constructor, which is correct, we should await to get the instance before injecting the `KeyValueStore` on `FeedbackSurveyCubit`.

It will bring more complexity to the code. 

I would make this class a `Singleton` class, and I don't see any problems using `shared_prefereces` once it is a singleton too.

Then, we can have a function to load some core dependencies, like we do here

```dart
late ConfigService configService;
late AppConfig config;
late ArweaveService arweave;

void main() async {
  WidgetsFlutterBinding.ensureInitialized();

  configService = ConfigService();
  config = await configService.getConfig();

  arweave = ArweaveService(
      Arweave(gatewayUrl: Uri.parse(config.defaultArweaveGatewayUrl!)));
  refreshHTMLPageAtInterval(Duration(hours: 12));
  runApp(App());
}
```

We can implement a new function called `loadCoreDependencies()` which will load those above

```dart
loadCoreDependencies() {
  /// load config service into the global variable
  configService = ConfigService();
  /// load config into the global variable
  config = await configService.getConfig();
  /// load arweave into the global variable
  arweave = ArweaveService(
      Arweave(gatewayUrl: Uri.parse(config.defaultArweaveGatewayUrl!)));
  /// load local storage
  await KeyValueStorage.init();
  /// or add a new global variable like `localStorage`
  /// and then `localStorage = KeyValueStorage.getInstance();`
}
```

And then the full code

```dart
late ConfigService configService;
late AppConfig config;
late ArweaveService arweave;

void main() async {
  WidgetsFlutterBinding.ensureInitialized();
  await loadCoreDependencies();
  refreshHTMLPageAtInterval(Duration(hours: 12));
  runApp(App());
}

loadCoreDependencies() {
  /// load config service into the global variable
  configService = ConfigService();
  /// load config into the global variable
  config = await configService.getConfig();
  /// load arweave into the global variable
  arweave = ArweaveService(
      Arweave(gatewayUrl: Uri.parse(config.defaultArweaveGatewayUrl!)));
  /// load local storage
  await KeyValueStorage.init();
  /// or add a new global variable like `localStorage`
  /// and then `localStorage = KeyValueStorage.getInstance();`
}
```

test/utils/key_value_store_test.dart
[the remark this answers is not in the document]
matibat: > Also the description should be updated once expanded

I just renamed the description.


> If the default is false

The previous test sets it to `true`. So the test begins with the value being `true`, then after `remove()` the value turns `null`

lib/blocs/feedback_survey/feedback_survey_cubit.dart
matibat: I didn't apply the global singleton approach. Let's see if what I did is OK, if not I'd need some help to understand the suggested approach

lib/components/create_manifest_form.dart
matibat:
```suggestion
          context.read<FeedbackSurveyCubit>().openRemindMe();
```

lib/components/drive_share_dialog.dart
matibat:
```suggestion
                 context.read<FeedbackSurveyCubit>().openRemindMe();
```

lib/components/drive_share_dialog.dart
```diff
@@ -98,9 +98,9 @@ class _DriveShareDialogState extends State<DriveShareDialog> {
           actions: [
             if (state is DriveShareLoadSuccess)
               ElevatedButton(
-                onPressed: () {
+                onPressed: () async {
```
matibat, line 101:
```suggestion
                onPressed: () {
```

lib/components/feedback_survey.dart
matibat, line 105:
```suggestion
                                  onPressed: () {
                                    context
```

lib/components/feedback_survey.dart
matibat:
```suggestion
                                  context
```

lib/components/file_share_dialog.dart
```diff
@@ -99,9 +99,9 @@ class _FileShareDialogState extends State<FileShareDialog> {
           actions: [
             if (state is FileShareLoadSuccess)
               ElevatedButton(
-                onPressed: () {
+                onPressed: () async {
```
matibat, line 102:
```suggestion
                onPressed: () {
```

lib/components/file_share_dialog.dart
matibat:
```suggestion
                  context.read<FeedbackSurveyCubit>().openRemindMe();
```

test/test_utils/fakes.dart
matibat: It was @fedellen 's catch 😎

matibat (commented):
* self-review checkpoint *

lib/utils/key_value_store.dart
matibat: Really useful comment! This was applied. Thanks!

thiagocarvalhodev (approved):
Great job!

karlprieb (approved):

PR #545, opened 2022-06-14
Release v1.21.0

thiagocarvalhodev:
#### Release notes:
- Logging out and logging back in will no longer prompt the user to attach the previous selected drive
- Files that are too large to upload will now be filtered out before checking for name conflict resolution
- Drive explorer view now supports sorting by numerical values in the order in which they appear on the number line


agsuy (approved):
1.21.0 fixing conflicts check

PR #546, opened 2022-06-14
Fix not checking for conflicting files when we don't have any files above limit

thiagocarvalhodev:


lib/blocs/upload/upload_cubit.dart
```diff
@@ -131,7 +131,11 @@ class UploadCubit extends Cubit<UploadState> {
         tooLargeFileNames: tooLargeFiles,
         isPrivate: _targetDrive.isPrivate,
       ));
+      return;
     }
+    
+    // If we don't have any file above limit, we can check conflicts
+    checkConflicts();
```
fedellen, line 138: 👀

fedellen: Was this skipping conflict resolution on uploads with no over-sized files?

agsuy: Yes :sob:

fedellen (approved):

karlprieb (approved):

PR #550, opened 2022-06-16
PE-1580: Survey Feedback Modal

matibat:
Thew integration branch of the sub-tasks.


lib/components/create_manifest_form.dart
```diff
@@ -37,7 +38,7 @@ class CreateManifestForm extends StatelessWidget {
   @override
   Widget build(BuildContext context) =>
       BlocConsumer<CreateManifestCubit, CreateManifestState>(
-          listener: (context, state) {
+          listener: (context, state) async {
```
fedellen, line 41: Is this still async?

karlprieb: good catch @fedellen I believe @matibat just forgot to clean this up

lib/components/feedback_survey.dart
```diff
@@ -0,0 +1,150 @@
...
+    );
+
+class FeedbackSurveyModal extends StatelessWidget {
+  const FeedbackSurveyModal({Key? key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) =>
+      BlocBuilder<FeedbackSurveyCubit, FeedbackSurveyState>(
+        builder: (context, state) => Stack(
+          children: [
```
fedellen, line 26: I wonder if we could break some if this up here into individual widgets with some descriptive names.

lib/l10n/app_en.arb
```diff
@@ -1291,5 +1291,33 @@
...
+  "@leaveFeedback": {
+    "description": "The action of leaving feedback"
+  },
+  "noThanks": "No thanks! Don't remind me again",
+  "@noThanks": {
+    "description": "Don't leave feedback, and don't ask again for it"
+  },
+  "weWontRemindYou": "Ok cool! We won't remind you again.",
+  "@weWontRemindYou": {
+    "description": "We know the user don't want to leave feedback ATM"
```
fedellen, line 1313:
```suggestion
    "description": "We let the user know that we won't remind them to leave feedback again"
```

lib/components/feedback_survey.dart
matibat: I actually would love to split all of the widgets into individual reusable components. I think we don't yet have a place in the repo to place them...

lib/components/create_manifest_form.dart
matibat: Nice catch bbe927425a72de7430a16306c32165426d294638

lib/components/feedback_survey.dart
karlprieb: @thiagocarvalhodev correct me if I'm wrong, but I feel that's the way flutter developers are used to do things.
I really dislike, but all our code is like that right now. We should invest some time creating more individual widgets when we start to work on the redesign.

karlprieb (dismissed):

karlprieb (dismissed):

lib/components/feedback_survey.dart
thiagocarvalhodev: @karlprieb , is not. Most of the projects that I worked we used our component library to build the UI, based on our design system. I mean, we had our own buttons, texts, inputs, etc...

I was talking with Lucas about our own design system, and it is one of our goals on the UI side, but we won't have it on time to implement the redesign. So, I strongly recommend building our component library too, but starting smaller and simple. We can start creating Buttons, Texts, Inputs, Modals, etc... every new component from the redesign should be on that library, while we don't have our design system well defined on the UI side. After that, we can build a true design system, integrating UI/UX with devs.

We also must be very critical of the new components' implementation.

karlprieb (dismissed):

jdaev (approved):

PR #551, opened 2022-06-16
PE-1603: Spanish translations for the Survey Feedback Modal

matibat:


lib/l10n/app_es.arb
```diff
@@ -1291,5 +1291,33 @@
...
+  },
+  "feedbackContent": "¿Te interesaría dejar comentarios sobre tu experiencia en la aplicación?",
+  "@feedbackContent": {
+    "description": "Asks for feedback"
+  },
+  "leaveFeedback": "Dejar comentarios",
+  "@leaveFeedback": {
+    "description": "The action of leaving feedback"
+  },
+  "noThanks": "No, ¡gracias! No me lo recuerdes otra vez",
```
matibat, line 1307: This is syntacticly correct. The other valid option could be:
```suggestion
  "noThanks": "¡No, gracias! No me lo recuerdes otra vez",
```
That's only aesthetics :)

karlprieb: Can we have this fixed?

karlprieb (approved):

PR #556, opened 2022-06-22
Add logs to sync Cubit and verify if the current SyncCubit instance i…

thiagocarvalhodev:


lib/blocs/profile/profile_cubit.dart
```diff
@@ -179,7 +179,9 @@ class ProfileCubit extends Cubit<ProfileState> {
   Future<void> deleteTables() async {
     // Delete all table data.
     await _db.transaction(() async {
+      print('Is table empty: ${_db.allTables.isEmpty}');
       for (final table in _db.allTables) {
+        print('deleting table: ${table.actualTableName}');
```
matibat, line 184: Do we want these logs here? If not then...
```suggestion
      for (final table in _db.allTables) {
```

lib/blocs/sync/sync_cubit.dart
```diff
@@ -114,30 +116,50 @@ class SyncCubit extends Cubit<SyncState> {
         _db = db,
         super(SyncIdle()) {
     // Sync the user's drives on start and periodically.
+    syncFormatedPrint('Building Sync Cubit...');
+
     createSyncStream();
```
matibat, line 121: Hmmm after this point in the constructor we do have `_syncSub` initialized. Consider making it non-nullable and `late` if possible.

lib/blocs/sync/sync_cubit.dart
```diff
@@ -114,30 +116,50 @@ class SyncCubit extends Cubit<SyncState> {
...
+  }
+
+  void _restartSync() {
+    syncFormatedPrint(
+        'Trying to create a sync subscription when window get focused again. This Cubit is active? ${!isClosed}');
+
+    if ((_lastSync != null &&
+            DateTime.now().difference(_lastSync!).inMinutes <
+                kSyncTimerDuration) ||
+        isClosed) {
```
matibat, line 155: Is there a way to fix this indentation here? If not that's ok 😅

lib/blocs/sync/sync_cubit.dart
```diff
@@ -430,21 +469,12 @@ class SyncCubit extends Cubit<SyncState> {
       if (firstBlockHeight == null) {
         firstBlockHeight = t.first.node.block!.height;
         totalBlockHeightDifference = currentBlockheight - firstBlockHeight;
```
matibat, line 471: I wonder if there's a way of making `totalBlockHeightDifference` non-nullable.

Maybe by wrapping the conditional into an anonymous self-called function that returns the value directly to the variable declaration. Something like...
```dart
final totalBlockHeightDifference = () {
    if (firstBlock == null) {
         firstBlockHeight = t.first.node.block!.height;
         return currentBlockheight - firstBlockHeight;
    }
    return 0; // or something
}()
```

lib/blocs/sync/sync_cubit.dart
```diff
@@ -114,30 +116,50 @@ class SyncCubit extends Cubit<SyncState> {
         _db = db,
         super(SyncIdle()) {
     // Sync the user's drives on start and periodically.
+    syncFormatedPrint('Building Sync Cubit...');
+
     createSyncStream();
     restartSyncOnFocus();
     // Sync ArConnect
     createArConnectSyncStream();
```
matibat, line 124: After this point we also have `_arconnectSyncSub` defined, let's try to make it non-nullable and late if possible as well.

lib/utils/html/implementations/html_web.dart
```diff
@@ -1,17 +1,22 @@
...
+late StreamSubscription _onVisibilityChangeStream;
+
 void whenTabIsUnhidden(Function onShow) {
-  document.addEventListener('visibilitychange', (event) {
-    if (document.visibilityState != 'hidden') {
-      onShow();
-    }
+  _onVisibilityChangeStream = document.onVisibilityChange.listen((event) {
+    print('listening visibility change ${event.toString()}');
+    if (!isTabHidden()) onShow();
```
matibat, line 13: Prefer conditionals with body over single-line conditionals 🤓

matibat (commented):
Left some comments and questions here

matibat:
Would you mind double-checking if we do want the newly introduced debug messages? That's not a blocker anyways ;)

lib/blocs/sync/sync_cubit.dart
```diff
@@ -114,30 +116,50 @@ class SyncCubit extends Cubit<SyncState> {
...
+    if ((_lastSync != null &&
+            DateTime.now().difference(_lastSync!).inMinutes <
+                kSyncTimerDuration) ||
+        isClosed) {
+      syncFormatedPrint(
+          'Not possible restart sync when window get focused due to: is current active? ${!isClosed} or the last sync was ${DateTime.now().difference(_lastSync!).inMinutes} minutes ago. It should be $kSyncTimerDuration');
+      return;
+    }
+
+    Future.delayed(const Duration(seconds: 2)).then((value) {
```
karlprieb, line 161: Can we add a comment explaining that we wait 2 seconds to display the sync modal so it doesn't abruptly open the modal when the user is back to ArDrive browser tab?

thiagocarvalhodev: Nice!

lib/blocs/sync/sync_cubit.dart
thiagocarvalhodev: As this method is called more than one time we try to cancel the previous subscription with
``` dart
await _syncSub?.cancel();
```
If we don't make it nullable, we'll cancel something null, and we'll got a exception.

lib/blocs/sync/sync_cubit.dart
thiagocarvalhodev: As this method is called more than one time we try to cancel the previous subscription with
``` dart
_arconnectSyncSub?.cancel();
```
If we don't make it nullable, we'll cancel something null, and we'll got a exception.

lib/blocs/sync/sync_cubit.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/556/commits/0cec5904d60d5eb9c6d41e5ddbdc5cb1c4165505

lib/utils/html/implementations/html_web.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/556/commits/e1dd574f9a9c3d0f4c319f2fc5ebade3fbf827f3

lib/blocs/sync/sync_cubit.dart
thiagocarvalhodev: We can just make `totalBlockHeightDifference` as `late` 🤓

thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/556/commits/65313195a01dbd0e1d8385ad0233f8d628dad8c0

karlprieb (approved):

matibat (approved):
Nice improvements. Thanks! 😎

PR #559, opened 2022-06-23
PE-1757: Put the flame icon

matibat:
Changes:

- Adds a placeholder for the Flame 🔥 icon on its place.
- When the icon is clicked it opens the Inferno Rules.
- There's a tooltip wording: "The Inferno Rewards Program is in full swing! Click here to check out the rules for participation!"

TODO:

- Replace the placeholder with the actual design (surely addressed in another sub-task)

Screen Shot 2022-06-23 at 15 57 40

### DIFF
--- a/lib/components/app_drawer/app_drawer.dart
+++ b/lib/components/app_drawer/app_drawer.dart
@@ -32,11 +32,11 @@ class AppDrawer extends StatelessWidget {
                   Expanded(
                     child: Column(
                       children: [
-                        SizedBox(
+                        const SizedBox(
                           height: 32,
                         ),
                         _buildLogo(),
-                        SizedBox(
+                        const SizedBox(
                           height: 32,
                         ),
                         BlocBuilder<ProfileCubit, ProfileState>(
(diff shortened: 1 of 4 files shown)


lib/components/app_drawer/app_drawer.dart
```diff
@@ -116,45 +119,74 @@ class AppDrawer extends StatelessWidget {
...
-                                      .appVersion(snapshot.data!.version),
-                                  style: Theme.of(context)
-                                      .textTheme
-                                      .caption!
-                                      .copyWith(color: Colors.grey),
+                                child: FloatingActionButton(
+                                  elevation: 0,
+                                  tooltip: appLocalizationsOf(context).help,
+                                  onPressed: () =>
+                                      launch('https://ardrive.zendesk.com/'),
```
matibat, line 136: It's not in the scope of the ticket, but we could move this string to the resources file.

matibat: Here: 55c8707dd5e35d7046c7217b71a5bf9257658905

karlprieb (approved):

PR #560, opened 2022-06-27
PE-1785: Treat .tgz and .gz files as GZip

matibat:


test/utils/mime_lookup.dart
```diff
@@ -4,12 +4,14 @@ import 'package:flutter_test/flutter_test.dart';
...
-      'path/to/my/file.tar.gz': 'application/x-tar',
+      '.my.file.tar.gz': 'application/gzip',
+      'abcdefghijklmnñopqrstuvwxyz.tar.gz': 'application/gzip',
+      'path/to/my/file.tar.gz': 'application/gzip',
+      'path/to/my/file.gz': 'application/gzip',
+      'path/to/my/file.tgz': 'application/gzip',
       'path/to/my/non tar/file.iso': 'application/x-iso9660-image',
       'ardrive.png': 'image/png',
-      '.tar.gz': null,
+      '.tar.gz': 'application/gzip',
```
matibat, line 14: Previously, this filename was considered to match the MIME `null` (application/octet-stream) because there was no  file name (extension less) because we were considering `tar.gz` as the extension after the dot. Now, being `gz` the extension, the name of the file is `.tar` and the extension is `gz`.

matibat:
The same changes were applied at https://github.com/dart-lang/mime/pull/66 as well

karlprieb (approved):

PR #561, opened 2022-06-28
fix: Properly handle 5xx errors on sync PE-1747 

thiagocarvalhodev:
### Overview
For getting the metadata data while syncing we are not retrying for 5xx status code responses, this leads to a false positive and we also don’t retry to get the metadata properly.

### Errors
In order to track in a better way the errors, the class `ArDriveError` will be responsible for being the base class for exceptions.
The class `NetworkError` will be the base class for all network errors, having its implementation on:
`ServerError`
`RateLimitError`
`NetworkTimeOut`
`UnknownNetworkError`
Those errors are well known and will help us to improve our error handling.

### HttpRetry
This class abstracts the retry logic for `dart:http` library requests. For handling the response, we'll use the `GatewayResponseHandler` class.
For `RateLimitError` we won't retry. It is because we need to wait until be free of the rate limit.


matibat (commented):
The code looks good to me!

---------

I would simply be more explicit on the tests' description:
- The group description talking about what component (e.g. class) are we testing
- The test description specifying:
  - **What inputs** are being tested,
  - over **which methods** (e.g. "handle()"),
  - and what's the expected behavior (this last point seems to be correctly covered)

One example for this would be:
`handle method should return RateLimitError when response has err code 429`

test/utils/network_error_handler_test.dart
```diff
@@ -0,0 +1,25 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:http/http.dart';
+
+void main() {
+  final sut = NetworkErrorHandler();
+
+  final rateLimit = Response('body', 429);
+  final unknownError = Response('body', 400);
+  final serverError502 = Response('body', 502);
+  group('description', () {
```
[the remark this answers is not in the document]
thiagocarvalhodev: Good catch!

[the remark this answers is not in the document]
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/34c2f5175deb104767c83443d237d6ef73fb6acf

lib/utils/network_error_handler.dart
```diff
@@ -0,0 +1,24 @@
...
+
+class NetworkErrorHandler {
+  NetworkError handle(Response response) {
+    final requestRoute = response.request?.url.path;
+    final statusCode = response.statusCode;
+    if (response.statusCode >= 500) {
+      return ServerError(
+          statusCode: statusCode,
+          requestRoute: requestRoute,
+          reasonPhrase: response.reasonPhrase ?? '');
```
arielmelendez, line 12: The reason phrase null coalescing is happening in 3 different places. You might as well factor it out to a variable up above.

test/utils/http_retry_test.dart
```diff
@@ -0,0 +1,48 @@
+import 'package:ardrive/utils/error.dart';
+import 'package:ardrive/utils/http_retry.dart';
+import 'package:ardrive/utils/network_error_handler.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:http/http.dart';
+
+void main() {
+  HttpRetry sut = HttpRetry(NetworkErrorHandler());
+
+  group('Testing HttpRetry', () {
```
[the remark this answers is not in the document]
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/a175eaafeba08f538eb71641fbf3647e6045a78d

test/utils/http_retry_test.dart
```diff
@@ -0,0 +1,48 @@
...
+    test('should throw UnknownNetworkError', () async {
+      expect(() async => await sut.call(() async => unknownError),
+          throwsA(const TypeMatcher<UnknownNetworkError>()));
+    }, timeout: timeoutForWaitRetries);
+
+    test('should throw ServerError for status code 502', () async {
+      expect(() async => await sut.call(() async => serverError502),
+          throwsA(const TypeMatcher<ServerError>()));
+    }, timeout: timeoutForWaitRetries);
+    test('should throw ServerError for status code 504', () async {
```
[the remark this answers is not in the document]
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/a175eaafeba08f538eb71641fbf3647e6045a78d

lib/utils/network_error_handler.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:ardrive/utils/error.dart';
+import 'package:http/http.dart';
+
+class NetworkErrorHandler {
+  NetworkError handle(Response response) {
```
arielmelendez, line 5: Why would we return any kind of error for a 200? Why not make NetworkError nullable?

lib/utils/network_error_handler.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:ardrive/utils/error.dart';
+import 'package:http/http.dart';
+
+class NetworkErrorHandler {
```
arielmelendez, line 4: This is less of a "handler" and more of a "mapper"

lib/utils/network_error_handler.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/917a78849009b68c69b2932a9a2e98fe711954e8

arielmelendez:
Why did we build all these new classes if the dart retry library supports conditional retries via `retryIf`? Would have been a much smaller PR...

lib/utils/network_error_handler.dart
thiagocarvalhodev: Interesting point. My point with this class was to map the correct exception for each status code. The client class should verify any errors before mapping the response with this class. What do you think?

arielmelendez: I think this code is all kind of unusual to see in application code and a bit heavy handed when useful, tested alternatives already exist off the shelf. See my comment here: https://github.com/ardriveapp/ardrive-web/pull/561#issuecomment-1169127862

thiagocarvalhodev:
> Why did we build all these new classes if the dart retry library supports conditional retries via retryIf? Would have been a much smaller PR... See: https://github.com/google/dart-neats/tree/master/retry#using-retry

My goal was to map different error cases, so this can be useful not only on `retry` but for every http request. With that, we can improve our logs: for instance printing the type of exception, the status code, and the route for that exception...

lib/utils/http_retry.dart
```diff
@@ -0,0 +1,36 @@
...
+
+import 'error.dart';
+import 'network_error_handler.dart';
+
+class HttpRetry {
+  HttpRetry(this.errorHandler);
+
+  final NetworkErrorHandler errorHandler;
+
+  Future<Response> call(Future<Response> Function() request) {
```
arielmelendez, line 12: This is not a great name for a function. It's not explaining what it's actual behaviors are, namely:
- retrying the provided closure if an exception occurs
- passing http responses to handler which itself is decided whether the http response should caused an exception to be thrown.

Consider a more useful name like: `processRequest`

arielmelendez: Also consider naming the input parameter `requestFactory` since it's a factory function.

lib/utils/http_retry.dart
```diff
@@ -0,0 +1,36 @@
+import 'package:http/http.dart';
+import 'package:retry/retry.dart';
+
+import 'error.dart';
+import 'network_error_handler.dart';
+
+class HttpRetry {
+  HttpRetry(this.errorHandler);
+
+  final NetworkErrorHandler errorHandler;
```
arielmelendez, line 10: Consider an interface called `ResponseHandler` and then create an implementation like `GatewayResponseHandler`.

arielmelendez: I think the `handle` function would just be a void function that may or may not throw based on the response object passed to it.

lib/utils/http_retry.dart
```diff
@@ -0,0 +1,36 @@
...
+
+      if (response.statusCode == 200) {
+        return response;
+      }
+
+      throw errorHandler.handle(response);
+    }, onRetry: (exception) {
+      if (exception is NetworkError) {
+        print(
+          'Retrying for ${exception.runtimeType} exception\n'
```
arielmelendez, line 24: I don't think we need BOTH an exception AND a status code - we just need one or the other in the log. In the case of exceptions, there may not even be a status code.

lib/utils/http_retry.dart
```diff
@@ -0,0 +1,36 @@
...
+class HttpRetry {
+  HttpRetry(this.errorHandler);
+
+  final NetworkErrorHandler errorHandler;
+
+  Future<Response> call(Future<Response> Function() request) {
+    return retry(() async {
+      final response = await request();
+
+      if (response.statusCode == 200) {
```
arielmelendez, line 16: This would go away if the handler were responsible for deciding whether or not to throw.

lib/utils/http_retry.dart
```diff
@@ -0,0 +1,36 @@
...
+          'Retrying for ${exception.runtimeType} exception\n'
+          'for route ${exception.requestRoute}\n'
+          'and status code ${exception.statusCode}',
+        );
+        return;
+      }
+
+      print('Retrying for unknown exception: ${exception.toString()}');
+    }, retryIf: (exception) {
+      return exception is! RateLimitError;
```
arielmelendez, line 33: This makes me sad. Large drives with fast internet connections simply will not work. Let's open a tech debt ticket for this. I think we need to make a PR to dart-neats for the retry library so that we can provide for dynamic delay duration calculation based on exception type (i.e.  rate limit exception).

arielmelendez (changes requested):
See my comments about the handler class and the tech debt ticket.

lib/utils/http_retry.dart
thiagocarvalhodev: We can have more than one status code for a specific exception, that is the cade of ServerError that can have 5xx status code. I guess for log porpuses it will help us to underdant which error we are facing and also the request route related to that. What do you think?

lib/utils/http_retry.dart
thiagocarvalhodev: Great suggestion!

thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/97e0dbdda5302d64c23bf14fc238b91fd69bf791

lib/utils/http_retry.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/876963f818ada6359d4caf9388e2f9c87e3295ad

lib/utils/error.dart
```diff
@@ -0,0 +1,76 @@
+// TODO(@thiagocarvalhodev): Add more error constants
```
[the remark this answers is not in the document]
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/fc885b45257082ca708ba932805179170463f260

lib/utils/error.dart
```diff
@@ -0,0 +1,75 @@
+import 'package:equatable/equatable.dart';
+
+// ignore: constant_identifier_names
+const int RATE_LIMIT_ERROR = 429;
+
+abstract class ArDriveError implements Exception {}
```
arielmelendez, line 6: Remove this class. It has no clear or documented value right now.

lib/utils/response_handler.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:ardrive/utils/error.dart';
+import 'package:http/http.dart';
+
+abstract class ResponseHandler {
+  Response? handle(Response response);
```
arielmelendez, line 5: Make this a void function

lib/utils/response_handler.dart
```diff
@@ -0,0 +1,38 @@
...
+}
+
+class GatewayResponseHandler implements ResponseHandler {
```
arielmelendez, line 8: Move this class to be a peer to arweave_service.dart in the services/arweave folder so that the handling and errors are in the logical domain of Arweave code.

lib/utils/error.dart
```diff
@@ -0,0 +1,75 @@
...
+
+/// Handles every `http` exception
+abstract class NetworkError extends ArDriveError with EquatableMixin {
```
arielmelendez, line 9: Rename NetworkError to GatewayError.

lib/utils/response_handler.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/8bd1bdb4ea435b1dff02a68c16af6ba9e2ec1076

lib/utils/response_handler.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/8bd1bdb4ea435b1dff02a68c16af6ba9e2ec1076

lib/utils/error.dart
```diff
@@ -0,0 +1,75 @@
...
+  NetworkError(
+      {this.requestRoute,
```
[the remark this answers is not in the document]
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/8bd1bdb4ea435b1dff02a68c16af6ba9e2ec1076

lib/utils/response_handler.dart
```diff
@@ -0,0 +1,38 @@
...
+  Response handle(Response response) {
+    /// Success
+    if (response.statusCode == 200) {
+      return response;
+    }
+
+    throw _mapError(response);
+  }
+
+  NetworkError _mapError(Response response) {
```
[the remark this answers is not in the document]
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/commit/8bd1bdb4ea435b1dff02a68c16af6ba9e2ec1076

lib/utils/error.dart
```diff
@@ -0,0 +1,75 @@
...
+      : super(
+            reasonPhrase: reasonPhrase,
+            statusCode: statusCode,
+            requestRoute: requestRoute);
+
+  @override
+  List<Object?> get props => [statusCode, reasonPhrase, requestRoute];
+}
+
+class NetworkTimeOut extends NetworkError {
```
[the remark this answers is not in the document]
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/commit/8bd1bdb4ea435b1dff02a68c16af6ba9e2ec1076

lib/utils/error.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/e1145156d6b54984a915cdc78b25c56a4804abda

lib/services/arweave/error/gateway_error.dart
```diff
@@ -0,0 +1,76 @@
...
+  final String reasonPhrase;
+
+  factory GatewayError.fromResponse(Response response) {
+    final requestUrl = response.request?.url.path;
+    final statusCode = response.statusCode;
+    final reasonPhrase = response.reasonPhrase ?? '';
+    if (response.statusCode >= 500) {
+      return ServerError(
+          statusCode: statusCode,
+          requestRoute: requestUrl,
```
arielmelendez, line 23: The word "route" is still leaking here 😬

lib/services/arweave/error/gateway_response_handler.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:ardrive/services/arweave/error/gateway_error.dart';
+import 'package:ardrive/utils/response_handler.dart';
+import 'package:http/http.dart';
+
+/// Throws a `GatewayError` exception for `statusCode` different from 200
```
arielmelendez, line 5: nit: not in the 2xx range

test/services/arweave/error/response_handler_test.dart
```diff
@@ -0,0 +1,52 @@
...
+  final success = Response('body', 200);
+  final success250 = Response('body', 250);
+  final success201 = Response('body', 201);
+
+  final rateLimit = Response('body', 429);
+  final unknownError = Response('body', 400);
+  final unknownError300 = Response('body', 300);
+  final serverError502 = Response('body', 502);
+  group('Testing GatewayResponseHandler class', () {
+    test('should return the response for success', () {
```
arielmelendez, line 17: should return null

test/services/arweave/error/response_handler_test.dart
```diff
@@ -0,0 +1,52 @@
...
+      sut.handle(success);
+      expect(null, null);
+    });
+    test('should return the response for success', () {
+      /// Should not throw exception. We can do `expect(null, null);` safely here as this function
+      /// doesnt return nothing in case of success.
+      sut.handle(success250);
+      expect(null, null);
+    });
+    test('should return the response for success', () {
```
arielmelendez, line 29: also null

test/services/arweave/error/response_handler_test.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:ardrive/services/arweave/error/gateway_error.dart';
+import 'package:ardrive/services/arweave/error/gateway_response_handler.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:http/http.dart';
+
+void main() {
+  final sut = GatewayResponseHandler();
+  final success = Response('body', 200);
+  final success250 = Response('body', 250);
```
arielmelendez, line 9: this should probably be 299

test/utils/http_retry_test.dart
```diff
@@ -0,0 +1,35 @@
...
+  HttpRetry sut = HttpRetry(mockResponseHandler);
+
+  group('Testing HttpRetry class', () {
+    final tResponse = Response('body', 200);
+
+    setUp(() {
+      registerFallbackValue(tResponse);
+    });
+
+    test('Should return the response when dont have any errors', () async {
```
arielmelendez, line 22: null

arielmelendez (changes requested):
left some minor comments re: test names and inputs

test/utils/http_retry_test.dart
thiagocarvalhodev: Here we returns the response.

test/services/arweave/error/response_handler_test.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/6042fdf5b0d5aeb47182e10f29e2dab3d6e285f2

test/services/arweave/error/response_handler_test.dart
```diff
@@ -0,0 +1,52 @@
...
+  final unknownError300 = Response('body', 300);
+  final serverError502 = Response('body', 502);
+  group('Testing GatewayResponseHandler class', () {
+    test('should return the response for success', () {
+      /// Should not throw exception. We can do `expect(null, null);` safely here as this function
+      /// doesnt return nothing in case of success.
+      sut.handle(success);
+      expect(null, null);
+    });
+    test('should return the response for success', () {
```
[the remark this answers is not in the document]
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/6042fdf5b0d5aeb47182e10f29e2dab3d6e285f2

test/services/arweave/error/response_handler_test.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/6042fdf5b0d5aeb47182e10f29e2dab3d6e285f2

lib/services/arweave/error/gateway_response_handler.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/6042fdf5b0d5aeb47182e10f29e2dab3d6e285f2

test/services/arweave/error/response_handler_test.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/a8e855cc084c4bb80384ee2d91e6ab7675ce0f35

lib/services/arweave/error/gateway_error.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/561/commits/b17012ca00fe91ca23ac7b43584b2f12c5e15bc1

arielmelendez (approved):

test/services/arweave/error/response_handler_test.dart
```diff
@@ -0,0 +1,52 @@
...
+      sut.handle(success299);
+      expect(null, null);
+    });
+    test('should return null for success', () {
+      /// Should not throw exception. We can do `expect(null, null);` safely here as this function
+      /// doesnt return nothing in case of success.
+      sut.handle(success201);
+      expect(null, null);
+    });
+    test('should throws a ServerError', () {
```
agsuy, line 35: Nit(grammar):
```suggestion
    test('should throw a ServerError', () {
```

thiagocarvalhodev:
Improvements:

- Added `UnExpectedRedirection` error
- Test if the `HttpRetry` is retrying how many times we choose
- Pass the callbacks functions on a object `HttpRetryOptions` in order to test
- Verify only possible 2xx status code

karlprieb (approved):

lib/utils/http_retry.dart
```diff
@@ -0,0 +1,45 @@
...
+          }
+
+          return response;
+        },
+        maxAttempts: httpRetryOptions.maxAttempts,
+        randomizationFactor: 0,
+        onRetry: (e) {
+          ++retryAttempts;
+          httpRetryOptions.onRetry?.call(e);
+        },
```
matibat, line 42: The exponential backoff for error 429 can be done here if we declare the method as async (a future).

PR #562, opened 2022-06-28
feat(survey feedback modal): set the pref on dontRemindMe() PE-1773

matibat:
The preference was previously set after the second view of the modal: you first clicked "don't remind me again" and nothing happened, then you clicked "Got it" and the flag was set after.
Now the flag is set when clicking the first button.

### DIFF
--- a/lib/blocs/feedback_survey/feedback_survey_cubit.dart
+++ b/lib/blocs/feedback_survey/feedback_survey_cubit.dart
@@ -43,12 +43,12 @@ class FeedbackSurveyCubit extends Cubit<FeedbackSurveyState> {
     emit(FeedbackSurveyDontRemindMe(isOpen: false));
   }
 
-  void dontRemindMe() {
+  Future<void> dontRemindMe() async {
+    await (await _store).putBool(dontRemindMeAgainKey, true);
     emit(FeedbackSurveyDontRemindMe(isOpen: true));
   }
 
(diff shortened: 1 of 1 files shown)


lib/blocs/feedback_survey/feedback_survey_cubit.dart
```diff
@@ -43,12 +43,12 @@ class FeedbackSurveyCubit extends Cubit<FeedbackSurveyState> {
     emit(FeedbackSurveyDontRemindMe(isOpen: false));
   }
 
-  void dontRemindMe() {
+  Future<void> dontRemindMe() async {
+    await (await _store).putBool(dontRemindMeAgainKey, true);
```
thiagocarvalhodev, line 47: You can use the cascade notation:

```dart
await _store..putBool(dontRemindMeAgainKey, true);
```

thiagocarvalhodev (approved):
LGTM! Left just a non-block comment.

lib/blocs/feedback_survey/feedback_survey_cubit.dart
matibat: The linter gives me back:
> Avoid single cascade in expression statements.dart[avoid_single_cascade_in_expression_statements](https://dart-lang.github.io/linter/lints/avoid_single_cascade_in_expression_statements.html)

PR #564, opened 2022-06-28
feat(leave feedback button): align to the left PE-1778

matibat:


matibat:
## Before

Screen Shot 2022-06-28 at 18 36 01

## After

Screen Shot 2022-06-28 at 18 36 01

PR #565, opened 2022-06-28
PE-1772: Open Feedback Modal after Share Dialog is dismissed by clicking the screen behind

matibat:
Changes:
- Makes AppDialog to take an extra onWillPopCallback function.
- Pass the callback at the Share Dialog to open the Feedback Modal.

### DIFF
--- a/lib/components/app_dialog.dart
+++ b/lib/components/app_dialog.dart
@@ -25,9 +25,7 @@ class AppDialog extends StatelessWidget {
   @override
   Widget build(BuildContext context) => WillPopScope(
         onWillPop: () async {
-          if (onWillPopCallback is Function) {
-            await onWillPopCallback!();
-          }
+          await onWillPopCallback?.call();
           return dismissable;
         },
         child: AlertDialog(
--- a/lib/components/app_dialog.dart
+++ b/lib/components/app_dialog.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:ardrive/theme/theme.dart';
 import 'package:flutter/material.dart';
 
@@ -6,7 +8,7 @@ class AppDialog extends StatelessWidget {
   final EdgeInsetsGeometry contentPadding;
   final Widget content;
   final List<Widget> actions;
-  final Future<void> Function()? onWillPopCallback;
+  final FutureOr<void> Function()? onWillPopCallback;
 
   final bool dismissable;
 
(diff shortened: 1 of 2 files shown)


lib/components/app_dialog.dart
```diff
@@ -6,20 +6,27 @@ class AppDialog extends StatelessWidget {
   final EdgeInsetsGeometry contentPadding;
   final Widget content;
   final List<Widget> actions;
+  final Future<void> Function()? onWillPopCallback;
```
karlprieb, line 9: why onWillPopCallback should be a `Future`? If I don't want to pass a `Future`, but just a function, it will work?

thiagocarvalhodev: We can use `FutureOr` class for this purpose.

thiagocarvalhodev: Futures should be `async` function. Using the `FutureOr` you can pass both `Future` and non Future functions.

matibat: There. Thanks!
fa629a09ae3fb7f31a20ddef7cc14ca6bd406712

lib/components/app_dialog.dart
```diff
@@ -6,20 +8,28 @@ class AppDialog extends StatelessWidget {
...
-  });
+    Key? key,
+  }) : super(key: key);
 
   @override
   Widget build(BuildContext context) => WillPopScope(
-        onWillPop: () async => dismissable,
+        onWillPop: () async {
+          if (onWillPopCallback is Function) {
+            await onWillPopCallback!();
```
thiagocarvalhodev, line 29: Why verify the type: `(onWillPopCallback is Function)`? As you are defining the type up above, you don't need to do that.

If you want to prevent of crash in the case of not providing this function you can use that:

```dart
onWillPop: () async {
    await onWillPopCallback?.call();
    return dismissable;
 }
 ```
Every `Function` has the method `call`. For instance call `foo()` or `foo.call()` has the same result and you can check if it is not null

thiagocarvalhodev (approved):

lib/components/app_dialog.dart
matibat: Nice suggestion!

matibat: 5d81ea93495af223afc195ac750bd5ccf6bffc15

PR #566, opened 2022-06-28
chore(update version) change version to 1.22.1

thiagocarvalhodev:


agsuy (approved):
Change version as its only bugfixes

PR #571, opened 2022-06-30
PE-1580: Various fixes

matibat:
Changes:
- Use different style rules to align the button to the left
  - Align the button when not logged in as well
- Make the Leave Feedback button to be underlined and bold. Also add a blank space above.
- Show modal after Drive Share Dialog is dismissed (same behavior than File Share Dialog)


thiagocarvalhodev (approved):
LGTM!

fedellen (approved):

PR #572, opened 2022-07-01
PE-1751: Link to Inferno from Web App (Fire Icon)

matibat:
Changes:

- Adds the fire icon near the Help button
  - It has a tooltip wording: `The Inferno Rewards Program is in full swing! Click here to check out the rules for participation!`
  - When clicked it opens a new tab to the [Inferno rules](https://ardrive.io/inferno/)
- Aligns both the Help and Flame buttons with the Drives section above.


matibat:
## Fire icon

Screen Shot 2022-07-01 at 14 27 02

PR #573, opened 2022-07-01
feat: direct inferno icon to chinese rules PE-1808

thiagocarvalhodev:
The rules for Inferno are written in English and Chinese, yet the in-app fire icon only directs users to the English rules page. We need to ensure that users of the Chinese version of the app are redirected to the Chinese rules page.


karlprieb:
Se could have a dictionary pointing language code to specific URL and default to English

jdaev (approved):
LGTM

PR #583, opened 2022-07-08
PE-1848/Remove Gitopia Action

javdhu:
Removes gitopia.yaml


agsuy (approved):
Gitopia action was not in use.

PR #584, opened 2022-07-11
feat: update mandarin string literals

fewwwww:
- Added new translated strings
- Update better translation for previous ones

PR #585, opened 2022-07-11
PE-1856: Translate Feedback Modal and Inferno Link to Chinese

matibat:


matibat:
Following up of #584

matibat:
I just verified that all of the missing keys are now present in the file 👍

karlprieb (approved):

PR #590, opened 2022-07-14
refactor: update validation PE-1873

javdhu:
Updates validation message and changes flow to remain in the same page if key is incorrect.


fedellen (approved):

matibat (commented):
Looks good to me. Thanks!

I will approve after we have some consensus with the team regarding the requirements 😎

matibat (approved):
LGTM

PR #598, opened 2022-07-20
PE-1498 Upload Fail Dialog

javdhu:
Improves error handling for uploads.

### DIFF
--- a/lib/l10n/app_es.arb
+++ b/lib/l10n/app_es.arb
@@ -140,7 +140,7 @@
   "@passwordMismatch": {
     "description": "Error message for when the confirmation password don't match the actual pass"
   },
-  "aggreeToTerms_body": "Acepto los términos de servicio y la política de privacidad de ArDrive",
+  "aggreeToTerms_body": "Acepto los términos de servicio y la política de privacidad de ArDrive.",
   "@aggreeToTerms_body": {
     "description": "Checkbox for agreeing the terms of of service and privacy policy"
   },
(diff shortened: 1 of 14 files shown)


lib/l10n/app_en.arb
```diff
@@ -1355,5 +1355,13 @@
...
   "@infernoIsInFullSwing": {
     "description": "Text inviting the users to go read the Inferno rules"
+  },
+  "uploadFailed": "Upload Failed",
+  "@uploadFailed": {
+    "description": "Upload Failed dialog title"
+  },
+  "yourUploadFailed": "Upload Failed",
+  "@yourUploadFailed": {
+    "description": "Your upload failed. You have not been charged for the failed attempt."
```
matibat, line 1365: Is the description and the actual wording toggled here?

matibat (commented):
I think there's one single mistake in the ARB file

lib/blocs/upload/upload_handles/bundle_upload_handle.dart
```diff
@@ -1,117 +1,117 @@
...
+      bundle.blob,
+      wallet,
+    );
+
+    await pstService.addCommunityTipToTx(bundleTx);
+
+    await bundleTx.sign(wallet);
+
+    // Write entities to database
+    for (var folder in folderDataItemUploadHandles) {
```
karlprieb, line 69: What's the difference between for and forEach in this case?

lib/l10n/app_en.arb
karlprieb: I believe these are new keys, right?

karlprieb (approved):

matibat (dismissed):
We are gonna move forward with this as it is just "to have something".
Then we are fully address the pending case (partial failed upload) in a following up Story which:
- Handles the partial upload
- Applies the wording specified at the spec

_**Note** we are gonna go with English only for now because the actual wording is gonna change_

thiagocarvalhodev (dismissed):

thiagocarvalhodev (approved):

PR #609, opened 2022-07-28
refactor: add upload fail mandarin translations PE-1933

javdhu:
Adds mandarin translations


matibat (commented):
We are missing tryAgain non-capitalized. Are we missing the key for English as well?

matibat (approved):
LGTM then 😎

thiagocarvalhodev (approved):

PR #621, opened 2022-08-05
PE-1918: PST composite

matibat:
**Changes**

- Ports the safe types: `ArweaveAddress`, `Winston`, and `TransactionID` from CORE into Dart
- Implements a Builder for the contract state
- Implements the `ContractReader` interface
  - Implements `RedstoneContractReader`
  - Implements `VertoContractReader`
  - Implements `SmartweaveContractReader`
    - which refactors the implementation of web-pst to simply read the contract (removed the TS code that is already ported to Dart)
- Implements the `ContractOracle` ~interface~ class
  - Implements `ArDriveContractOracle` (the composite class; implements `ContractOracle`)
- Implements the `CommunityOracle` ~interface~ class
  - which makes use of a `ContractOracle` (e.i. the composite class) to make high-level computations
- Injects the `CommunityOracle` to the `PSTService`

**TODO**

- ~Make the `weightedRandom` method testable~


lib/pst/pst_contract_data.dart
```diff
@@ -0,0 +1,105 @@
+import 'package:ardrive/types/arweave_address.dart';
+import 'package:equatable/equatable.dart';
+
+typedef CommunityTipPercentage = double;
+
+/// Shape of the ArDrive Community Smart Contract
```
matibat, line 6: ArDrive Community Smart Contract STATE

lib/pst/pst_contract_data.dart
```diff
@@ -0,0 +1,105 @@
...
+  final int lockMinLength;
+  final int lockMaxLength;
+  final String communityAppUrl;
+  final List<String> communityDiscussionLinks;
+  final String communityDescription;
+  final ArweaveAddressType communityLogo;
+  final int fee;
+}
+
+class VaultItem extends Equatable {
```
matibat, line 92: TODO: it doesn't need to extend Equatable, that's only a facility to run expectations in the tests.

test/pst/pst_contract_data_builder_test.dart
```diff
@@ -0,0 +1,78 @@
...
+      expect(firstVote.start, 516926);
+      expect(firstVote.totalWeight, 1182600000000);
+      expect(
+        firstVote.recipient,
+        ArweaveAddress('rcO1Hi4_chcHVFBpSGOAVHVBBcljdypxlCBCbPfnu-c'),
+      );
+      expect(firstVote.qty, 50000);
+      expect(firstVote.lockLength, 262800);
+    });
+  });
```
matibat, line 77: TODO: test the un-happy paths

matibat (commented):
\* self review checkpoint *

test/pst/pst_contract_data_builder_test.dart
```diff
@@ -0,0 +1,78 @@
...
+  group('CommunityContractDataBuilder', () {
+    test('parses a CommunityContractData when fed with healthy data', () {
+      final builder = CommunityContractDataBuilder(rawHealthyContractData);
+      final output = builder.parse();
+
+      expect(output.name, 'ArDrive');
+      expect(output.ticker, 'ARDRIVE');
+
+      expect(output.balances, isA<Map<ArweaveAddressType, int>>());
+      expect(output.balances.length, 1288);
```
matibat, line 18: TODO: make expectations for the fist item in the Map

lib/pst/pst_contract_data.dart
matibat: 77b9484b308bc1e2b85e3b45d7bc5be2b041d27e

lib/pst/pst_contract_data.dart
matibat: 829c6f42616b796087ce01a762b7275ced10a5c0

test/pst/pst_contract_data_builder_test.dart
matibat: 2af4dd2ba3ac03bfb4f56d3231516a1b02aa8cbe

test/pst/pst_contract_data_builder_test.dart
matibat: A bunch of changes:
- 2af4dd2ba3ac03bfb4f56d3231516a1b02aa8cbe
- ccae7ebae0071ef91a72d4d3e7317b8bc6c80eae
- 5ac90537e2c81747616caaf1096ed0cc3ef4fbc2
- 79f3cceb591bd465ba0e0ffb44e3872c924d49b6

matibat:
Only missing:
- the implementations of ContractReader and ContractOracle, then the composite, and the CommunityOracle which is the high level class to be used in the codebase;
- integrate `web-pst` to the ContractOracle composite (only for web);
- start using the high-level class in the codebase

lib/pst/pst_contract_data_builder.dart
```diff
@@ -0,0 +1,363 @@
...
+                    'Expected the field .settings[number][1] ($key) to be a string, got $value',
+              );
+            }
+            break;
+
+          case 'quorum':
+          case 'support':
+          case 'voteLength':
+          case 'lockMaxLength':
+          case 'lockMinLength':
```
matibat, line 236: I'd say these should be integer; the other ones double :opionion:

lib/pst/pst_contract_data_builder.dart
```diff
@@ -0,0 +1,363 @@
...
+            final end = vaultItem['end'];
+            return VaultItem(balance: balance, start: start, end: end);
+          }).toList(growable: false);
+          return MapEntry(key, value);
+        },
+      ),
+    );
+
+    return CommunityContractData(
+        votes: votes, settings: settings, balances: balances, vault: vault);
```
matibat, line 88: Nit: add a comma to fix indentation

lib/pst/pst_contract_data.dart
```diff
@@ -0,0 +1,103 @@
+import 'package:ardrive/types/arweave_address.dart';
+
+typedef CommunityTipPercentage = double;
+
+/// Shape of the ArDrive Community Smart Contract data
```
matibat, line 5:
```suggestion
/// Shape of the ArDrive Community Smart Contract state
```

matibat (commented):
\* self review checkpoint *

lib/pst/pst_contract_data_builder.dart
matibat: 5d2f79dff58bab62d4c046bb38e7fe0605f11f4e

matibat:
All classes and unit tests are now implemented!

TODO: 
- Start using the implementation in the codebase
- Conditionally add web-pst when running in web

lib/pst/ardrive_contract_oracle.dart
```diff
@@ -0,0 +1,79 @@
+import 'package:ardrive/pst/contract_oracle.dart';
+import 'package:ardrive/pst/pst_contract_data.dart';
+import 'package:equatable/equatable.dart';
+
+const maxReadContractAttempts = 3;
```
thiagocarvalhodev, line 5: Do we use this constant in another place? If not, consider making it private.

lib/pst/ardrive_contract_oracle.dart
```diff
@@ -0,0 +1,79 @@
...
+
+class ArDriveContractOracle implements ContractOracle {
```
thiagocarvalhodev, line 7: Why implement a concrete class? We can just extend and override the superclass methods.

thiagocarvalhodev (commented):
Started the review. Tomorrow I'll finish!

lib/pst/ardrive_contract_oracle.dart
matibat: YES! That's the kind of questions I wanted here 🤩 

So, I was supposed to port the **same** code from CORE into DART. BUT, I made it different here: In dart we don't have interfaces, that's why I initially used abstract classes; now, the deal with `ContractOracle` is that in CORE it implemented `ContractReader` instead of using the Injection Pattern (thing we are using here). Given that `ContractOracle` does not work differently depending on the `ContractReader` (but instead it can take any `ContractReader` to work with), we'd only need a single implementation to be used with any `ContractReader`, and one single for the composite class: `ArDriveContractOracle`.

So now, the only abstract class (classes that does needs a custom implementation per each) is `ContractReader`. `ContractOracle` would take any `ContractReader`, `ArDriveContractOracle` would take an array of `ContractOracle`s to be composite, and `CommunityOracle` takes any `ContractOracle` (e.i. the composite `ArDriveContractOracle`).

I don't see the need of using more abstract classes

lib/types/winston.dart
```diff
@@ -0,0 +1,81 @@
+import 'package:equatable/equatable.dart';
+
+class Winston extends Equatable {
+  final int _amount;
```
matibat, line 4: Let's use BigInt instead; we already have the dependency

matibat: Feat: bd7b9b5574f5937f2cf2deea39b848ce5b61340e
Tests: 24b57dd203b77ea4e6dfeb77d4e0d6bca0b7f0d6

lib/pst/ardrive_contract_oracle.dart
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+
+  @override
+  Future<CommunityContractData> getCommunityContract() async {
+    CommunityContractData? data = await _getContractFromOracles();
+
+    if (data == null) {
+      throw const CouldNotReadContractState(
+        reason: 'Max retry attempts reached',
```
karlprieb, line 23: we don't know for sure if that's really the case here, right?
Why not get and error from `getCommunityContract` and throw here?

karlprieb (commented):
we should implement error handling on redstone and verto contract reader.

lib/pst/ardrive_contract_oracle.dart
matibat: Hmmm the ContractReaders don't retry

matibat: BTW the errors from ContractReader are bing caught here: https://github.com/ardriveapp/ardrive-web/pull/621/files#diff-d0a46439667cba712d6de87096e1e23c7416bb2453666bc1a62499758c317ffdR54

matibat:
> we should implement error handling on redstone and verto contract reader.

Indeed we do expect the `readContract` method to fail: it's supposed to trigger one single read attempt (with no retries or fallback), and the network is always open to errors

lib/pst/utils.dart
```diff
@@ -0,0 +1,17 @@
+import 'dart:math';
+
+import '../types/arweave_address.dart';
+
+ArweaveAddress? weightedRandom(Map<ArweaveAddress, double> dict) {
```
karlprieb, line 5: Why we have it on utils if we only use it on `CommunityOracle`?

test/pst/community_oracle_test.dart
```diff
@@ -0,0 +1,60 @@
...
+        expect(tipAmount, Winston(BigInt.from(10000000)));
+      });
+    });
+
+    group('selectTokenHolder method', () {
+      test('returns a valid address', () async {
+        final addr = await myCommunityOracle.selectTokenHolder();
+        expect(addr, ArweaveAddress('the expected value'));
+      });
+    }, skip: 'this method uses a random number - FIXME: make me testable');
```
karlprieb, line 58: Probably we can mock random to be not random :D

web-pst/rollup.config.js
```diff
@@ -22,7 +22,7 @@ export default {
     },
     {
       ...commonOutputOpts,
-      file: '../web/pst.min.js',
+      file: '../web/js/pst.min.js',
```
karlprieb, line 25: Thank you :D

lib/types/winston.dart
```diff
@@ -0,0 +1,82 @@
...
+  @override
+  List<Object?> get props => [_amount];
+
+  Winston(BigInt amount) : _amount = amount {
+    if (amount < BigInt.zero || !amount.isValidInt) {
+      throw InvalidWinstonValue();
+    }
+  }
+
+  Winston plus(Winston winston) {
```
karlprieb, line 15: if we're not using these methods I believe we don't need to keep them in the codebase. We can add them when needed.

test/pst/community_oracle_test.dart
matibat: Yup, I see two approaches:
- To mock the actual Random class globally
- To inject the value, taking as default the random

lib/pst/utils.dart
matibat: I wouldn't put it in the same source code because the method doesn't need to be imported from anywhere else, but if we instead make it private, we couldn't import the method from the tests. Same issues if we put the method in the context of the class

lib/types/winston.dart
matibat: Well... Indeed flutter lets you make custom implementations for the operators. We could implement the operator directly

matibat: Feature: f1708edb2557e81c0b0ac571b6a3ddfca911b78a
Tests: b9965fee6464a77a1c0208d87736e78acd83670d

test/pst/community_oracle_test.dart
matibat: I took the 2nd approach:
chore: 13abae48b2d653f1718658ef4445d37128b68160
tests: 817cfc4753182d0dda55c24b323c0cf7ee418dc8

lib/pst/ardrive_contract_oracle.dart
thiagocarvalhodev: Awesome!

lib/pst/ardrive_contract_oracle.dart
```diff
@@ -0,0 +1,95 @@
...
+
+  /// attempts multiple retries to read the given contract oracle
+  Future<CommunityContractData?> _getContractWithRetries(
+    ContractOracle contractOracle, {
+    int maxAttempts = _maxReadContractAttempts,
+  }) async {
+    CommunityContractData? data;
+    int readContractAttempts = 0;
+
+    while (data == null && readContractAttempts < maxAttempts) {
```
thiagocarvalhodev, line 52: We can use the `retry` package here.

matibat: Does retry work with http requests directly? If so that wouldn't work x.x

lib/pst/ardrive_contract_oracle.dart
```diff
@@ -0,0 +1,95 @@
...
+
+    return data;
+  }
+
+  /// iterates over all contract readers attempting to read the contract
+  Future<CommunityContractData?> _getContractFromOracles() async {
+    int contractOracleIndex = 0;
+    CommunityContractData? data;
+
+    while (data == null && _contractOracles.length > contractOracleIndex) {
```
thiagocarvalhodev, line 35: We can use the retry package here. Take a look at graphql_retry.dart

lib/pst/ardrive_contract_oracle.dart
thiagocarvalhodev, line 52: The same for using `retry` package.

lib/pst/community_oracle.dart
```diff
@@ -0,0 +1,93 @@
...
+
+    // Get the total number of token holders
+    int total = 0;
+    for (final addr in balances.keys) {
+      total += balances[addr]!;
+    }
+
+    // Check for how many tokens the user has staked/vaulted
+    for (final addr in vault.keys) {
+      if (vault[addr]!.isEmpty) continue;
```
thiagocarvalhodev, line 47: We are avoiding the `!` operator.

In this case, what you can do:

```dart
final vault = vault[addr];
if(vault == null || vault.isEmpty) continue;
```
and use the `vault` variable at the rest of the function.

lib/pst/contract_readers/redstone_contract_reader.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:convert';
+
+import 'package:ardrive/pst/contract_reader.dart';
+import 'package:ardrive/types/transaction_id.dart';
+import 'package:http/http.dart' as http;
+
+const cacheUrl = 'd2440r7x1v6779.cloudfront.net';
+const path = 'cache/state';
+
+class RedstoneContractReader extends ContractReader {
```
thiagocarvalhodev, line 10: As `ContractReader` only defines the function `readContract`, Consider implement the `ContractReader` interface with the keyword `implements`.

lib/pst/contract_readers/smartweave_contract_reader.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:ardrive/pst/contract_reader.dart';
+import 'package:ardrive/services/pst/implementations/pst_web.dart'
+    if (dart.library.io) 'package:ardrive/services/pst/implementations/pst_stub.dart'
+    as implementation;
+import 'package:ardrive/types/transaction_id.dart';
+
+class SmartweaveContractReader extends ContractReader {
```
thiagocarvalhodev, line 7: The same about `implements`.

lib/pst/contract_readers/verto_contract_reader.dart
```diff
@@ -0,0 +1,19 @@
+import 'dart:convert';
+
+import 'package:ardrive/pst/contract_reader.dart';
+import 'package:ardrive/types/transaction_id.dart';
+import 'package:http/http.dart' as http;
+
+const cacheUrl = 'v2.cache.verto.exchange';
+
+class VertoContractReader extends ContractReader {
```
thiagocarvalhodev, line 9: The same about implements.

lib/pst/pst_contract_data_builder.dart
```diff
@@ -0,0 +1,375 @@
+import 'package:ardrive/pst/pst_contract_data.dart';
+import 'package:ardrive/types/arweave_address.dart';
+import 'package:equatable/equatable.dart';
+
+class CommunityContractDataBuilder {
+  final Map _rawData;
+
+  CommunityContractDataBuilder(Map jsonData) : _rawData = jsonData;
+
+  CommunityContractData parse() {
```
thiagocarvalhodev, line 10: What do you think of calling it `build()`?

lib/pst/pst_contract_data_builder.dart
```diff
@@ -0,0 +1,375 @@
...
+          nays: nays,
+          voted: voted,
+          start: start,
+          totalWeight: totalWeight,
+          recipient: recipient,
+          qty: qty,
+          lockLength: lockLength);
+    }).toList(growable: false);
+    final CommunityContractSettings settings = CommunityContractSettings(
+      quorum: rawSettings['quorum'].toDouble(),
```
thiagocarvalhodev, line 56: Try to call a function on a dynamic is not safe. You can achieve the same result casting `as double` and will let the compiler knows what type we are getting instead of dynamic.

lib/pst/pst_contract_data_builder.dart
matibat: I'd say it makes more sense +1

lib/pst/pst_contract_data_builder.dart
matibat: TODO: make CommunityContractSettings::quorum to be of type num instead of double and remove the toDouble call

lib/pst/community_oracle.dart
matibat: In that case, adds is a key of vault, and it is an array. Would you say it is needed to check for null there anyways (even knowing that the value is a non-null array)?

lib/pst/pst_contract_data_builder.dart
matibat: The thing here was that sometimes I got a double, sometimes an integer. And none of both extends the other (but num)

lib/pst/ardrive_contract_oracle.dart
thiagocarvalhodev: Does not! Basically you pass a function that will be retried if it throws something.

```dart
final response = await retry(
  // will retry this function
  () {
    throws Exception();
  },
  // Retry on SocketException or TimeoutException
  retryIf: (e) => e is SocketException || e is TimeoutException,
);

File: lib/pst/pst_contract_data_builder.dart (hunk 23 above)
Comment (thiagocarvalhodev) [reply lib/pst/pst_contract_data_builder.dart:56]: You can set the property to `num`, and the client of this class uses it as wanted.

File: lib/pst/pst_contract_data_builder.dart (hunk 22 above)
Comment (thiagocarvalhodev) [inline lib/pst/pst_contract_data_builder.dart:10]: Consider spliting the `parse` function into smaller functions.

File: lib/types/transaction_id.dart [hunk 24]
```diff
@@ -0,0 +1,33 @@
+import 'package:equatable/equatable.dart';
+
+final txIdRegex = RegExp(r'^(\w|-){43}$');
+
+class TransactionID extends Equatable {
```
thiagocarvalhodev, line 5: @matibat , the use of this class is confusing to me.

Its usage is correct, for instance 

```dart
  Future<dynamic> readContract(TransactionID txId) async {
    final apiUrl = Uri.https(cacheUrl, '/$path/$txId');
    final response = await http.get(apiUrl);
    final data = response.body;
    final Map dataAsJson = jsonDecode(data);

    return dataAsJson['state'];
  }
```
The problem is that we can only get the value if we transform it to String. Is that common in JS / TS?

thiagocarvalhodev: But TBH I liked that! I was just confused.

lib/pst/utils.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:math';
+
+import '../types/arweave_address.dart';
+
+ArweaveAddress? weightedRandom(
+  Map<ArweaveAddress, double> dict, {
+  double? testingRandom, // for testing purposes only
```
thiagocarvalhodev, line 7: Consider pass the `Random` object by dependency injection instead of this test variable.

thiagocarvalhodev (commented):
Almost finishing! Reviewed all implementations. Just need to review the tests.

test/pst/community_oracle_test.dart
```diff
@@ -0,0 +1,73 @@
...
+
+      myContractOracle = ContractOracle(myContractReader);
+      myCommunityOracle = CommunityOracle(myContractOracle);
+    });
+
+    group('getCommunityWinstonTip method', () {
+      test('returns a percentage of the actual amount', () async {
+        final tipAmount = await myCommunityOracle
+            .getCommunityWinstonTip(Winston(BigInt.from(999999999999)));
+        expect(tipAmount, Winston(BigInt.from(149999999999)));
```
thiagocarvalhodev, line 43: nit: add the percentage computation to be clear what this function does.

```dart
final wiston = Winston(BigInt.from(999999999999));
final tipAmount = await myCommunityOracle
             .getCommunityWinstonTip(wiston);
expect(tipAmount, Winston(BigInt.from(wiston * 0.15)));

File: test/pst/community_oracle_test.dart [hunk 27]
```diff
@@ -0,0 +1,73 @@
...
+        expect(tipAmount, Winston(BigInt.from(10000000)));
+      });
+    });
+
+    group('selectTokenHolder method', () {
+      const double randomA = .2;
+      const double randomB = 1;
+
+      test('returns a valid address', () async {
+        final addr =
```
thiagocarvalhodev, line 58: Got it why you need the random here! With the Random class you can achieve the same result, but LGTM.

lib/pst/ardrive_contract_oracle.dart
thiagocarvalhodev, line 35: We can use the `for` here

```dart
for (ContractOracle oracle in _contractOracles) {
  final data = await _getContractWithRetries(oracle);
  if(data != null {
    return data;
  }
}

**Review by thiagocarvalhodev — COMMENTED** [review]
Finished review! Left some comments.

File: lib/pst/ardrive_contract_oracle.dart (hunk 16 above)
Comment (matibat) [reply lib/pst/ardrive_contract_oracle.dart:52]: Oops! My methods does not throw, but returns null. I think I can work around that anyways

File: lib/pst/community_oracle.dart (hunk 18 above)
Comment (matibat) [reply lib/pst/community_oracle.dart]: ca28104823132a43622d3253bb9ae8528e970f9d

File: lib/pst/contract_readers/redstone_contract_reader.dart (hunk 19 above)
Comment (matibat) [reply lib/pst/contract_readers/redstone_contract_reader.dart]: 53d31827faf5d5d5c582af312c18ffa1ba8645ce

File: lib/pst/contract_readers/smartweave_contract_reader.dart (hunk 20 above)
Comment (matibat) [reply lib/pst/contract_readers/smartweave_contract_reader.dart]: 53d31827faf5d5d5c582af312c18ffa1ba8645ce

File: lib/pst/contract_readers/verto_contract_reader.dart (hunk 21 above)
Comment (matibat) [reply lib/pst/contract_readers/verto_contract_reader.dart]: 53d31827faf5d5d5c582af312c18ffa1ba8645ce

File: lib/pst/pst_contract_data_builder.dart (hunk 22 above)
Comment (matibat) [reply lib/pst/pst_contract_data_builder.dart]: b3e402e28e34e918b2659e8a316058fed01dcef9

File: lib/pst/pst_contract_data_builder.dart (hunk 23 above)
Comment (matibat) [reply lib/pst/pst_contract_data_builder.dart]: c01737b11a487b000b3939c395812ae5f6833e48

File: lib/types/transaction_id.dart (hunk 24 above)
Comment (matibat) [reply lib/types/transaction_id.dart:5]: > Is that common in JS / TS?

It is not! But we've made the types to validate the input.
Maybe that type could extend String directly... At least the toString function does the trick

File: test/pst/community_oracle_test.dart (hunk 26 above)
Comment (matibat) [reply test/pst/community_oracle_test.dart]: ceb2b8d2b4f1e65b76523e2ce3e88df821e5d474

File: test/pst/community_oracle_test.dart (hunk 27 above)
Comment (matibat) [reply test/pst/community_oracle_test.dart:66]: The thing is that we don't want random behavior in the tests. We do want it to be deretministic 😎

File: lib/pst/ardrive_contract_oracle.dart (hunk 17 above)
Comment (matibat) [reply lib/pst/ardrive_contract_oracle.dart]: Way way cleaner! 6de90253e0e211ac03f880c1945ad01413e48684

File: lib/pst/ardrive_contract_oracle.dart (hunk 16 above)
Comment (matibat) [reply lib/pst/ardrive_contract_oracle.dart]: Applied 314d44c99b9e8cabf4e19328a17288bef74bf43d

File: lib/pst/ardrive_contract_oracle.dart (hunk 17 above)
Comment (matibat) [reply lib/pst/ardrive_contract_oracle.dart]: This part is not retrying, it's only iterating the passed contract readers

File: lib/pst/ardrive_contract_oracle.dart (hunk 16 above)
Comment (matibat) [reply lib/pst/ardrive_contract_oracle.dart]: Hmmm duplicate comment here. Already addressed

File: lib/pst/pst_contract_data_builder.dart (hunk 22 above)
Comment (matibat) [reply lib/pst/pst_contract_data_builder.dart]: Here! d79f448ab48c8e9e64d9e8056243ce43f0d8b5ba

File: lib/pst/ardrive_contract_oracle.dart (hunk 16 above)
Comment (thiagocarvalhodev) [reply lib/pst/ardrive_contract_oracle.dart]: Awesome! Thanks!

File: lib/types/transaction_id.dart (hunk 24 above)
Comment (thiagocarvalhodev) [reply lib/types/transaction_id.dart:5]: For me LGTM as it is!

File: lib/pst/ardrive_contract_oracle.dart (hunk 17 above)
Comment (thiagocarvalhodev) [reply lib/pst/ardrive_contract_oracle.dart]: Great! You can use `for` here.

**Review by thiagocarvalhodev — APPROVED** [review]
Just left a small comment about changing the `while` approach with `for`. But it's good to go! thanks for addressing all of my comments.

File: lib/pst/ardrive_contract_oracle.dart (hunk 17 above)
Comment (matibat) [reply lib/pst/ardrive_contract_oracle.dart]: That was addressed in another comment - 6de90253e0e211ac03f880c1945ad01413e48684

**Review by karlprieb — APPROVED** [review]

File: lib/types/winston.dart [hunk 28]
```diff
@@ -0,0 +1,88 @@
+import 'package:equatable/equatable.dart';
+
+class Winston extends Equatable {
+  final BigInt _amount;
+
+  @override
+  List<Object?> get props => [_amount];
+
+  Winston(BigInt amount) : _amount = amount {
+    if (amount < BigInt.zero || !amount.isValidInt) {
```
matibat, line 10: The getter isValidInt would return false if the value is bigger than the max safe integer value, it is not a "is integer" check.
The values of BigInt instances are already integer.

```suggestion
    if (amount < BigInt.zero) {
```

matibat (commented):
Left one single comment. Imma apply the change and manually test the PR

karlprieb (approved):

PR #624, opened 2022-08-09
PE 1719/ Share Page Activity & Details Panel

javdhu:
Redesigns file share page to include activity and details panel.

### DIFF
--- a/lib/blocs/shared_file/shared_file_cubit.dart
+++ b/lib/blocs/shared_file/shared_file_cubit.dart
@@ -1,4 +1,5 @@
 import 'package:ardrive/entities/entities.dart';
+import 'package:ardrive/entities/string_types.dart';
 import 'package:ardrive/l11n/validation_messages.dart';
 import 'package:ardrive/models/models.dart';
 import 'package:ardrive/services/services.dart';
@@ -8,6 +9,7 @@ import 'package:equatable/equatable.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:reactive_forms/reactive_forms.dart';
+import 'package:url_launcher/url_launcher.dart';
 
 part 'shared_file_state.dart';
 
@@ -84,6 +86,7 @@ class SharedFileCubit extends Cubit<SharedFileState> {
     oldestRevision = fileEntities.first.toRevision(
       performedAction: RevisionAction.create,
     );
+    // Remove oldest revision from list so we dont compute it again.
     fileEntities.removeAt(0);
     final revisions = <FileRevision>[oldestRevision];
     for (final entity in fileEntities) {
@@ -103,8 +106,8 @@ class SharedFileCubit extends Cubit<SharedFileState> {
 
       revisions.add(revision);
     }
-
-    return revisions;
+    // Reverse list so it is in chronological order.
+    return revisions.reversed.toList();
   }
 
   Future<void> loadFileDetails(SecretKey? fileKey) async {
@@ -125,6 +128,12 @@ class SharedFileCubit extends Cubit<SharedFileState> {
       return;
     }
     emit(SharedFileNotFound());
+  }
+
+  Future<void> launchPreview(TxID dataTxId) {
+    return launchUrl(
+      Uri.parse('${_arweave.client.api.gatewayUrl}/$dataTxId'),
+    );
   }
 
   void submit() async {
--- a/lib/pages/shared_file/shared_file_page.dart
+++ b/lib/pages/shared_file/shared_file_page.dart
@@ -81,9 +81,9 @@ class SharedFilePage extends StatelessWidget {
                           ListTile(
                             contentPadding: EdgeInsets.zero,
                             leading: const Icon(Icons.text_snippet),
-                            title: Text(state.fileRevisions.last.name),
+                            title: Text(state.fileRevisions.first.name),
                             subtitle:
-                                Text(filesize(state.fileRevisions.last.size)),
+                                Text(filesize(state.fileRevisions.first.size)),
                           ),
                           const SizedBox(height: 24),
                           ElevatedButton.icon(
@@ -91,7 +91,7 @@ class SharedFilePage extends StatelessWidget {
                             label: Text(appLocalizationsOf(context).download),
                             onPressed: () => promptToDownloadSharedFile(
                               context: context,
-                              fileId: state.fileRevisions.last.fileId,
+                              fileId: state.fileRevisions.first.fileId,
                               fileKey: state.fileKey,
                             ),
                           ),
--- a/lib/blocs/drive_detail/drive_detail_cubit.dart
+++ b/lib/blocs/drive_detail/drive_detail_cubit.dart
@@ -171,7 +171,8 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
 
   Future<void> launchPreview(TxID dataTxId) {
     return launchUrl(
-        Uri.parse('${_config.defaultArweaveGatewayUrl}/$dataTxId'));
+      Uri.parse('${_config.defaultArweaveGatewayUrl}/$dataTxId'),
+    );
   }
 
   void sortFolder({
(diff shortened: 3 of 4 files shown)


karlprieb (commented):
I think we can have, at least, some drafts for tests and a tech debt ticket to implement them next week when we will be able to run tests again.

When we call `getAllFileEntitiesWithId` we could get the `owner` of the first tx and check if the owner is the same when we call `getLatestFileEntityWithId`. So we ensure that the file shared is from the same wallet that first uploaded it.

karlprieb (commented):
can we check owner on `getLatestFileEntityWithId` as well?
Also do we have tests for `getLatestFileEntityWithId`?

test/services/arweave/arweave_service_test.dart
```diff
@@ -0,0 +1,22 @@
...
+
+const gatewayUrl = 'https://arweave.net';
+void main(List<String> args) {
+  group('Arweave Service Tests', () {
+    final arweave = ArweaveService(
+      Arweave(
+        gatewayUrl: Uri.parse(gatewayUrl),
+      ),
+    );
+    test('AllFileEntitiesWithId returns all the file entities for a known Id',
```
karlprieb, line 13: add a test for when `getAllFileEntitiesWithId` returns null
why did you add `expect(true, isTrue);`?

lib/models/file_revision.dart
```diff
@@ -35,8 +35,9 @@ extension FileEntityExtensions on FileEntity {
   /// Converts the entity to an instance of [FileRevisionsCompanion].
   ///
   /// This requires a `performedAction` to be specified.
-  FileRevisionsCompanion toRevisionCompanion(
-          {required String performedAction}) =>
+  FileRevisionsCompanion toRevisionCompanion({
+    required String performedAction,
```
karlprieb, line 39: I know that's not the scope of this ticket, but can we have an enum here instead of string?

test/services/arweave/arweave_service_test.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:math';
+
+import 'package:ardrive/services/services.dart';
+import 'package:arweave/arweave.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+const gatewayUrl = 'https://arweave.net';
+void main(List<String> args) {
+  group('Arweave Service Tests', () {
```
karlprieb, line 9: this test is not reliable... it depends on network to work and will break in the future.
We should mock http calls here with `MockClient` from `http/testing`.

lib/models/file_revision.dart
matibat: You can also simply use the getter `MyEnum.EnumValue.name` which does the same without having to split it. See: [name property](https://api.flutter.dev/flutter/dart-core/EnumName/name.html).

test/services/arweave/arweave_service_test.dart
```diff
@@ -0,0 +1,31 @@
...
+    );
+    test('AllFileEntitiesWithId returns all the file entities for a known Id',
+        () async {
+      const knownFileId = '0f029ecf-6593-4942-ad07-a57401e3861c';
+      const knownRevisionCount = 9;
+      final fileEntities = await arweave.getAllFileEntitiesWithId(knownFileId);
+      expect(fileEntities?.length, equals(knownRevisionCount));
+    });
+    test('AllFileEntitiesWithId returns null for invalid Id', () async {
+      final invalidFileId = Random().nextInt(1000).toString();
```
matibat, line 23: Let's use an explicit value instead of making use of Random to make the test deterministic

test/services/arweave/arweave_service_test.dart
matibat: We are currently injecting an `Arweave` intense into `ArweaveService`, if we also injected the `ArtemisClient` we could mock both

matibat (commented):
Left some comments

lib/models/file_revision.dart
karlprieb: Can you create a tech debt for it?

lib/pages/shared_file/shared_file_page.dart
```diff
@@ -17,91 +20,142 @@ class SharedFilePage extends StatelessWidget {
...
+                              label: Text(appLocalizationsOf(context).download),
+                              onPressed: () => promptToDownloadSharedFile(
+                                context: context,
+                                fileId: state.fileRevisions.last.fileId,
+                                fileKey: state.fileKey,
+                              ),
+                            ),
+                            const SizedBox(height: 16),
+                            _buildReturnToAppLink(context),
+                          } else if (state is SharedFileNotFound) ...{
```
matibat, line 100: What does the "..." do in here? `if() ...{}`
Is it destructuring?

lib/services/arweave/arweave_service.dart
```diff
@@ -533,6 +535,66 @@ class ArweaveService {
...
+        FirstFileEntityWithIdOwnerQuery(
+            variables: FirstFileEntityWithIdOwnerArguments(fileId: fileId)));
+
+    if (firstOwnerQuery.data!.transactions.edges.isEmpty) {
+      return null;
+    }
+
+    final fileOwner =
+        firstOwnerQuery.data!.transactions.edges.first.node.owner.address;
+    while (true) {
```
matibat, line 554: I always fear infinite loops. Could we track a concise exit condition instead of using break?
Such as...
```ts
var hasNextPage = true;
while(hasNextPage) {
  ...
  hasNextPage = !queryEdges.isEmpty && pageInfo.hasNextPage;
```

matibat: Or maybe a `do {} while ()` 🤔

test/services/arweave/arweave_service_test.dart
matibat: Perfect! Now the class is testable!
We're just missing the update in the tests to remove any call to the network

matibat (commented):
Left some comments - _TODO_ I still want to manually test it

lib/blocs/shared_file/shared_file_cubit.dart
```diff
@@ -56,20 +60,73 @@ class SharedFileCubit extends Cubit<SharedFileState> {
...
+  Future<List<FileRevision>> computeRevisionsFromEntities(
+    List<FileEntity> fileEntities,
+  ) async {
+    late FileRevision oldestRevision;
+
+    fileEntities.sort((a, b) => a.createdAt.compareTo(b.createdAt));
+    oldestRevision = fileEntities.first.toRevision(
+      performedAction: RevisionAction.create,
+    );
+    fileEntities.removeAt(0);
```
thiagocarvalhodev, line 87: Why do we are removing the first index? Consider create a private function or adding a comment to explain that.

lib/blocs/shared_file/shared_file_cubit.dart
```diff
@@ -56,20 +60,73 @@ class SharedFileCubit extends Cubit<SharedFileState> {
...
-    } else {
-      final file = await _arweave.getLatestFileEntityWithId(fileId, fileKey);
+      return;
+    }
+    final allEntities = await _arweave.getAllFileEntitiesWithId(
+      fileId,
+      fileKey,
+    );
+    if (allEntities != null) {
+      {
```
thiagocarvalhodev, line 122: Remove this extra `{`

test/test_utils/fake_data.dart
```diff
@@ -1,19 +1,19 @@
...
+
+// Entity Sample Data
+
+const testEntityName = 'test-name';
+const testEntitySize = 10;
+const testEntityDataContentType = 'test-data-mime-type';
+const testEntityDataTxId = 'FxgiSM3JOAsjnZF-Yzv2z7h-Io4K6GmCarhQ0qAUL6I';
+final testEntityId = const Uuid().v4();
+final driveId = const Uuid().v4();
+final rootFolderId = const Uuid().v4();
```
thiagocarvalhodev, line 19: Why add `final` to a `const`?

thiagocarvalhodev: OH, got it!

karlprieb (dismissed):

test/test_utils/fake_data.dart
matibat, line 19: These inputs are still not deterministic but random x.x
Let's try to use an explicit value for each

test/services/arweave/arweave_service_test.dart
matibat: The inputs are still random values - see my review comment

matibat (commented):
Thanks for addressing my previous comments. Left a comment regarding the tests

karlprieb (approved):

PR #634, opened 2022-08-15
PE-1944: Multiselect Files + Folders and Move

javdhu:
Adds the ability to move multiple files and folders at once. Hold shift to select items.

Suggested Review Order:

- lib/pages/drive_detail/
- lib/blocs/drive_detail/
- lib/blocs/fs_entry_move/
- lib/blocs/keyboard_listener/
- lib/components/fs_entry_move_form.dart
- lib/components/keyboard_handler.dart

--- Releases ---
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/26irdbl6rl4b0
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/3rk3obpc788ro

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -1401,5 +1401,53 @@
   "invalidWalletFileDescription": "The file you have chosen is invalid.",
   "@invalidWalletFileDescription": {
     "description": "Invalid wallet file description"
+  },
+  "biometricLoginEnabled": "Biometric Login Enabled",
+  "@biometricLoginEnabled": {
+    "description": "Biometric Login is Enabled."
+  },
+  "biometricLoginDisabled": "Biometric Login Disabled",
+  "@biometricLoginDisabled": {
+    "description": "Biometric Login is Disabled."
+  },
+  "enableBiometricLogin": "Enable Biometric Login",
+  "@enableBiometricLogin": {
+    "description": "Enable Biometric Login"
+  },
+  "enableBiometricsOnSettings": "This function is currently disabled on your device. To use it, please go to your device Settings and enable biometric authentication.",
+  "@enableBiometricsOnSettings": {
+    "description": "Open settings to enable biometric"
+  },
+  "loginUsingBiometricCredential": "Login using your biometric credential",
+  "@loginUsingBiometricCredential": {
+    "description": "Login using your biometric credential"
+  },
+  "goToDeviceSettings": "Go to Settings",
+  "@goToDeviceSettings": {
+    "description": "Go to device settings"
+  },
+  "biometricTemporarilyDisabled": "Biometric login temporarily disabled due to too many unsuccessful attempts.",
+  "@biometricTemporarilyDisabled": {
+    "description": "Biometric is locked out"
+  },
+  "deviceNotSupportBiometrics": "This device does not support the use of biometrics.",
+  "@deviceNotSupportBiometrics": {
+    "description": "Biometric is not supported"
+  },  
+  "biometricsDisabledOnYourDevice": "This function is currently disabled on your device. To use it, please go to your device Settings and enable biometric authentication.",
+  "@biometricsDisabledOnYourDevice": {
+    "description": "Biometric is disabled"
+  },
+  "biometricsPasscodeNotSetAndroid": "Your device requires a PIN/pattern/password in order to use biometrics. Please create one in order to proceed.",
+  "@biometricsPasscodeNotSetAndroid": {
+    "description": "Passcode it not set on Android"
+  },
+  "biometricsPasscodeNotSetIOS": "Your device requires a passcode in order to use biometrics. Please create one by going to Settings > Face ID & Passcode in order to proceed.",
+  "@biometricsPasscodeNotSetIOS": {
+    "description": "Passcode it not set on Android"
+  },
+  "biometricsPermanentlyLockedOut": "Biometric login disabled due to too many unsuccessful attempts. Please enter your PIN/passcode to unlock biometric functionality.",
+  "@biometricsPermanentlyLockedOut": {
+    "description": "Biometrics permanently locked out"
   }
 }
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -9,8 +9,10 @@ import 'package:ardrive/pst/contract_oracle.dart';
 import 'package:ardrive/pst/contract_readers/redstone_contract_reader.dart';
 import 'package:ardrive/pst/contract_readers/smartweave_contract_reader.dart';
 import 'package:ardrive/pst/contract_readers/verto_contract_reader.dart';
+import 'package:ardrive/services/authentication/biometric_authentication.dart';
 import 'package:ardrive/utils/html/html_util.dart';
 import 'package:ardrive/utils/local_key_value_store.dart';
+import 'package:ardrive/utils/secure_key_value_store.dart';
 import 'package:arweave/arweave.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
@@ -19,6 +21,8 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:flutter_localizations/flutter_localizations.dart';
 import 'package:flutter_portal/flutter_portal.dart';
+import 'package:flutter_secure_storage/flutter_secure_storage.dart';
+import 'package:local_auth/local_auth.dart';
 
 import 'blocs/blocs.dart';
 import 'models/models.dart';
@@ -42,9 +46,7 @@ void main() async {
   );
 
   arweave = ArweaveService(
-    Arweave(
-      gatewayUrl: Uri.parse(config.defaultArweaveGatewayUrl!),
-    ),
+    Arweave(gatewayUrl: Uri.parse(config.defaultArweaveGatewayUrl!)),
   );
   if (kIsWeb) {
     refreshHTMLPageAtInterval(const Duration(hours: 12));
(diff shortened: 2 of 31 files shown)


lib/pages/drive_detail/components/drive_detail_data_table_source.dart
```diff
@@ -100,12 +107,19 @@ DataRow _buildFolderRow({
...
     onSelectChanged: (_) => onPressed(),
     selected: selected,
     index: index,
     cells: [
+      if (checkBoxEnabled)
+        DataCell(
+          AbsorbPointer(child: Checkbox(value: selected, onChanged: (_) {})),
+        )
+      else
+        DataCell(Container()),
```
matibat, line 122: If we put here and below a SizedBox instead of Container we can get rid of the flicker effect: the table content moves to the right a couple of pixels when you press Ctrl/Cmd

lib/components/keyboard_handler.dart
```diff
@@ -0,0 +1,80 @@
...
+                      event.isKeyPressed(LogicalKeyboardKey.metaLeft) ||
+                          event.isKeyPressed(LogicalKeyboardKey.metaRight);
+                } else {
+                  ctrlMetaKeyPressed =
+                      event.isKeyPressed(LogicalKeyboardKey.controlLeft) ||
+                          event.isKeyPressed(LogicalKeyboardKey.controlRight);
+                }
+
+                // detect if ctrl + v or cmd + v is pressed
+                if (ctrlMetaKeyPressed) {
```
matibat, line 42: Is there a way to make use of `lib/utils/html/html_util.dart` to turn this false when the user leaves the tab? 

Not a blocker, but when I press CMD + TAB it switches to another window and the checkboxes are still being shown

matibat (commented):
I didn't review but just played with the app. Thanks for addressing the details we've discussed on Slack, everything seems to be working as expected.

I left two comments regarding
- the flickering effect and
- the user using CMD/CTRL + TAB to leave the browser window

karlprieb (commented):
Why don't we listen to keyboard events only on the data table? I don't think it makes sense to read for a specific keyboard event on all app if we just use it on the drive data table. We should start listening the keyboard event when data table is loaded and remove the listener when we're not displaying it.

lib/pages/drive_detail/components/drive_detail_data_table_source.dart
```diff
@@ -119,7 +119,7 @@ DataRow _buildFolderRow({
           AbsorbPointer(child: Checkbox(value: selected, onChanged: (_) {})),
         )
       else
-        DataCell(Container()),
+        const DataCell(SizedBox(width: 24)),
```
matibat, line 122: It seems that the behavior didn't change. Please take a look to #660: it uses a size of 32 instead of 24. That seems to fix the flicker animation

matibat: **Current behavior**
https://user-images.githubusercontent.com/22564541/187264200-abf8b7cb-0ddf-4b84-82cf-bb6945735754.mov

**Fixed**
https://user-images.githubusercontent.com/22564541/187264373-0923cf54-8959-45ea-a218-8a066220b2e7.mov

matibat: [88eb896](https://github.com/ardriveapp/ardrive-web/pull/634/commits/88eb896a0823be71312e43e8f3c8d3dbe32c249a)

lib/components/keyboard_handler.dart
matibat: Yup, I agree. We don't have to address that nit in this PR

lib/blocs/fs_entry_move/fs_entry_move_bloc.dart
```diff
@@ -0,0 +1,254 @@
...
+
+  @override
+  Future<void> close() {
+    return super.close();
+  }
+
+  @override
+  void onError(Object error, StackTrace stackTrace) {
+    super.onError(error, stackTrace);
+  }
```
matibat, line 253: These overrides are unnecessary. They are only calling the super, and that's equivalent to do nothin'
```suggestion
```

lib/blocs/keyboard_listener/keyboard_listener_event.dart
```diff
@@ -0,0 +1,18 @@
+part of 'keyboard_listener_bloc.dart';
+
+abstract class KeyboardListenerEvent extends Equatable {
+  const KeyboardListenerEvent();
+
+  @override
+  List<Object> get props => [];
```
matibat, line 7: We don't need this attribute for the abstract class, it's being overridden in the implementation

lib/l10n/app_en.arb
```diff
@@ -1363,5 +1363,19 @@
...
+  },
+  "itemMoveNameConflict": "The following item(s) could not be moved because '{destinationFolderName}' contains item(s) with the same name.",
+  "@itemMoveNameConflict": {
+    "description": "Description for dialog to skip conflicting files or cancel",
+    "placeholders": {
+      "destinationFolderName": {
+        "type": "String",
+        "example": "Folder A"
+      }
+    }
```
matibat, line 1379: I don't know if we did consider using plurals here, but we do have the information needed to do it
```json
  "itemMoveNameConflict": "{itemsCount,plural, =1{The following item could not be moved because '{destinationFolderName}' contains a item with the same name.} =other{The following items could not be moved because '{destinationFolderName}' contains items with the same name.}}",
  "@itemMoveNameConflict": {
    "description": "Description for dialog to skip conflicting files or cancel",
    "placeholders": {
      "itemsCount": {
        "type": "int",
        "format": "decimalPattern",
        "example": "2"
      },
      "destinationFolderName": {
        "type": "String",
        "example": "Folder A"
      }
    }
```

```dart
appLocalizationsOf(context)
  .itemMoveNameConflict(state.conflictingItems.length, state.folderInView.name)
```

_Not a blocker, but a nice-to-have_

lib/utils/html/implementations/html_web.dart
```diff
@@ -1,35 +1,36 @@
...
+    }
+  });
+}
+
+Future<void> closeVisibilityChangeStream() async =>
+    await _onVisibilityChangeStream.cancel();
+
+void refreshPageAtInterval(Duration duration) {
+  Future.delayed(duration, () {
+    window.location.reload();
```
matibat, line 24: Nit: DRY
```suggestion
    reload();
```

pubspec.yaml
```diff
@@ -57,14 +57,17 @@ dependencies:
   flutter_svg: ^1.1.3
   ardrive_io:
     path: ./ardrive_io
+  bloc_concurrency: ^0.2.0
+  universal_html: ^2.0.8
+  bloc_test: ^9.1.0
```
matibat, line 62: This one is a dev dep, right?

matibat (commented):
I've played a little with the deployment and it's working as expected now. Thanks for applying the fixes 🚀 

Left some comments regarding the code

matibat:
I've tested moving stuff around and found this

### The modal with the spinning icon don't appear when moving

https://user-images.githubusercontent.com/22564541/187314831-0bfa01f5-3e12-4fa6-99ce-156e3e1570f9.mov

### The table column headers don't shrink when Details is opened
(doesn't matter if holding CMD)
Screen Shot 2022-08-29 at 20 01 39

matibat (dismissed):
Thanks for the fixes! It works as expected on my side

lib/blocs/data_export/data_export_cubit.dart
```diff
@@ -1,73 +1,76 @@
...
+    required String gatewayURL,
+  })  : _driveDao = driveDao,
+        _gatewayURL = gatewayURL,
+        super(DataExportInitial());
+
+  Future<String> getFilesInDriveAsCSV(String driveId) async {
+    final files = await _driveDao
+        .filesInDriveWithRevisionTransactions(driveId: driveId)
+        .get();
+    final export = <List<String>>[
```
karlprieb, line 28: use `const`

lib/components/keyboard_handler.dart
```diff
@@ -0,0 +1,82 @@
...
+  final Widget child;
+  const KeyboardHandler({Key? key, required this.child}) : super(key: key);
+
+  @override
+  State<KeyboardHandler> createState() => _KeyboardHandlerState();
+}
+
+class _KeyboardHandlerState extends State<KeyboardHandler> {
+  final _focusTable = FocusNode();
+  var checkboxEnabled = false;
```
karlprieb, line 17: use `bool` instead of `var` here.

lib/components/keyboard_handler.dart
```diff
@@ -0,0 +1,82 @@
...
+            child: widget.child,
+          );
+        },
+      ),
+    );
+  }
+}
+
+bool isApple(String userAgent) {
+  final platforms = [
```
karlprieb, line 67: use `const`

lib/l10n/app_en.arb
karlprieb: great nice to have

lib/main.dart
```diff
@@ -33,11 +37,20 @@ void main() async {
...
 
   arweave = ArweaveService(
-      Arweave(gatewayUrl: Uri.parse(config.defaultArweaveGatewayUrl!)));
-  refreshHTMLPageAtInterval(const Duration(hours: 12));
+    Arweave(
+      gatewayUrl: Uri.parse(config.defaultArweaveGatewayUrl!),
+    ),
+  );
+  if (kIsWeb) {
+    refreshHTMLPageAtInterval(const Duration(hours: 12));
```
karlprieb, line 45: Just curious about why we're doing that. :)

karlprieb: It make sense :D Thank you for explaining!

lib/blocs/data_export/data_export_cubit.dart
karlprieb: Sorry... it should be `final` as you're mutating the list.

karlprieb (approved):

karlprieb (dismissed):

karlprieb (approved):

PR #636, opened 2022-08-15
feat(localization es): adds translation for infernoIsInFullSwing PE-1752

matibat:


thiagocarvalhodev (approved):
I didn't understand a word, but I believe in your potential!

PR #633, opened 2022-08-15
chore(iOS configs) configure iOS app PE-2065

thiagocarvalhodev:
- Update Podfile with the supported iOS version
- Apply settings upgrade from XCode (especially related to an older iOS version, update packages, and fix deprecates)
- Use the correct app name
- Use the correct bundle ID
- Add permissions
- Define device screen orientation

PR #638, opened 2022-08-16
feat(launcher icon) add launch icon for android and iOS

thiagocarvalhodev:
Provide a launcher icon that works on iPhone and Android-based Smartphones of different sizes and also iPads and Tablets.

It should be merged after #633 #635


matibat:
Beware: this branch is carrying commits from other branches.

Let's try to remove those by either:
- Chaining the PRs' bases (e.g. dev > iOS config > Android config > Launch icons)
- removing the commits (this might imply a --force push 👎 )

matibat:
I don't know how easily this can be accomplished with Flutter, but we can use [Adaptive icons](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive) for Android which uses SVG images instead of multi-sized PNG files

matibat (approved):
LGTM

PR #639, opened 2022-08-17
fix: correct public file size limit PE-2123

jdaev:
Also PE-2122 I guess?


agsuy (approved):
LGTM!

thiagocarvalhodev (approved):

matibat:
Downloads fails on Chrome:
- You cannot download a file of size 1.25 GiB (RangeError: Array buffer allocation failed)
- You can download a file of size 1.25 GB and 1.25 GB plus a byte

PR #641, opened 2022-08-17
fix(dropzone): validate if the dropped entity is not a folder

thiagocarvalhodev:
Use `XFile` to validate if the file is not a Folder.

### DIFF
--- a/lib/pages/drive_detail/components/drive_file_drop_zone.dart
+++ b/lib/pages/drive_detail/components/drive_file_drop_zone.dart
@@ -75,7 +75,7 @@ class DriveFileDropZoneState extends State<DriveFileDropZone> {
       try {
         final htmlUrl = await controller.createFileUrl(htmlFile);
 
-        /// We use xFile to get the bytes and also validate if it is a file
+        // We use xFile to get the bytes and also validate if it is a file
         final bytes = await XFile(htmlUrl).readAsBytes();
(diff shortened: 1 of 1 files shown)


lib/pages/drive_detail/components/drive_file_drop_zone.dart
```diff
@@ -70,17 +71,17 @@ class _DriveFileDropZoneState extends State<DriveFileDropZone> {
...
-      final selectedFiles = [
-        UploadFile(ioFile: ioFile, parentFolderId: folderId)
-      ];
+      final selectedFiles = <UploadFile>[];
       try {
-        //This is the only way to know whether the dropped file is a folder
-        await ioFile.readAsBytes();
+        final htmlUrl = await controller.createFileUrl(htmlFile);
+
+        /// We use xFile to get the bytes and also validate if it is a file
```
arielmelendez, line 78:
```suggestion
        // We use xFile to get the bytes and also validate if it is a file
```

arielmelendez: How does the "validation" work here? Will it simply throw if this is a folder?

thiagocarvalhodev: It will.

arielmelendez (approved):

PR #650, opened 2022-08-23
PE-2152: PST fix in JS interpolation 

matibat:
Change log
- Applies a workaround to the issue with interpolation between JS and Dart: makes the JS method to return a `string` to be parsed in the Dart side


lib/pst/contract_readers/smartweave_contract_reader.dart
```diff
@@ -7,6 +7,6 @@ import 'package:ardrive/types/transaction_id.dart';
 class SmartweaveContractReader implements ContractReader {
   @override
   Future<dynamic> readContract(TransactionID txId) async {
-    return implementation.readContract(txId);
+    return await implementation.readContract(txId);
```
karlprieb, line 10: we don't need the `await` keyword if it's the last thing or the only thing (this case) that this method will execute.

web-pst/src/index.ts
```diff
@@ -10,8 +10,9 @@ const arweave = Arweave.init({
   logging: false,
 });
 
-export async function readContractPromise(
+export async function readContractAsStringPromise(
   contractId: string,
 ): Promise<unknown> {
-  return readSmartweaveContract(arweave, contractId);
+  const contract = await readSmartweaveContract(arweave, contractId);
+  return JSON.stringify(contract);
```
karlprieb, line 17: if you're resolving the `Promise` here and returning a `string` this function type is now wrong.

karlprieb: Also the name is wrong as we don't return a `Promise` anymore.

web/js/pst.min.js
```diff
@@ -12,4 +12,4 @@ var pst=function(t,e){"use strict";function r(t){return t&&"object"==typeof t&&"
...
...nction*(){const r=yield t.api.post("graphql",{query:"query Transactions($tags: [TagFilter!]!, $blockFilter: BlockFilter!, $first: Int!, $after: String) {\n    transactions(tags: $tags, block: $blockFilter, first: $first, sort: HEIGHT_ASC, after: $after) {\n      pageInfo {\n        hasNextPage\n      }\n      edges {\n        node {\n          id\n          owner { address }\n          recipient\n          tags {\n            name\n            value\n          }\n          block {\n            height\n            id\n          }\n          fee { winston }\n          quantity { winston }\n          parent { id }\n        }\n        cursor\n      }\n    }\n  }",variables:e});if(200!==r.status)throw new Error(`Unable to retrieve transactions. Arweave gateway responded with status ${r.status}.`);return r.data.data.transactions}))}})),Ht=At.init({host:"arweave.net",port:443,protocol:"https",logging:!1});return t.readContractAsStringPromise=function(t){return o(this,void 0,void 0,(function(){var e;return i(this,(function(r){switch(r.label){case 0:return[4,Ft.readContract(Ht,t)];case 1:return e=r.sent(),[2,JSON.stringify(e)]}}))}))},Object.defineProperty(t,"__esModule",{value:!0}),t}({});
```
karlprieb, line 15: Did you updated smartweavejs?

lib/pst/contract_readers/smartweave_contract_reader.dart
matibat: Nice catch - 62a76f834304af5948ee5ae39d40140baf24cc32

web-pst/src/index.ts
matibat: > if you're resolving the Promise here and returning a string this function type is now wrong.

Take a look to the Function signature: it is async. Every method that is async MUST return a Promise.

> Also the name is wrong as we don't return a Promise anymore.

It is ;)

web/js/pst.min.js
matibat: I didn't. This minified file was simply re-compiled to return Promise<string> instead of the object

matibat: There are no changes in the lock file in this PR

web-pst/src/index.ts
karlprieb: You're right :D

karlprieb (approved):

karlprieb (approved):

PR #651, opened 2022-08-23
fix: fix deriveKey function passing Uint8List

thiagocarvalhodev:
Pass a nonce as Uint8List(0) throws an error to deriveKey function.

We should pass Uint8List(1). 

Related issue on Github: https://github.com/dint-dev/cryptography/issues/42


karlprieb (dismissed):

docs/private_drive_kdf_reference.dart
```diff
@@ -34,7 +34,7 @@ void main() async {
   final driveKey = await kdf.deriveKey(
     secretKey: SecretKey(walletSignature),
     info: utf8.encode(password),
-    nonce: Uint8List(0),
+    nonce: Uint8List(1),
```
matibat, line 37: Would this change what deriveKey returns? Would this break backwards compatibility?

karlprieb: No. Here you have a brief explanation of what nonce is https://www.hypr.com/security-encyclopedia/nonce

karlprieb (approved):

PR #658, opened 2022-08-26
Pe 1990 the survey feedback modal is not centered on mobile

thiagocarvalhodev:


karlprieb (approved):

matibat (approved):
Now being displayed at the center in the Mobile view

PR #661, opened 2022-08-30
PE-2176 Fix Shared File Activity Panel Revision Ordering

javdhu:
Changes the order of revisions to make the oldest revision last.


matibat (approved):
LGTM

PR #662, opened 2022-08-30
PE-2177 Fix Shared File Acitivity Panel Download and Preview Button

javdhu:
Adds functionality back to the download and preview button for each revision.

### DIFF
--- a/lib/blocs/shared_file/shared_file_cubit.dart
+++ b/lib/blocs/shared_file/shared_file_cubit.dart
@@ -1,4 +1,5 @@
 import 'package:ardrive/entities/entities.dart';
+import 'package:ardrive/entities/string_types.dart';
 import 'package:ardrive/l11n/validation_messages.dart';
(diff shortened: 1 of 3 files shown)


lib/blocs/shared_file/shared_file_cubit.dart
```diff
@@ -127,6 +129,12 @@ class SharedFileCubit extends Cubit<SharedFileState> {
     emit(SharedFileNotFound());
   }
 
+  Future<void> launchPreview(TxID dataTxId) {
+    return launchUrl(
+      Uri.parse('${_arweave.client.api.gatewayUrl}/$dataTxId'),
```
matibat, line 134: Should we use the custom gw url? (the one set at LocalStorage)
```dart
      Uri.parse('${_config.defaultArweaveGatewayUrl}/$dataTxId'),
```

thiagocarvalhodev (approved):
LGTM!

PR #663, opened 2022-08-30
Pe 2078 pick image from gallery or camera on mobile

thiagocarvalhodev:
Currently, the ardrive_io package only picks files from the file system. This PR introduces two new capabilities: `Camera` and `Gallery`


ardrive_io/lib/src/io_file.dart
```diff
@@ -190,8 +203,8 @@ class _DataFile implements IOFile {
   }
 }
 
-class _WebXFile implements IOFile {
-  _WebXFile(
+class _XFile implements IOFile {
```
karlprieb, line 206: Lets chose another name for this one. It's not using XFile, just reimplementing its API.

ardrive_io/lib/src/web/web_io.dart
```diff
@@ -24,7 +25,13 @@ class WebIO implements ArDriveIO {
   final FolderPicker _folderPicker;
 
   @override
-  Future<IOFile> pickFile({List<String>? allowedExtensions}) async {
+  Future<IOFile> pickFile(
+      {List<String>? allowedExtensions,
+      FileSource fileSource = FileSource.fileSystem}) async {
+    if (fileSource != FileSource.fileSystem) {
```
karlprieb, line 31: noice

ardrive_io/lib/src/mobile/mobile_io.dart
```diff
@@ -8,33 +8,37 @@ import 'package:mime/mime.dart' as mime;
...
-      {required IOFileAdapter fileAdapter,
-      required FileSaver fileSaver,
-      required IOFolderAdapter folderAdapter})
-      : _fileAdapter = fileAdapter,
-        _fileSaver = fileSaver,
-        _folderAdapter = folderAdapter;
+  MobileIO({
+    required FileSaver fileSaver,
+    required IOFolderAdapter folderAdapter,
+    required FileProviderFactory fileProviderFactory,
```
karlprieb, line 14: Why don't we move this reponsability to IOFileAdapter? When using this abstraction I don't want to think about FileProvider, I just want to use its adapters and them should handle it

thiagocarvalhodev: The `FileProvider` is an abstraction not for `IOFile` but for the source of the file.
The `FileProvider` could be a `GalleryProvider`, `NetworkFileProvider`, or any other source of files.
In this PR we have two: `FilePickerProvider` and `CameraProvider`, which are two different sources of files. If we need to change or add a new source of files e.g from the network, we just need to implement its interface.

ardrive_io/lib/src/io_file.dart
thiagocarvalhodev: [e42ad78](https://github.com/ardriveapp/ardrive-web/pull/663/commits/e42ad78272f91fa570f77dc824a7ea2a2d20d925)

lib/components/file_picker_modal.dart
```diff
@@ -0,0 +1,79 @@
...
+            child: Column(
+              mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+              children: [
+                ListTile(
+                  onTap: () async {
+                    content = await pickFromCamera();
+
+                    Navigator.pop(context);
+                  },
+                  title: const Text('Camera'),
```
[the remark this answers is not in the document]
thiagocarvalhodev: Very good catch!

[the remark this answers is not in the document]
thiagocarvalhodev: [8a0f91d](https://github.com/ardriveapp/ardrive-web/pull/663/commits/8a0f91d6584f761a17a261ab88995217fea17e47)

karlprieb:
We use Provider/Adapter for files, but for folders the Provider is included on IOFolder, let's normalize and split these responsibilities between Provider and Adapter for folders as well.

lib/components/file_picker_modal.dart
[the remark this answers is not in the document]
matibat: Oh! New strings being introduced here. Let's let the PO to know it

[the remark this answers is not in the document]
thiagocarvalhodev: >Oh! New strings being introduced here. Let's let the PO to know it
She knows and they were validated!

karlprieb (approved):

karlprieb (dismissed):

karlprieb (dismissed):

ardrive_io/lib/src/file_provider.dart
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    throw ActionCanceledException();
+  }
+
+  FileType _getFileType(
+    FileSource fileSource,
+    bool hasExtensionFilters,
+  ) {
+    switch (fileSource) {
```
matibat, line 116: This switch..case has no default - although I guess it's ok because the enum has no more extra cases

matibat (approved):
LGTM

PR #665, opened 2022-08-30
tests: update tests to use ardrive_io

karlprieb:


test/blocs/upload_cubit_test.dart
```diff
@@ -33,6 +32,7 @@ void main() {
   const tDriveId = 'drive_id';
   const tRootFolderId = 'root-folder-id';
   const tRootFolderFileCount = 5;
+  final tDefaultDate = DateTime.now();
```
matibat, line 35: To have this non-deterministic value in a single place reduces entropy -thanks for that 😎- but we still have entropy until we start using an explicit value

test/test_utils/fake_data.dart
```diff
@@ -14,6 +14,6 @@ const testEntityName = 'test-name';
 const testEntitySize = 10;
 const testEntityDataContentType = 'test-data-mime-type';
 const testEntityDataTxId = 'FxgiSM3JOAsjnZF-Yzv2z7h-Io4K6GmCarhQ0qAUL6I';
-const testEntityId = Uuid().v4();
-const driveId = Uuid().v4();
-const rootFolderId = Uuid().v4();
+final testEntityId = const Uuid().v4();
+final driveId = const Uuid().v4();
+final rootFolderId = const Uuid().v4();
```
matibat, line 19: I've already seen these variables in another PR (don't remember which one); same comment regarding entropy: instead of generating that random value we could have a mock for each

matibat (commented):
Nit: determinism in tests

test/blocs/upload_cubit_test.dart
agsuy: I agree we want a specific date here.

test/test_utils/fake_data.dart
agsuy: A mock or a stub? Why do you think mocks would be better here?

Generating UUIDv4 out of the lib is something we do every time we create an ArFS entity.
If there is an issues with random generation is something we want to know.

karlprieb:
Closed in favor of https://github.com/ardriveapp/ardrive-web/pull/678

PR #668, opened 2022-08-30
fix: wallet not being recognized if not named in a certain way PE-2170

thiagocarvalhodev:
Remove the filter on Android

- If the file provided matches the allowed extension
  - returns the file
- if not
  - show an error message


karlprieb (approved):

lib/l10n/app_en.arb
```diff
@@ -1363,5 +1363,13 @@
   "yourUploadFailed": "Your upload failed. You have not been charged for the failed attempt.",
   "@yourUploadFailed": {
     "description": "Upload fail dialog description."
+  },
+  "invalidWalletFile": "Invalid Wallet File",
```
matibat, line 1367: Do we have a ticket for translating these?

matibat (dismissed):
LGTM

fedellen (approved):

PR #678, opened 2022-09-06
tests(blocks): fix upload and data export cubit tests

karlprieb:
Changes from #665

### DIFF
--- a/test/blocs/upload_cubit_test.dart
+++ b/test/blocs/upload_cubit_test.dart
@@ -1,9 +1,7 @@
-@Tags(['broken'])
 import 'dart:io';
 import 'dart:typed_data';
 
 import 'package:ardrive/blocs/profile/profile_cubit.dart';
-import 'package:ardrive/blocs/upload/enums/conflicting_files_actions.dart';
 import 'package:ardrive/blocs/upload/models/upload_file.dart';
 import 'package:ardrive/blocs/upload/models/upload_plan.dart';
 import 'package:ardrive/blocs/upload/upload_cubit.dart';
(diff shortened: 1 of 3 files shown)


test/test_utils/fake_data.dart
```diff
@@ -14,6 +12,6 @@ const testEntityName = 'test-name';
 const testEntitySize = 10;
 const testEntityDataContentType = 'test-data-mime-type';
 const testEntityDataTxId = 'FxgiSM3JOAsjnZF-Yzv2z7h-Io4K6GmCarhQ0qAUL6I';
-const testEntityId = Uuid().v4();
-const driveId = Uuid().v4();
-const rootFolderId = Uuid().v4();
+const testEntityId = 'c1dda6a2-0d74-4c35-9671-54e450a778fd';
```
agsuy, line 15: I do not think removing random `Uuid().v4()` provides us better value.

When to use random generation vs. when to use constants usually comes down to what I am testing and if I care about the function generating those values or not. (e.g. if we use that random function only in tests or its a function used by our codebase to perform certain tasks.)

In this case, we use `Uuid().v4()` in our upload plan, so we definitely care about that function hence we should keep it.

agsuy (changes requested):
Please keep `Uuid().v4()`

test/blocs/upload_cubit_test.dart
```diff
@@ -58,10 +55,10 @@ void main() {
     registerFallbackValue(Wallet());
     registerFallbackValue(FolderEntry(
         id: '',
-        dateCreated: DateTime.now(),
+        dateCreated: tDefaultDate,
```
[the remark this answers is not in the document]
karlprieb: dateCreated doesn't matter here, so we're changing to calculate it just once.

test/test_utils/fake_data.dart
karlprieb: agreed!
[bede10e](https://github.com/ardriveapp/ardrive-web/pull/678/commits/bede10e9a857626319c0f94ffbc898795c29ce74)

PR #681, opened 2022-09-09
PE-2237: Plus/New button for Main Screen [Mobile MVP]

matibat:
- Factors out the newButton component to be reused
- Re-uses the newButton component to implement the Plus button for mobile only

--- Releases ---
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/3tc8er7pid4k8
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/237lg1e7md73g


lib/components/new_button.dart
```diff
@@ -0,0 +1,409 @@
...
+    constraints: constraints,
+    offset: offset,
+    onSelected: (callback) => callback(context),
+    itemBuilder: (context) => menuItems,
+    rotable: isPlusButton,
+    child: button,
+  );
+}
+
+class PopupMenuButtonRotable<T> extends StatefulWidget {
```
matibat, line 48: This is a custom implementation of the PopupMenuButton Widget. It tracks _rotable in the widget, and _opened in the state; the build method now returns an AnimatedRotation Widget.

This was the only approach that lets us know when the PopupMenu is opened.

lib/components/new_button.dart
```diff
@@ -0,0 +1,409 @@
...
+  required ProfileState profileState,
+  required DriveDetailState driveDetailState,
+  // String? title,
+}) {
+  if (profileState.runtimeType == ProfileLoggedIn) {
+    final minimumWalletBalance = BigInt.from(10000000);
+    final profile = profileState as ProfileLoggedIn;
+    final hasMinBalance = profile.walletBalance >= minimumWalletBalance;
+    return [
+      // if (title != null)
```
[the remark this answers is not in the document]
matibat: A tittle for the popup was in the initial design, then we decided we don't want that. I'm gonna remove that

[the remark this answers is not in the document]
matibat: b2dcdbd65bc657c2182fac0f83834af179649bae

lib/components/new_button.dart
```diff
@@ -0,0 +1,379 @@
...
+  BuildContext context, {
+  required Widget button,
+  required DrivesState drivesState,
+  required ProfileState profileState,
+  required DriveDetailState driveDetailState,
+  // String? title,
+  bool isPlusButton = false,
+}) {
+  final width = MediaQuery.of(context).size.width;
+  // TODO: double check if it's OK to use this context, if not then build twice
```
matibat, line 20: **For the reviewer**
I build the items to calculate the total height of the popup so I can calculate the offset of it.
Do you see any counterpart of using this context instead of the one at itemBuilder of PopupMenuButton? It doesn't seem to affect the functionality.

karlprieb: It's fine

lib/components/new_button.dart
karlprieb: Can't we just extend PopupMenuButton instead of recreating it here?

matibat: That was my initial approach

**TLDR** If I extend the Widget class I won't be able to use a custom state because createState returns a PopupMenuButtonState, not letting me to return a custom PopupMenuButtonRotableState instead

lib/components/new_button.dart
matibat:
```suggestion
```

lib/components/new_button.dart
thiagocarvalhodev: @matibat , we can implement it with a simpler implementation. I tested the feature and what I think better fits with your `newButton` is a FloatingActionButton (the button on the bottom right side). Also, we don't need the PopupMenuButton to show a modal. We can do it using the [showMenu function](https://api.flutter.dev/flutter/material/showMenu.html). 

We can achieve almost the same result with that implementation:

```dart
FloatingActionButton(
        onPressed: () {
          rotate();
          showMenu(
              context: context,
              position: RelativeRect.fromLTRB(
                double.maxFinite,
                double.maxFinite,
                60,
                140,
              ),
              items: [
                PopupMenuItem(
                  value: (context) {},
                  child: ListTile(
                    title: Text('some action'),
                  ),
                ),
              ]);
        },
        tooltip: 'Increment',
        child: AnimatedRotation(
          turns: turns,
          duration: const Duration(milliseconds: 400),
          curve: Curves.elasticIn,
          child: Icon(Icons.add),
        ),
      ),
   ```
where `rotate()` was a function I implemented just to rotate the icon.

thiagocarvalhodev: https://user-images.githubusercontent.com/32248947/191033839-38b6ed6b-be40-4b29-b737-8054edeb050c.mov

matibat: Tech debt ticket
- 15b7ef76cad6aa83bf8eeec94458bfb0e9b9abfb
- https://ardrive.atlassian.net/browse/PE-2314

karlprieb (approved):

karlprieb (approved):

PR #683, opened 2022-09-13
fix: fix async size and conflict handling PE-2258

javdhu:


lib/blocs/upload/upload_cubit.dart
```diff
@@ -347,12 +349,13 @@ class UploadCubit extends Cubit<UploadState> {
...
   }
 
-  void removeBigFiles() async {
+  Future<void> skipLargeFiles() async {
     for (final file in files) {
       if (await file.ioFile.length > sizeLimit) {
         files.remove(file);
       }
     }
+    checkConflicts();
```
fedellen, line 358: Is it okay to not `await` this?

thiagocarvalhodev (dismissed):
LGTM!

lib/blocs/upload/upload_cubit.dart
fedellen: Also a naming nit here, `skipLargeFiles` name does not really indicate that we'd also be checking conflicts

fedellen (dismissed):
Left some nits. I suspect the non await will be okay because we are in an `onPressed` event

lib/blocs/upload/upload_cubit.dart
karlprieb: If it's the last thing that your function will do it's fine to omit `await` keyword.

karlprieb (dismissed):

matibat (approved):

matibat (approved):
LGTM

thiagocarvalhodev (approved):
LGTM!

PR #685, opened 2022-09-13
PE-1863/1868 I/O: integration branch for I/O library's implementations

thiagocarvalhodev:


karlprieb (approved):

lib/components/upload_form.dart
```diff
@@ -31,7 +32,7 @@ Future<void> promptToUpload(
         .toList();
     selectedFiles.addAll(uploadFiles);
   } else {
-    final ioFiles = await io.pickFiles();
+    final ioFiles = await showMultipleFilesFilePickerModal(context);
```
thiagocarvalhodev, line 35: We need to validate here if we are on `Android` or `iOS`

karlprieb (approved):

PR #689, opened 2022-09-15
PE-2264: fix(top icons): fix top icon glitch outside the screen on small devices

thiagocarvalhodev:
- Added a horizontal scroll to top bar icons
- Set the status bar font to black


thiagocarvalhodev:
Some files were committed but I'll remove.

karlprieb (approved):

PR #693, opened 2022-09-20
PE-2322: show file source options just for mobile

karlprieb:
When you click on upload file button it will show file picker on web and display multiple options (gallery, file picker and camera) on mobile

--- Releases ---
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/71sc6r63e01rg


thiagocarvalhodev (approved):
Thanks for that!

PR #699, opened 2022-09-23
PE-2324: open url on external browsers by default

karlprieb:
Create `openUrl` abstraction that will open urls in external browsers by default.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/328l5pjqj1ol8
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/7eecbouqmtmlg


lib/app_shell.dart
```diff
@@ -43,11 +43,9 @@ class AppShellState extends State<AppShell> {
...
                     icon: const Icon(Icons.people_alt),
                     tooltip: 'CommunityXYZ',
-                    onPressed: () => launchUrl(
-                      Uri.parse(
-                        'https://community.xyz/#-8A6RexFkpfWwuyVO98wzSFZh0d6VJuI-buTJvlwOJQ',
-                      ),
-                    ),
+                    onPressed: () => openUrl(
+                        url:
+                            'https://community.xyz/#-8A6RexFkpfWwuyVO98wzSFZh0d6VJuI-buTJvlwOJQ'),
```
thiagocarvalhodev, line 48: Would be great to have all of those URLs in the same place. I think we have a ticket for that.

thiagocarvalhodev (approved):
LGTM!

PR #701, opened 2022-09-27
PE-2345: Only allow multiselect on writable drives

javdhu:


lib/pages/drive_detail/drive_detail_page.dart
```diff
@@ -59,13 +59,16 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
...
+                  if (keyListenerState is KeyboardListenerCtrlMetaPressed) {
+                    // Only allow multiselect on user drives and only if logged in
+                    if (state.hasWritePermissions) {
+                      checkboxEnabled = keyListenerState.isPressed;
+                      context
+                          .read<DriveDetailCubit>()
+                          .setMultiSelect(checkboxEnabled);
+                      setState(() => {});
+                    }
                   }
```
karlprieb, line 72:
```suggestion
                  // Only allow multiselect on user drives and only if logged in
                  if (keyListenerState is KeyboardListenerCtrlMetaPressed && state.hasWritePermissions) {
                    checkboxEnabled = keyListenerState.isPressed;
                    context
                        .read<DriveDetailCubit>()
                        .setMultiSelect(checkboxEnabled);
                    setState(() => {});
                  }
```

karlprieb: Let's create less blocks if we can :D

karlprieb (approved):

PR #708, opened 2022-09-30
PE-2373: fix mobile view item selection

javdhu:
Also handles an exception that happens only on mobile.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/3fm5usvdflqsg
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/65gmtfnajkhk8

### DIFF
--- a/lib/pages/drive_detail/components/drive_detail_data_list.dart
+++ b/lib/pages/drive_detail/components/drive_detail_data_list.dart
@@ -7,12 +7,10 @@ Widget _buildDataList(BuildContext context, DriveDetailLoadSuccess state) =>
           (folder) => _buildFolderListTile(
             context: context,
             folder: folder,
-            selected: state.selectedItems.isNotEmpty &&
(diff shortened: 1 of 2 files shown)


lib/components/keyboard_handler.dart
```diff
@@ -63,6 +51,24 @@ class _KeyboardHandlerState extends State<KeyboardHandler> {
   }
 }
 
+Future<bool> isCtrlOrMetaKeyPressed(RawKeyEvent event) async {
+  try {
+    final userAgent = (await DeviceInfoPlugin().webBrowserInfo).userAgent;
```
karlprieb, line 56: this will throw and exception on mobile

karlprieb (approved):

PR #722, opened 2022-10-04
PE-2365: Conditional platform tag

matibat:
Suggested changes to make the `App-Platform` tag to be conditional when the platform cannot be determined.

--- Releases ---
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/32gl2b62q2400

### DIFF
--- a/lib/utils/app_platform.dart
+++ b/lib/utils/app_platform.dart
@@ -1,7 +1,7 @@
 import 'package:flutter/foundation.dart';
 import 'package:platform/platform.dart';
 
-String getPlatform({Platform platform = const LocalPlatform()}) {
+String? getPlatform({Platform platform = const LocalPlatform()}) {
   if (kIsWeb) {
     return 'Web';
   }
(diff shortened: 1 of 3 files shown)


lib/utils/app_platform.dart
```diff
@@ -15,6 +15,6 @@ String getPlatform({Platform platform = const LocalPlatform()}) {
     case 'ios':
       return 'iOS';
     default:
-      throw Exception('Unsupported platform $operatingSystem!');
+      return null;
```
matibat, line 18: If we decide to put the tag anyways, then the only required change is to make this method to return `"unknown"` instead of `null`.

matibat: I'd say that's what we want: it's not possible to make a GQL query for TXs with certain tag being absent; in other words: you cannot query for items that don't have certain tag.

matibat:
Won't apply this change.

PR #728, opened 2022-10-06
PE-2401: Adds unit tests - App-Platform

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5jberrsvvl59g
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/7sbsk4odvd5v0


test/models/entity_version_tag_test.dart
```diff
@@ -1,16 +1,19 @@
-@Tags(['broken'])
```
matibat, line 1: These tests were working before - I don't know why did we disable these.

I've re-enabled these and added tests for the App-Platform tag

lib/entities/entity.dart
```diff
@@ -91,9 +91,16 @@ extension TransactionUtils on TransactionBase {
     required String version,
     DateTime? unixTime,
     Platform platform = const LocalPlatform(),
+    bool isWeb = kIsWeb,
```
karlprieb, line 94: Just use `kIsWeb` we don't need to create this new pointer.

test/utils/app_platform_test.dart
```diff
@@ -0,0 +1,36 @@
...
+  group('getPlatform method', () {
+    test('returns "Web" for browser platforms', () {
+      final platformString =
+          getPlatform(platform: unknownFakePlatform, isWeb: true);
+      expect(platformString, 'Web');
+    });
+
+    test('returns "Android" for Android devices', () {
+      final platformString =
+          getPlatform(platform: androidFakePlatform, isWeb: false);
```
karlprieb, line 20: Can we have one of these getPlatform calls without passing isWeb? Just to make sure that this method is returning the right output even when we don't send this optional paramenter.

matibat:
Cost estimate tests [WIP] - #729

test/utils/app_platform_test.dart
matibat: The optional parameter defaults to `kIsWeb` which I don't know how to mock (or if it's possible to mock). It should be false on the context of the tests (unless we start running them on web). I'd prefer not to test the default case because that behavior would be dynamic

matibat: The reason why I'm injecting this value is to not rely on the dynamic value of kIsWeb

karlprieb (approved):

lib/utils/app_platform.dart
```diff
@@ -1,8 +1,11 @@
 import 'package:flutter/foundation.dart';
 import 'package:platform/platform.dart';
 
-String getPlatform({Platform platform = const LocalPlatform()}) {
-  if (kIsWeb) {
+String getPlatform({
```
thiagocarvalhodev, line 4: Maybe would be better if we have an enum to represent the supported platforms.

Similar to the `Platform` object from `dart:io`, but one that abstracts the logic to get the platform.

lib/utils/app_platform.dart
```diff
@@ -1,8 +1,11 @@
...
+  Platform platform = const LocalPlatform(),
+  isWeb = kIsWeb,
```
thiagocarvalhodev, line 6: Can we type `isWeb`?

thiagocarvalhodev (dismissed):
Some comments, but it's good to go!

lib/utils/app_platform.dart
```diff
@@ -18,3 +21,16 @@ String getPlatform({Platform platform = const LocalPlatform()}) {
       return 'unknown';
   }
 }
+
+class SystemPlatform {
```
matibat, line 25: TODO: add a unit test for this class

matibat (commented):
Self review checkpoint

lib/utils/app_platform.dart
matibat: 424682e8aebae955167010437618ff5250098d43

lib/utils/app_platform.dart
matibat: There it is - 29f4f2e2ad9a41359298184a3b56f5b6f8e564ad

lib/utils/app_platform.dart
matibat: I've just tried it, but it just adds more complexity at the moment of transforming the enum value into a String: 
1. The values of the enum has to be in camelCase (that is to begin with lower case)
2. Not all of the values follow the same case standard. E.g. "Web" (SnakeCase) vs "Android" (SnakeCase) vs "iOS" (camelCase) vs "unknown" (camelCase).
So to work around that I should add another switch..case to return the correct string.

It was a nice suggestion, but hard to apply.

PR #742, opened 2022-10-12
PE-2410: folder and multiple files and cannot be uploaded using ar connect

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/1ns91i56h6vf8
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/6j7jelopf2250

### DIFF
--- a/lib/blocs/upload/upload_handles/bundle_upload_handle.dart
+++ b/lib/blocs/upload/upload_handles/bundle_upload_handle.dart
@@ -51,6 +51,7 @@ class BundleUploadHandle implements UploadHandle {
     required Wallet wallet,
   }) async {
     final bundle = await DataBundle.fromHandles(
+      parallelize: false,
       handles: List.castFrom<FileDataItemUploadHandle, DataItemHandle>(
               fileDataItemUploadHandles) +
           List.castFrom<FolderDataItemUploadHandle, DataItemHandle>(
--- a/lib/blocs/upload/upload_handles/bundle_upload_handle.dart
+++ b/lib/blocs/upload/upload_handles/bundle_upload_handle.dart
@@ -49,9 +49,10 @@ class BundleUploadHandle implements UploadHandle {
     required DriveDao driveDao,
     required PstService pstService,
     required Wallet wallet,
+    bool isArConnect = false,
   }) async {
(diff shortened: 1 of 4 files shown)


pubspec.yaml
```diff
@@ -35,7 +35,7 @@ dependencies:
   arweave:
     git:
       url: https://github.com/ardriveapp/arweave-dart
-      ref: v3.3.0
+      ref: PE-2410-folder-and-multiple-files-and-cannot-be-uploaded-using-ar-connect
```
matibat, line 38: Reminder to update this when the Arweave-dart gets released

lib/blocs/upload/upload_handles/bundle_upload_handle.dart
```diff
@@ -50,21 +51,37 @@ class BundleUploadHandle implements UploadHandle {
     required Wallet wallet,
   }) async {
     final bundle = await DataBundle.fromHandles(
+      parallelize: false,
```
matibat, line 54: I'm not sure about this line. I'll see if I find out when reviewing Arweave-Dart

matibat (commented):
Looks good -

lib/blocs/upload/upload_handles/bundle_upload_handle.dart
matibat: If running this method synchronously is the fix and this affects ArConnect only, then we could do something like...
```suggestion
      parallelize: !isArConnect,
```

thiagocarvalhodev: That makes total sense. Thanks for this good catch!

thiagocarvalhodev: [fc40ea6](https://github.com/ardriveapp/ardrive-web/pull/742/commits/fc40ea60ee52bb7cccabbfd548f10234e43ae643)

karlprieb (approved):

karlprieb (approved):

PR #762, opened 2022-11-01
PE-2362: Add copy wallet address button

jdaev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5ffhgikbj9818
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/0hnh0qu2friko

### DIFF
--- a/lib/components/copy_icon_button.dart
+++ b/lib/components/copy_icon_button.dart
@@ -1,0 +1,34 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+class CopyIconButton extends StatelessWidget {
+  final String value;
+  final String tooltip;
+  final double? size;
+  final Function? onTap;
+  const CopyIconButton({
+    Key? key,
+    required this.value,
+    required this.tooltip,
+    this.size,
+    this.onTap,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return IconButton(
+      icon: Icon(
+        Icons.copy,
+        color: Colors.black54,
+        size: size,
+      ),
+      tooltip: tooltip,
+      onPressed: () {
+        Clipboard.setData(ClipboardData(text: value));
+        if (onTap != null) {
+          onTap!();
+        }
+      },
+    );
+  }
+}
--- a/lib/components/copy_icon_button.dart
+++ b/lib/components/copy_icon_button.dart
@@ -25,9 +25,7 @@ class CopyIconButton extends StatelessWidget {
       tooltip: tooltip,
       onPressed: () {
         Clipboard.setData(ClipboardData(text: value));
-        if (onTap != null) {
-          onTap!();
-        }
+        onTap?.call();
       },
     );
   }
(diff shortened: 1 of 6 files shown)


lib/components/profile_overlay.dart
```diff
@@ -54,9 +70,34 @@ class ProfileOverlay extends StatelessWidget {
...
+                              baseline: TextBaseline.alphabetic,
+                              child: CopyIconButton(
+                                value: state.walletAddress,
+                                tooltip: appLocalizationsOf(context)
+                                    .copyWalletAddress,
+                                size: 12,
+                                onTap: () {
+                                  showCopyToast();
+                                },
+                              ),
```
matibat, line 89: Nit: When you hover the button the splash-radius (idk if that's the actual name of the property, but I mean the area that appears enclosing the clickable area of the button) covers the text also

lib/l10n/app_en.arb
```diff
@@ -1481,5 +1481,14 @@
   "forgetWalletDescription": "Are you sure you want to forget your wallet? You will need to select another wallet to login with another profile.",
   "@forgetWalletDescription": {
     "description": "Modal description to forget wallet"
+  },
+  "addressCopied": "Address Copied!",
```
matibat, line 1485: Let's let the PO know about these if we haven't already

lib/l10n/app_en.arb
```diff
@@ -1481,5 +1481,14 @@
...
+  "@addressCopied": {
+    "description": "Toast message when you copy wallet address"
+  }
+  ,
```
matibat, line 1489: Nit
```suggestion
  },
```

matibat (dismissed):
Approved with nit comments

https://user-images.githubusercontent.com/22564541/199556571-95b9eaca-593a-42a3-a6c0-ef1754823a32.mov

lib/components/copy_icon_button.dart
```diff
@@ -0,0 +1,34 @@
...
+        Icons.copy,
+        color: Colors.black54,
+        size: size,
+      ),
+      tooltip: tooltip,
+      onPressed: () {
+        Clipboard.setData(ClipboardData(text: value));
+        if (onTap != null) {
+          onTap!();
+        }
```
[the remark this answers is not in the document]
thiagocarvalhodev: I agree with mati, but the correct way is: `onTap?.call()`.

lib/l10n/app_en.arb
thiagocarvalhodev: @matibat, I checked the ticket and these texts match.

thiagocarvalhodev (dismissed):
LGTM! Just one observation: on mobile the text "Address Copied!" is under the wallet address, for me, it's looking good! I would just change it to use a SnackBar but I'll let the suggestion on the P&D group.



https://user-images.githubusercontent.com/32248947/199574980-dfbe79cf-90eb-487f-8c0b-006352e78059.mov

karlprieb (approved):

thiagocarvalhodev (approved):

PR #764, opened 2022-11-02
PE-2309: fix: previous versions cannot be downloaded in sharing landing page mode

thiagocarvalhodev:
If we access a shared file from a private drive and paste it in a new tab, it is not possible to download any revision of the file from the 'Details' and 'Activity' panel.

Now we are using the `FileRevision` to download the file.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/4b2dr4krpceh8
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/5a2i81bjemppg

### DIFF
--- a/lib/blocs/file_download/file_download_cubit.dart
+++ b/lib/blocs/file_download/file_download_cubit.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:developer';
 import 'dart:io';
+import 'dart:typed_data';
 
 import 'package:ardrive/blocs/blocs.dart';
 import 'package:ardrive/entities/entities.dart';
--- a/lib/blocs/file_download/file_download_cubit.dart
+++ b/lib/blocs/file_download/file_download_cubit.dart
@@ -1,6 +1,6 @@
 import 'dart:async';
+import 'dart:developer';
 import 'dart:io';
-import 'dart:typed_data';
 
 import 'package:ardrive/blocs/blocs.dart';
 import 'package:ardrive/entities/entities.dart';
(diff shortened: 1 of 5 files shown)


lib/pages/profile_auth/components/profile_auth_prompt_wallet_screen.dart
```diff
@@ -64,6 +65,7 @@ class ProfileAuthPromptWalletScreen extends StatelessWidget {
                 textAlign: TextAlign.center,
               ),
             ),
+            // const Spacer(),
```
matibat, line 68: Nit: unnecessary comment line

matibat (approved):
The code LGTM - I haven't yet manually tested the feature

matibat:
I tested on Safari. It does work 💯

lib/blocs/file_download/file_download_cubit.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:developer';
```
karlprieb, line 2: Why do we need this package?

karlprieb (approved):

lib/blocs/file_download/file_download_cubit.dart
thiagocarvalhodev: Oh, that's to use the `log` function to avoid using prints

PR #792, opened 2022-11-17
PE-2565: implement the radio button component

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/7km6mbol4j2bo
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/7t0djcuhagj9o


packages/ardrive_ui_library/lib/src/components/radio_button.dart
```diff
@@ -0,0 +1,228 @@
...
+
+    super.initState();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return ListView.builder(
+      itemBuilder: (context, i) {
+        return Align(
+          alignment: Alignment.center,
```
karlprieb, line 56: Shouldn't it be left aligned? So you don't lose alignment when you have group of radio buttons?

image

packages/ardrive_ui_library/lib/src/components/radio_button.dart
```diff
@@ -0,0 +1,228 @@
...
+    if (!widget.isEnabled) {
+      state = RadioButtonState.disabled;
+      _value = false;
+    } else if (widget.value) {
+      state = RadioButtonState.checked;
+      _value = true;
+    } else {
+      state = RadioButtonState.unchecked;
+      _value = false;
+    }
```
karlprieb, line 122: As we use this same logic twice (here in the `initState` and on `didUpdateWidget`) can we abstract that?

packages/ardrive_ui_library/test/src/components/radio_button_test.dart
```diff
@@ -0,0 +1,191 @@
...
+      /// Taps the radio button
+      await tester.tap(radioButtonClick);
+
+      final state = tester.state<ArDriveRadioButtonState>(radioButton);
+
+      /// should be enabled
+      expect(state.state, RadioButtonState.checked);
+    });
+
+    // testWidgets('Should have only one the RadioButton checked at the time',
```
karlprieb, line 138: Can you remove this comment?

packages/ardrive_ui_library/lib/src/components/radio_button.dart
thiagocarvalhodev: In this case, we are centering the list of RadioButtons in the center of the list. `ListView` always expands the widgets to fill all the width when not aligned.
Aligning to the center we can let the parent widget decide where to align them on a page.

thiagocarvalhodev: Yeah, that makes sense. [51b546d](https://github.com/ardriveapp/ardrive-web/pull/792/commits/51b546d9056a819fdb8f6c9037a1fe44b487315e) on this commit I'm giving to the parent component the option to align where needed. You can test it with the `Alignment` knob on Storybook.

packages/ardrive_ui_library/test/src/components/radio_button_test.dart
thiagocarvalhodev: [5b7b96e](https://github.com/ardriveapp/ardrive-web/pull/792/commits/5b7b96e04d9e8938e98e665d83c8d24b957491ba)

packages/ardrive_ui_library/lib/src/components/radio_button.dart
thiagocarvalhodev: [51b546d](https://github.com/ardriveapp/ardrive-web/pull/792/commits/51b546d9056a819fdb8f6c9037a1fe44b487315e)

karlprieb:
One last thing that I think we missed is the behavior of a group. Currently, is possible to deselect a radio button and let the radio button group without any value.
That's not a behavior of a radio button group. The user should select at least one.

Basically, what we want is to have an option to have a default button selected or not (all buttons deselected) and also the user can only select a not selected button. If the user clicks on an already selected button nothing happens.

thiagocarvalhodev:
@karlprieb , thanks for this point, makes total sense. Here is the fix: [77c8bb2](https://github.com/ardriveapp/ardrive-web/pull/792/commits/77c8bb256c76601eaca8089aabf0d427c09e6d26)

karlprieb (approved):

PR #808, opened 2022-11-30
PE-2576: Add TabView

javdhu:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5v0ui48plqlso
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/5viu967cohe5g

### DIFF
--- a/packages/ardrive_ui_library/lib/src/components/tab_view.dart
+++ b/packages/ardrive_ui_library/lib/src/components/tab_view.dart
@@ -1,11 +1,11 @@
 import 'package:ardrive_ui_library/ardrive_ui_library.dart';
 import 'package:flutter/material.dart';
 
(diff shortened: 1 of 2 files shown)


karlprieb:
I saw that's possible to change tabs when doing a horizontal scrolling. If that is intended, shouldn't we allow people to also change tabs when swiping?

Another thing is how to use it. We need to send two lists (tabs and widgets) and ensure they have the same length and order. Wouldn't be easy to receive a list of maps? So we can keep tab and content together. What do you think?

karlprieb (approved):

PR #811, opened 2022-11-30
feat(localization fas): adds the Farsi language PE-2727

matibat:
Changes
- Adds Farsi to the `localizely.yml` file
- Adds the ARB file


lib/main.dart
```diff
@@ -189,6 +189,7 @@ class AppState extends State<App> {
                   countryCode: 'HK',
                 ), // generic traditional Chinese 'zh_Hant'
                 Locale('ja', ''), // Japanese, no country code
+                Locale('fas', ''), // Farsi, no country code
```
matibat, line 192:
```suggestion
                Locale('fa', ''), // Farsi, no country code
```

PR #841, opened 2023-01-05
[Draft] Support for large transactions in public folders (web only)

elliotsayes:
Making this to try and activate the PR workflow, feel free to ignore

### DIFF
--- a/lib/blocs/upload/limits.dart
+++ b/lib/blocs/upload/limits.dart
@@ -1,3 +1,3 @@
 const privateFileSizeLimit = 104857600;
-const publicFileSizeLimit = 1288490189;
-const mobileFileSizeLimit = 1073741823;
+const publicFileSizeLimit = 1288490189 * 100; // TODO: Decide a sane limit for stream transctions
+const mobilePrivateFileSizeLimit = 1073741823;
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -407,9 +407,9 @@ class UploadCubit extends Cubit<UploadState> {
     );
   }
 
-  int get sizeLimit => kIsWeb
-      ? (_targetDrive.isPrivate ? privateFileSizeLimit : publicFileSizeLimit)
-      : mobileFileSizeLimit;
+  int get sizeLimit => _targetDrive.isPrivate 
+      ? (kIsWeb ? privateFileSizeLimit: mobilePrivateFileSizeLimit)
+      : publicFileSizeLimit;
 
   void _removeFilesWithFolderNameConflicts() {
     files.removeWhere((file) => conflictingFolders.contains(file.ioFile.name));
--- a/lib/blocs/upload/upload_handles/file_v2_upload_handle.dart
+++ b/lib/blocs/upload/upload_handles/file_v2_upload_handle.dart
@@ -57,13 +57,24 @@ class FileV2UploadHandle implements UploadHandle {
     required Wallet wallet,
     required PstService pstService,
   }) async {
-    final fileData = await file.ioFile.readAsBytes();
     final packageInfo = await PackageInfo.fromPlatform();
     final String version = packageInfo.version;
+
+    Transaction transaction;
+    if (isPrivate) {
+      transaction = await createEncryptedTransaction(
+        await file.ioFile.readAsBytes(),
+        fileKey!,
+      );
+    } else {
+      transaction = TransactionStream.withBlobData(
+        dataStreamGenerator: file.ioFile.openReadStream,
+        dataSize: await file.ioFile.length,
+      );
+    }
+
     dataTx = await arweaveService.client.transactions.prepare(
-      isPrivate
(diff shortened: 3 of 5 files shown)


elliotsayes:
Known issues
- When Drag&Drop a folder, instead of showing an error modal, it treats the folder as a file
  - The previous method of detecting folders stalls with large files, and I haven't found another way of detecting folders. Would probably require a fork
- Preparing upload for large-ish files (~2GB) takes a long time on Android (Samsung S8 physical device, development apk). Not really sure what it is doing here

karlprieb:
> When Drag&Drop a folder, instead of showing an error modal, it treats the folder as a zip file with incorrect length...
> The previous method of detecting folders stalls with large files, and I haven't found another way of detecting folders. Would probably require a fork of the Drag&Drop package
For now we can keep the current behavior

> Preparing upload for large-ish files (~2GB) takes a long time on Android (Samsung S8 physical device, development apk). Not really sure what it is doing here
That is probably happening because `flutter_file_picker` cache the files first. Basically, it reads the file stream to create a new file inside the app, after that this cached file is available to use. In this case the file stream used by your code is from the cached file. From what I remember this is used when the user selects a file from a private folder (Downloads for example). If you don't cache you lose the reference for that file.
In our case we could get the file stream directly from the original file bypassing the cache step.

elliotsayes:
> For now we can keep the current behavior

👌

> That is probably happening because `flutter_file_picker` cache the files first. Basically, it reads the file stream to create a new file inside the app, after that this cached file is available to use. In this case the file stream used by your code is from the cached file.

In other words it first copies the whole file from the user directory to the app's directory, and then uses that instead?

> From what I remember this is used when the user selects a file from a private folder (Downloads for example). If you don't cache you lose the reference for that file.

Found in the `flutter_file_picker` changelog to support your claim:
Screenshot 2023-01-18 at 11 13 41 AM

> In our case we could get the file stream directly from the original file bypassing the cache step.

Is your suggestion that by creating a Stream *immediately*, then it won't matter if the reference to the original file is lost? This is possible but would be very messy, as you'd have to grab the number of streams you're planning to use in advance (2 in our case)

karlprieb:
> For now we can keep the current behavior

I mean the behavior that's currently on `dev`.

> In other words it first copies the whole file from the user directory to the app's directory, and then uses that instead?

Yep :(

> Is your suggestion that by creating a Stream immediately, then it won't matter if the reference to the original file is lost?

Yes, that's it. I can't think on a clean way to implement this right now, but it's something that we will need to work on in the future.

> This is possible but would be very messy, as you'd have to grab the number of streams you're planning to use in advance (2 in our case)

Exactly!

elliotsayes:
> I mean the behavior that's currently on `dev`.

The current behaviour relies on `readAsBytes` so this would effectively negate the point of the PR (for Drag&Drop at least)
https://github.com/ardriveapp/ardrive-web/pull/841/files#diff-63710b60e8dbf717ede1b79e436c1138b5d17ab6e929e9437b66771f9818ee5bL81

For a quick hack, I could try and detect attempted folder uploads based on exact file size + file extension that is reported by `flutter_dropzone`? Will have a tiny false positive rate, but in those cases it can just point the user to use the upload button.

> Yes, that's it. I can't think on a clean way to implement this right now, but it's something that we will need to work on in the future.

Fair enough. I'll do a bit of research myself to see if there's a cleaner way, otherwise I'll implement your suggestion.

elliotsayes:
Update: it seems like the maintaner of `flutter_file_picker` reverted the behaviour of skipping caching after regressions. I'm hesitant to try this now... will look into alternatives

Original PR: https://github.com/miguelpruivo/flutter_file_picker/pull/1123
Issue: https://github.com/miguelpruivo/flutter_file_picker/issues/1133 

Screenshot 2023-01-20 at 3 05 26 PM

elliotsayes:
**Regarding Drag&Drop folders**

The only way to detect these reliably was by forking `flutter_dropzone`. See the fork here: https://github.com/deakjahn/flutter_dropzone/compare/master...elliotsayes:flutter_dropzone:folder-detect

Let me know if this is acceptable, and I will merge the required changes to `ardrive-web`: https://github.com/elliotsayes/ardrive-web/compare/stream-tx...elliotsayes:ardrive-web:stream-tx-folder-hack

elliotsayes:
**Regarding Android file caching**

@ abhi16180 is maintaining a fork of `flutter_file_picker` with cache skipping, over at https://github.com/abhi16180/flutter_file_picker
I've made a test branch of `flutter_file_picker` based on this fork (along with my previous changes) here: https://github.com/miguelpruivo/flutter_file_picker/compare/master...elliotsayes:flutter_file_picker:skip-cache-power-stream

Test branch of `ardrive-web` using this fork with required changes: https://github.com/elliotsayes/ardrive-web/compare/stream-tx...elliotsayes:ardrive-web:stream-tx-open-queue

A cursory test on my Android device proves that it does indeed skip the prolonged "Preparing..." screen. I didn't test a full large file upload however (both my device and internet are very slow atm!)

PR #847, opened 2023-01-12
PE-2824: creates a table for SnapshotEntry

matibat:
Changes
- Declares a new table for Snapshot Entries: It tracks:
  - The snapshot id,
  - the drive id,
  - the block range and
  - the data range, and
  - the unix time.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/3kp3nbad076o8
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/2k58hof0rt4h0

### DIFF
--- a/lib/models/tables/snapshot_entries.drift
+++ b/lib/models/tables/snapshot_entries.drift
@@ -6,6 +6,8 @@ CREATE TABLE snapshot_entries (
     dataStart INTEGER NOT NULL,
     dataEnd INTEGER NOT NULL,
 
+    txId TEXT NOT NULL,
+
     dateCreated DATETIME NOT NULL DEFAULT (strftime('%s','now')),
     PRIMARY KEY (id, driveId)
 ) As SnapshotEntry;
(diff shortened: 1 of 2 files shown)


matibat:
This change implies a DB migration. Is something else I'm missing here for that?

thiagocarvalhodev (dismissed):
LGTM

karlprieb (dismissed):

lib/models/tables/snapshot_entries.drift
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE snapshot_entries (
+    id TEXT NOT NULL,
+    driveId TEXT NOT NULL,
+    blockStart INTEGER NOT NULL,
+    blockEnd INTEGER NOT NULL,
+    dataStart INTEGER NOT NULL,
+    dataEnd INTEGER NOT NULL,
```
matibat, line 7: There won't be revisions for Snapshots; revisions had the responsibility of tracking the metadataTxId. I would start tracking that field in this table as the relationship is 1:1 (1 snapshot id to 1 tx id).
```suggestion
    dataEnd INTEGER NOT NULL,

    txId TEXT NOT NULL,
```

thiagocarvalhodev (approved):

PR #848, opened 2023-01-12
PE-2824: feat(snapshot entity): declares SnapshotEntity

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/52rl6dn1458j0
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/3rv1tqj0m2sqg

### DIFF
--- a/lib/entities/snapshot_entity.dart
+++ b/lib/entities/snapshot_entity.dart
@@ -68,6 +68,4 @@ class SnapshotEntity extends Entity {
       ..addTag(EntityTag.dataStart, dataStart!)
       ..addTag(EntityTag.dataEnd, dataEnd!);
   }
-
-  Map<String, dynamic> toJson() => _$SnapshotEntityToJson(this);
 }
--- a/lib/entities/snapshot_entity.dart
+++ b/lib/entities/snapshot_entity.dart
@@ -1,0 +1,82 @@
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:ardrive/services/services.dart';
+import 'package:arweave/arweave.dart';
+import 'package:json_annotation/json_annotation.dart';
+
+import 'entities.dart';
+
+part 'snapshot_entity.g.dart';
+
+@JsonSerializable()
+class SnapshotEntity extends Entity {
+  @JsonKey(ignore: true)
+  String? id;
+  @JsonKey(ignore: true)
+  String? driveId;
+  @JsonKey(ignore: true)
+  String? blockStart;
+  @JsonKey(ignore: true)
+  String? blockEnd;
+  @JsonKey(ignore: true)
+  String? dataStart;
+  @JsonKey(ignore: true)
+  String? dataEnd;
+
+  SnapshotEntity({
+    this.id,
+    this.driveId,
+    this.blockStart,
+    this.blockEnd,
+    this.dataStart,
+    this.dataEnd,
+  });
+
+  static Future<SnapshotEntity> fromTransaction(
+    TransactionCommonMixin transaction,
+    Uint8List data,
+  ) async {
+    try {
+      Map<String, dynamic>? entityJson = json.decode(utf8.decode(data));
+
+      return SnapshotEntity.fromJson(entityJson!)
+        ..id = transaction.getTag(EntityTag.folderId)
+        ..driveId = transaction.getTag(EntityTag.driveId)
+        ..blockStart = transaction.getTag(EntityTag.blockStart)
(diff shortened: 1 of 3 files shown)


lib/entities/snapshot_entity.dart
```diff
@@ -0,0 +1,82 @@
...
+      ..addTag(EntityTag.blockStart, blockStart!)
+      ..addTag(EntityTag.blockEnd, blockEnd!)
+      ..addTag(EntityTag.dataStart, dataStart!)
+      ..addTag(EntityTag.dataEnd, dataEnd!);
+  }
+
+  factory SnapshotEntity.fromJson(Map<String, dynamic> json) =>
+      _$SnapshotEntityFromJson(json);
+
+  Map<String, dynamic> toJson() => _$SnapshotEntityToJson(this);
```
matibat, line 81: Is this method supposed to return the entity data itself?

lib/entities/snapshot_entity.dart
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  static Future<SnapshotEntity> fromTransaction(
+    TransactionCommonMixin transaction,
+    Uint8List data,
+  ) async {
+    try {
+      Map<String, dynamic>? entityJson = json.decode(utf8.decode(data));
+
+      return SnapshotEntity.fromJson(entityJson!)
```
thiagocarvalhodev, line 43: Can we validate if this variable is not null?

lib/entities/snapshot_entity.dart
```diff
@@ -0,0 +1,82 @@
...
+    this.dataStart,
+    this.dataEnd,
+  });
+
+  static Future<SnapshotEntity> fromTransaction(
+    TransactionCommonMixin transaction,
+    Uint8List data,
+  ) async {
+    try {
+      Map<String, dynamic>? entityJson = json.decode(utf8.decode(data));
```
thiagocarvalhodev, line 41: consider making this variable final

```dart
final entityJson = json.decode(utf8.decode(data));
```

thiagocarvalhodev (commented):
Left minor comments.

lib/entities/snapshot_entity.dart
thiagocarvalhodev: It will return a map with the object's properties

lib/entities/snapshot_entity.dart
matibat: I will actually remove the `data` argument as it's not required for snapshots: if you see the `_$SnapshotEntityFromJson` -autogenerated- method, it doesn't pass the argument to the factory because we don't use any field from the data here (as we do with the name of Folders for example).

matibat: Here - 5d573c6af1200ade5d2d8583410caab530b7a4e0

lib/entities/snapshot_entity.dart
matibat: Removed the `data` argument, and `entityJson` variable. - 5d573c6af1200ade5d2d8583410caab530b7a4e0

lib/entities/snapshot_entity.dart
matibat: In the case of Drive:
```dart
Map<String, dynamic> _$DriveEntityToJson(DriveEntity instance) =>
    <String, dynamic>{
      'name': instance.name,
      'rootFolderId': instance.rootFolderId,
    };
```

So yes, that method is supposed to return the data of the Snapshot Entity, which this class won't know. I would simply remove the method or make it throw: "Unimplemented".

matibat: Removed unnecessary method - 72ce10601320d2f04e6bfc0c1f4425c2628cbf43

thiagocarvalhodev (approved):
LGTM

karlprieb (approved):

karlprieb (approved):

lib/entities/snapshot_entity.dart
```diff
@@ -0,0 +1,71 @@
...
+        ..createdAt = transaction.getCommitTime();
+    } catch (_) {
+      throw EntityTransactionParseException(transactionId: transaction.id);
+    }
+  }
+
+  @override
+  void addEntityTagsToTransaction<T extends TransactionBase>(T tx) {
+    assert(id != null &&
+        driveId != null &&
```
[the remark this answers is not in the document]
matibat: snapshotId is actually id.

PR #860, opened 2023-01-25
PE-2871: decode json metadata in snapshots as base-64 for private drives

matibat:
Changes:

- Makes the `createDriveEntityHistoryFromTransactions` method to base64-decode tx data of private entities.
- Makes that method to filter out unknown entities (instead of throwing when accessing an attribute of null).

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/129d6tn7b4c80
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/7uuk7hda7u3a0


lib/services/arweave/arweave_service.dart
```diff
@@ -260,7 +266,13 @@ class ArweaveService {
             rawEntityData,
             driveKey: driveKey,
           );
+        } else if (entityType == EntityType.snapshot) {
+          entity = await SnapshotEntity.fromTransaction(
+            transaction,
+          );
+          print('Snapshot entity found - ${transaction.toJson()}');
         }
```
matibat, line 274: This is a placeholder for now. It won't be written to the DB. Without this, it throws an "unexpected null" in the lines below.

test/utils/snapshot_item_test.dart
```diff
@@ -319,19 +319,84 @@ void main() {
...
+            );
+            // further calls to the method results in a null response
+            expect(
+              await SnapshotItemOnChain.getDataForTxId(
+                'DRIVE_ID',
+                'tx-$height',
+                true,
+              ),
+              null,
+            );
```
matibat, line 395: No need to test this again
```suggestion
```

matibat: c681ac96b4ae0055ad53f2d6492d6747c54b3665

matibat (commented):
-

thiagocarvalhodev (dismissed):
LGTM.

I just have some concerns about the `SnapshotItem` responsibilities. Also, looking into the code, we have some `catchError` that just prints the error, and the flow goes forward. Is that intended?

For example:    [ final dataBytes = await ArDriveHTTP().getAsBytes(_dataUri).catchError(] (https://github.com/ardriveapp/ardrive-web/blob/c681ac96b4ae0055ad53f2d6492d6747c54b3665/lib/utils/snapshots/snapshot_item.dart#L157)

and https://github.com/ardriveapp/ardrive-web/blob/c681ac96b4ae0055ad53f2d6492d6747c54b3665/lib/utils/snapshots/snapshot_item.dart#L193

lib/utils/snapshots/snapshot_item.dart
```diff
@@ -227,12 +227,26 @@ class SnapshotItemOnChain implements SnapshotItem {
...
 
-    return value;
+    if (value != null) {
+      if (isPrivate) {
+        // then it's base64-encoded
+        return base64.decode(String.fromCharCodes(value));
+      }
+
+      // if it's public then it's a non-base64 plain text
+      return value;
```
matibat, line 246: Maybe do this in a method that knows how to base64-decode, and make this method pure

lib/services/arweave/arweave_service.dart
matibat: Let's instead add a conditional check for null entities instead of instantiating a new object that won't be used

matibat: Applied - 40ac43b3e879ba4cfc9421e0bdb04e7d122352ee

This is safer even for corrupted drives with unknown entities; those are now gonna be ignored instead of failing in the sync stage.

lib/utils/snapshots/snapshot_item.dart
```diff
@@ -154,11 +154,7 @@ class SnapshotItemOnChain implements SnapshotItem {
       return _cachedSource!;
     }
 
-    final dataBytes = await ArDriveHTTP().getAsBytes(_dataUri).catchError(
-      (e) {
-        print('Error while fetching Snapshot Data - $e');
-      },
-    );
+    final dataBytes = await ArDriveHTTP().getAsBytes(_dataUri);
```
matibat, line 157: Answering @thiagocarvalhodev's comment:

I've remove the catch: we do want this to throw when the retries fails - 9e08a2c0d8602136d22d7d9038d2a00c127e0bb6

lib/utils/snapshots/snapshot_item.dart
```diff
@@ -191,8 +187,10 @@ class SnapshotItemOnChain implements SnapshotItem {
       try {
         node = DriveHistoryTransaction.fromJson(item['gqlNode']);
       } catch (e, s) {
-        print('Error while parsing GQLNode - $e, $s');
-        rethrow;
+        print(
+          'Error while parsing GQLNode from snapshot item ($txId) - $e, $s',
+        );
+        continue;
```
matibat, line 193: Answering to @thiagocarvalhodev:

Here we do not want to throw, but instead log the corrupted transaction and continue parsing the others.

lib/utils/snapshots/snapshot_item.dart
matibat: 5efc69550de2e9eebd636be3bf111606df9926e3

matibat: This class ended up with no relevant changes

lib/utils/snapshots/snapshot_item.dart
thiagocarvalhodev: Got it. Do we have a ticket that handles corrupted transactions? Do we want to handle it?

lib/services/arweave/arweave_service.dart
```diff
@@ -213,9 +213,20 @@ class ArweaveService {
         (entity) async {
           final txId = entity.id;
           final Uint8List? cachedData =
-              await SnapshotItemOnChain.getDataForTxId(driveId, txId);
+              await SnapshotItemOnChain.getDataForTxId(
```
thiagocarvalhodev, line 216: can we split this logic into another function? Something like:

```dart
  Future<Uint8List> _getEntityData({
    required String entityId,
    required String driveId,
    required SecretKey? driveKey,
  }) async {
    final txId = entityId;

    final cachedData = await _getEntityDataFromSnapshot(
      driveId: driveId,
      txId: txId,
      driveKey: driveKey,
    );

    if (cachedData != null) {
      return cachedData;
    }

    return _getEntityDataFromNetwork(txId: txId);
  }

  Future<Uint8List> _getEntityDataFromNetwork({required String txId}) async {
    final Response data =
        (await httpRetry.processRequest(() => client.api.getSandboxedTx(txId)));

    return data.bodyBytes;
  }

  Future<Uint8List?> _getEntityDataFromSnapshot({
    required String txId,
    required String driveId,
    required SecretKey? driveKey,
  }) async {
    final Uint8List? cachedData = await SnapshotItemOnChain.getDataForTxId(
      driveId,
      txId,
    );

    if (cachedData != null) {
      final isPrivate = driveKey != null;

      if (isPrivate) {
        // then it's base64-encoded
        return base64.decode(String.fromCharCodes(cachedData));
      } else {
        // public data is plain text
        return cachedData;
      }
    }

    return null;
  }
 ```
 
 and then:
 
 ```dart
 final List<Uint8List> responses = await Future.wait(
      entityTxs.map(
        (entity) => _getEntityData(
          entityId: entity.id,
          driveId: driveId,
          driveKey: driveKey,
        ),
      ),
    );
 ```

**Review by thiagocarvalhodev — COMMENTED** [review]
Left just one comment! LGTM!

File: lib/services/arweave/arweave_service.dart (hunk 6 above)
Comment (matibat) [reply lib/services/arweave/arweave_service.dart]: I applied the suggestion, the only difference is that instead of passing the driveKey, I pass a boolean
fbb075edce371bcd1dd24f8ffff5f05c724f8eeb

File: lib/utils/snapshots/snapshot_item.dart (hunk 5 above)
Comment (matibat) [reply lib/utils/snapshots/snapshot_item.dart:196]: Indeed, as that data comes from GQL (at creation time) it shouldn't be malformed unless it was built wrongly. 

Not implemented at this point, but...
This case shouldn't happen because at creation time we're serializing those DriveHistoryTransaction instances back to JSON, which is the JSON we recover here in order to instantiate it again as a DriveHistoryTransaction.

File: lib/services/arweave/arweave_service.dart (hunk 6 above)
Comment (thiagocarvalhodev) [reply lib/services/arweave/arweave_service.dart]: That's even better!! Thanks Mati!

File: lib/services/arweave/arweave_service.dart [hunk 7]
```diff
@@ -298,6 +293,56 @@ class ArweaveService {
...
+        return cachedData;
+      }
+    }
+
+    return null;
+  }
+
+  Future<Uint8List> _getEntityDataFromNetwork({required String txId}) async {
+    final Response data =
+        (await httpRetry.processRequest(() => client.api.getSandboxedTx(txId)));
```
thiagocarvalhodev, line 341: Are there any reasons to don't use our networking package here? `ardrive_http` has retries, as I recall. Can you confirm?

matibat: It was because of the redirects. I will instead try to take from arweave the methods to generate the correct URL instead of eating the redirect

thiagocarvalhodev (approved):

lib/services/arweave/arweave_service.dart
```diff
@@ -260,7 +252,10 @@ class ArweaveService {
             rawEntityData,
             driveKey: driveKey,
           );
+        } else if (entityType == EntityType.snapshot) {
+          // TODO: instantiate entity and add to blockHistory
```
thiagocarvalhodev, line 256: Is this supposed to be implemented in this ticket? I'd say in another one since it's already in QA.

matibat: That's where we instantiate the entities to be put in the DB. We're not doing this for snapshots for now

thiagocarvalhodev (approved):
LGTM

PR #887, opened 2023-02-09
PE-2962: removes unnecessary component

matibat:


matibat:
Fore pushed because I was carrying changes from the wrong (base) branch

thiagocarvalhodev (approved):
LGTM

PR #903, opened 2023-02-16
PE-3012: Resume prepare/sign on tab focus when ArConnect wallet

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/4ovmu33dens1o
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/6acsnett6b2ao

### DIFF
--- a/lib/blocs/create_snapshot/create_snapshot_cubit.dart
+++ b/lib/blocs/create_snapshot/create_snapshot_cubit.dart
@@ -171,8 +171,9 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
 
     final profile = _profileCubit.state as ProfileLoggedIn;
     final wallet = profile.wallet;
+    final isArConnectProfile = await _profileCubit.isCurrentProfileArConnect();
 
-    if (await _profileCubit.isCurrentProfileArConnect()) {
+    if (isArConnectProfile) {
       await _prepareEntityTxArConnect();
     } else {
       _preparedTx = await _arweave.prepareEntityTx(
@@ -186,7 +187,7 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
 
     await _pst.addCommunityTipToTx(_preparedTx);
 
-    if (await _profileCubit.isCurrentProfileArConnect()) {
+    if (isArConnectProfile) {
       await _signTxWithArConnect();
     } else {
       // ignore: avoid_print
--- a/lib/blocs/create_snapshot/create_snapshot_cubit.dart
+++ b/lib/blocs/create_snapshot/create_snapshot_cubit.dart
@@ -173,38 +173,24 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
     final wallet = profile.wallet;
     final isArConnectProfile = await _profileCubit.isCurrentProfileArConnect();
 
-    if (isArConnectProfile) {
-      await _prepareEntityTxArConnect();
-    } else {
-      _preparedTx = await _arweave.prepareEntityTx(
-        snapshotEntity,
-        wallet,
-        null,
-        // We'll sign it just after adding the tip
-        skipSignature: true,
-      );
-    }
+    await _prepareEntityTx(isArConnectProfile);
 
     await _pst.addCommunityTipToTx(_preparedTx);
 
-    if (isArConnectProfile) {
-      await _signTxWithArConnect();
-    } else {
-      // ignore: avoid_print
-      print('Signing snapshot with JSON wallet');
-      await _preparedTx.sign(wallet);
-    }
+    await _signTx(isArConnectProfile);
 
     snapshotEntity.txId = _preparedTx.id;
   }
 
-  Future<void> _prepareEntityTxArConnect() async {
+  Future<void> _prepareEntityTx(bool isArConnectProfile) async {
     final profile = _profileCubit.state as ProfileLoggedIn;
     final wallet = profile.wallet;
 
     try {
       // ignore: avoid_print
-      print('Preparing snapshot transaction with ArConnect');
+      print(
+        'Preparing snapshot transaction with ${isArConnectProfile ? 'ArConnect' : 'JSON wallet'}',
+      );
       _preparedTx = await _arweave.prepareEntityTx(
         _snapshotEntity!,
         wallet,
@@ -213,13 +199,13 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
         skipSignature: true,
       );
     } catch (_) {
-      if (isBrowserTabHidden()) {
+      if (isArConnectProfile && isBrowserTabHidden()) {
         // ignore: avoid_print
         print(
           'Preparing snapshot transaction while user is not focusing the tab. Waiting...',
         );
         await whenBrowserTabIsUnhiddenFuture(
-          _prepareEntityTxArConnect,
+          _prepareEntityTx,
         );
       } else {
         rethrow;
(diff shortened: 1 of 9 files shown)


matibat:
TODO: unit test the ArConnect tab focus stuff when preparing/signing, on the Snapshot Cubit

matibat (commented):
Self-review checkpoint

lib/blocs/create_snapshot/create_snapshot_cubit.dart
```diff
@@ -161,35 +162,95 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
...
     final wallet = profile.wallet;
 
-    final preparedTx = await _arweave.prepareEntityTx(
-      snapshotEntity,
-      wallet,
-      null,
-      // We'll sign it just after adding the tip
-      skipSignature: true,
-    );
+    if (await _profileCubit.isCurrentProfileArConnect()) {
```
karlprieb, line 175: just create a variable for that and use it on both `if` blocks

lib/blocs/create_snapshot/create_snapshot_cubit.dart
```diff
@@ -161,35 +162,95 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
     return snapshotItemToBeCreated;
   }
 
-  Future<CreateSnapshotParameters> _snapshotParametersFromEntityAndData(
+  Future<void> _prepareSnapshotTx(
```
karlprieb, line 165: You can abstract prepareEntity and signTx. As we do the same thing for both JSON wallet and arconnect(we wait for focus here, but the function called is the same with the same params) we can create these abstractions and check if it's arconnect or not there.
At the end we should have a more readable `_prepareSnapshotTx` function

lib/blocs/create_snapshot/create_snapshot_state.dart
```diff
@@ -33,6 +33,7 @@ class ComputeSnapshotDataFailure extends CreateSnapshotState {
   List<Object> get props => [errorMessage];
 }
 
+/// User has insufficient balance to create the snapshot
```
karlprieb, line 36: we don't need this comment. the class name is self-explanatory

karlprieb (changes requested):
Waiting for ArConnect tests. You can mock `whenTabIsUnhiddenFuture` and `isTabHidden` for that

lib/utils/html/implementations/html_web.dart
```diff
@@ -8,6 +8,17 @@ bool isTabHidden() {
 
 late StreamSubscription _onVisibilityChangeStream;
 
+Future<void> whenTabIsUnhiddenFuture(FutureOr<Function> onShow) async {
```
karlprieb, line 11: As this `Future` is meant to be used as a one-time thing, we should close the visibility stream (`closeVisibilityChangeStream`) after execute the callback. If we don't do that, the stream will remain opened and call the listener callback every time we have a window visibility change.

lib/utils/html/implementations/html_web.dart
```diff
@@ -8,6 +8,17 @@ bool isTabHidden() {
...
+  final completer = Completer<void>();
+  _onVisibilityChangeStream = document.onVisibilityChange.listen((event) async {
+    if (!isTabHidden()) {
+      await onShow;
```
karlprieb, line 15: should we execute `onShow` here?

lib/utils/html/html_util.dart
```diff
@@ -1,8 +1,13 @@
+import 'dart:async';
+
 import 'implementations/html_web.dart'
     if (dart.library.io) 'implementations/html_stub.dart' as implementation;
 
 bool isBrowserTabHidden() => implementation.isTabHidden();
```
karlprieb, line 6: Not related to this ticket, but we could use better names for these functions like:
- `isBrowserTabHidden` could be `isTabOnFocus`
- `whenBrowserTabIsUnhidden` could be `whenTabIsOnFocus` and the param to `onFocus`

lib/blocs/create_snapshot/create_snapshot_cubit.dart
matibat: There we are -> ab2fbc5d54ab94c889de3c1694577cfc5e32e516

lib/blocs/create_snapshot/create_snapshot_state.dart
matibat: Agree - a247d3f538ddedbcfe99403cdee29b81a4c5f9ed

lib/utils/html/implementations/html_web.dart
matibat: ad0f569a3d8ae6c664b23490aea05b15b22bf9d5

lib/utils/html/implementations/html_web.dart
matibat: Because it has the type `FutureOr<Function>`: `"The expression doesn't evaluate to a function, so it can't be invoked."`. So I simply await for it.

matibat:
> Waiting for ArConnect tests. You can mock whenTabIsUnhiddenFuture and isTabHidden for that

The thing is that we can't actually mock those methods. We gotta make the class testable.

lib/utils/html/implementations/html_web.dart
karlprieb: No... just call `closeVisibilityChangeStream` in `whenTabIsUnhiddenFuture` after you call `onShow`

lib/utils/html/html_util.dart
karlprieb: Can you rename them in this ticket?

karlprieb:
> The thing is that we can't actually mock those methods. We gotta make the class testable.

Let's to that then! Create a new subtask to implement arconnect tests

lib/utils/html/implementations/html_web.dart
matibat: Like this? a596e3cf80b230d20da1c3e3ae7b5db77bfe4754

lib/utils/html/html_util.dart
matibat: Renamed and updated the logic of isTabFocused to match its name - 06308208ae2b9cfb7b218dfe0c890d6f34f73028

lib/blocs/create_snapshot/create_snapshot_cubit.dart
```diff
@@ -161,35 +162,88 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
...
 
-    return uploadSnapshotItemParams;
+    if (isArConnectProfile) {
+      try {
+        await closeVisibilityChangeStream();
+      } catch (_) {
+        // The stream was not yet open. Nothing to do
+      }
+    }
+
```
matibat, line 223:
```suggestion
```

lib/utils/html/implementations/html_web.dart
```diff
@@ -2,16 +2,28 @@ import 'dart:async';
 
 import 'package:universal_html/html.dart';
 
-bool isTabHidden() {
-  return window.document.visibilityState != 'visible';
+bool isTabFocused() {
```
matibat, line 5: **For the reviewer** I've double checked this many times, but please don't trust me: please be thoughtful when reviewing this change around the codebase as it changed both: the name and the logic to make the opposite. I.e. ensure everywhere where it was called, that it's negated (either by adding the `!` "not" or by removing it)

matibat (commented):
Self-review checkpoint

lib/utils/html/implementations/html_web.dart
karlprieb: Yep 👍

lib/utils/html/implementations/html_web.dart
thiagocarvalhodev: I checked and seems good, Mati!

thiagocarvalhodev (approved):
LGTM!

karlprieb (approved):

PR #906, opened 2023-02-17
PE-3013: Adds extra dialog for the prepare/sign phase when ArConnect profile 

matibat:


lib/components/create_snapshot_dialog.dart
```diff
@@ -140,6 +141,10 @@ Widget _loadingDialog(
   BuildContext context,
   CreateSnapshotState snapshotCubitState,
 ) {
+  bool isArConnectProfile = snapshotCubitState is PreparingAndSigningTransaction
```
thiagocarvalhodev, line 144: Nit: use just `state`

lib/components/create_snapshot_dialog.dart
```diff
@@ -148,23 +153,40 @@ Widget _loadingDialog(
         }
       : null;
 
+  final String title = snapshotCubitState is ComputingSnapshotData
```
thiagocarvalhodev, line 156: These variables are not good to read. Consider implementing a new function with blocks of `if else` instead.

lib/blocs/create_snapshot/create_snapshot_cubit.dart
```diff
@@ -170,6 +170,9 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
     print('About to prepare and sign snapshot transaction');
 
     final isArConnectProfile = await _profileCubit.isCurrentProfileArConnect();
+    emit(PreparingAndSigningTransaction(
```
thiagocarvalhodev, line 173:
```suggestion
    
    emit(PreparingAndSigningTransaction(
```

thiagocarvalhodev (changes requested):
Looking good, but tests are failing. I left some suggestions too.

lib/components/create_snapshot_dialog.dart
matibat: a5b31282dfb30a3ef754a48463282443f92e8a99

lib/components/create_snapshot_dialog.dart
matibat: Here it is - 5360fd581197e1e9aaad8bb976886009890ba255

BTW you might encounter a difference in the logic, it's because requirements were missing details: 
- Make the title of it to be "Finishing things up!" in both cases,
- remove "This might take a while..." from the "Uploading" phase.

To add up: the CANCEL button is removed in this extra phase which is not actually cancellable (already applied in previous commits).

matibat (commented):
Self-review checkpoint

lib/components/create_snapshot_dialog.dart
```diff
@@ -344,13 +381,13 @@ Widget _confirmDialog(
...
-                              snapshotCubitState.arUploadCost,
+                              state.arUploadCost,
                             ),
                           ),
-                          if (snapshotCubitState.usdUploadCost != null)
+                          if (state.usdUploadCost != null)
                             TextSpan(
-                              text: snapshotCubitState.usdUploadCost! >= 0.01
-                                  ? ' (~${snapshotCubitState.usdUploadCost!.toStringAsFixed(2)} USD)'
+                              text: state.usdUploadCost! >= 0.01
```
thiagocarvalhodev, line 389: We use the same block of code on upload too if I'm not mistaken. Can you create a utils function and dedup it?

lib/components/create_snapshot_dialog.dart
```diff
@@ -138,37 +138,74 @@ Widget _explanationDialog(BuildContext context, Drive drive) {
...
+    return appLocalizationsOf(context).thisMayTakeAWhile;
+  } else if (state is PreparingAndSigningTransaction) {
+    if (isArConnectProfile) {
+      return appLocalizationsOf(context).pleaseRemainOnThisTabSnapshots;
+    } else {
+      return appLocalizationsOf(context).thisMayTakeAWhile;
+    }
+  } else {
+    // Description for uploading snapshot
+    return '';
```
thiagocarvalhodev, line 205: Is that intended?

thiagocarvalhodev (changes requested):
Looking good! I'm requesting only the utils function to dedup the logic to show price.

lib/components/create_snapshot_dialog.dart
matibat: Will do!

lib/components/create_snapshot_dialog.dart
matibat: Yes! I want the upload to DO NOT say "This may take a while..." as it previously did, but also I don't want a flicker effect so in that case I would render a Text with no content

lib/components/create_snapshot_dialog.dart
matibat: Factored out the method from (8bb4b749bd5ba88930ccc0b361253674bc71cab6):
- Upload Form
- Create Manifest Form
- Create Snapshot Dialog

Adds unit tests for it: d82ecf0bfcbdc08e56ccebbcda2c95f09dd35dc4

karlprieb (approved):

PR #907, opened 2023-02-22
PE-3048: unit test retry prepare/sign snapshot tx on tab focus when ArConnect

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/7a321udf9qsjg
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/6jtrihk3celt8

### DIFF
--- a/lib/blocs/create_snapshot/create_snapshot_cubit.dart
+++ b/lib/blocs/create_snapshot/create_snapshot_cubit.dart
@@ -32,8 +32,6 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
   @visibleForTesting
   bool throwOnDataComputingForTesting;
   @visibleForTesting
-  bool throwOnPrepareTxForTesting;
-  @visibleForTesting
   bool throwOnSignTxForTesting;
   @visibleForTesting
   bool returnWithoutSigningForTesting;
@@ -55,7 +53,6 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
     required PstService pst,
     required TabVisibilitySingleton tabVisibility,
     this.throwOnDataComputingForTesting = false,
-    this.throwOnPrepareTxForTesting = false,
     this.throwOnSignTxForTesting = false,
     this.returnWithoutSigningForTesting = false,
   })  : _arweave = arweave,
@@ -203,10 +200,6 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
       print(
         'Preparing snapshot transaction with ${isArConnectProfile ? 'ArConnect' : 'JSON wallet'}',
       );
-
-      if (throwOnPrepareTxForTesting) {
-        throw Exception('Throwing on purpose for testing');
-      }
 
       _preparedTx = await _arweave.prepareEntityTx(
         _snapshotEntity!,
--- a/lib/blocs/create_snapshot/create_snapshot_cubit.dart
+++ b/lib/blocs/create_snapshot/create_snapshot_cubit.dart
@@ -28,7 +28,15 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
   final DriveDao _driveDao;
   final ProfileCubit _profileCubit;
   final PstService _pst;
-  final bool _forceFailOnDataComputingForTesting;
+  final TabVisibilitySingleton _tabVisibility;
+  @visibleForTesting
+  bool throwOnDataComputingForTesting;
+  @visibleForTesting
+  bool throwOnPrepareTxForTesting;
+  @visibleForTesting
+  bool throwOnSignTxForTesting;
+  @visibleForTesting
+  bool returnWithoutSigningForTesting;
 
   late DriveID _driveId;
   late Range _range;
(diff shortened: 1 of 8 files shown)


test/blocs/create_snapshot_cubit_test.dart
```diff
@@ -281,6 +289,97 @@ void main() {
...
+          'waits for the tab to be focused again on prepareTx',
+          build: () => CreateSnapshotCubit(
+            arweave: arweave,
+            profileCubit: profileCubit,
+            driveDao: driveDao,
+            tabVisibility: tabVisibility,
+            pst: pst,
+            throwOnPrepareTxForTesting: true,
+          ),
+          // FIXME: Before calling confirm, you have to set throwOnPrepareTxForTesting to false delayed by less than 10 ms
```
matibat, line 342:
```suggestion
```

matibat (commented):
Self-review checkpoint

lib/blocs/create_snapshot/create_snapshot_cubit.dart
```diff
@@ -191,29 +203,38 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
       print(
         'Preparing snapshot transaction with ${isArConnectProfile ? 'ArConnect' : 'JSON wallet'}',
       );
+
+      if (throwOnPrepareTxForTesting) {
```
thiagocarvalhodev, line 207: Instead of using this boolean to throw an exception, you can mock the `_arweave.prepareEntityTx` to throw an Exception on your test. See this example from the `ardrive_auth_test.dart` at new login flow:

```dart
  test('should return false when password is wrong', () async {
      // arrange
      when(() => mockArweaveService.getFirstPrivateDriveTxId(wallet,
              maxRetries: any(named: 'maxRetries')))
          .thenAnswer((_) async => 'some_id');
      // mock cripto derive drive key
      when(
        () => mockArDriveCrypto.deriveDriveKey(
          wallet,
          any(),
          any(),
        ),
      ).thenThrow(Exception('wrong password'));

      // assert
      expectLater(() => arDriveAuth.login(wallet, 'password', ProfileType.json),
          throwsA(isA<AuthenticationFailedException>()));
    });
  });

File: lib/utils/html/html_util.dart [hunk 3]
```diff
@@ -3,18 +3,29 @@ import 'dart:async';
 import 'implementations/html_web.dart'
     if (dart.library.io) 'implementations/html_stub.dart' as implementation;
 
-bool isTabFocused() => implementation.isTabFocused();
+class TabVisibilitySingleton {
```
thiagocarvalhodev, line 6: Nice!

thiagocarvalhodev (changes requested):
@matibat , not sure I follow the variables to throw in the CreateSnapshotCubit. Can you verify if using the `thenThrow` at your mocked class satisfy your needs?

lib/blocs/create_snapshot/create_snapshot_cubit.dart
matibat: Nice catch!

I was initially forcing it to throw with that custom code, but at some point I made it to throw with of the mock (as you just said). That being said: I will remove the custom throw for prepare, BUT I cannot mock Transaction::sign to force it throw SO I'll leave the code for forcing the sign stage to throw as it is if that's OK

matibat: Cubit - 2c9871c9c4a64c74cfd973f3ea077d83d9f8263b
Test - 1a2389285a39271b8ea6d2924b747f87eb514207

lib/utils/html/html_util.dart
karlprieb: We can name it to `TabVisibility` there's no reason to enforce it is a singleton.

matibat: It indeed is a singleton (if you take a look to the constructors):
```dart

  /// Here it is the SINGLE (static) private instance of it.
  static final TabVisibilitySingleton _singleton =
      TabVisibilitySingleton._internal();

  /// The factory will return the same instance of the singleton on every call.
  factory TabVisibilitySingleton() {
    return _singleton;
  }

  /// This is the private constructor which is called once from the static _singleton.
  TabVisibilitySingleton._internal();
```

BTW I chose it to be a singleton so we don't need to be passing multiple instances of this class which doesn't actually track a state. I also considered multiple instances with static methods, but that's not testable so the singleton approach fits well for this case.

matibat: I added a unit test to verify this behavior: 86391b2d71dcdb16cb8d9656e75ecf7510c99614

karlprieb: I know it’s a singleton, I’m just saying we don’t need to name it as singleton

matibat: Ohh got it. In that case, then that's something I've learned from Clean Code: to be really specific on the way we name stuff: "Am I following the composite pattern?" suffix it as so: "MyClassComposite", "Am I following the Singleton pattern?" then suffix the class as so: "MyClassSingleton". But we can take our own approach for naming objects, if we prefer not to mention "singleton" I can remove it

karlprieb: I’m good with that, it’s just something that we don’t do for singletons currently.
We can try to name things that way.

lib/blocs/create_snapshot/create_snapshot_cubit.dart
thiagocarvalhodev: Sounds good! Thanks Mati!

thiagocarvalhodev (approved):

PR #908, opened 2023-02-22
PE-3074: improve upload feedbacks using large uploads

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5uuko5k3sl1f8
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/6ig38u86ket98


lib/blocs/upload/upload_file_checker.dart
```diff
@@ -0,0 +1,37 @@
...
+  UploadFileChecker({
+    required int publicFileSafeSizeLimit,
+    required int privateFileSafeSizeLimit,
+  })  : _publicFileSafeSizeLimit = publicFileSafeSizeLimit,
+        _privateFileSafeSizeLimit = privateFileSafeSizeLimit;
+
+  final int _publicFileSafeSizeLimit;
+  final int _privateFileSafeSizeLimit;
+
+  Future<bool> hasFileAboveSafePublicSizeLimit(
```
matibat, line 13: Why is that we don't need to know the names of the public ones?

matibat: I guess it's because for private ones, we have a different limit with a different UI (?

lib/l10n/app_en.arb
```diff
@@ -1384,7 +1384,7 @@
   "@uploadFailed": {
     "description": "Upload Failed dialog title"
   },
-  "yourUploadFailed": "Your upload failed. You have not been charged for the failed attempt.",
+  "yourUploadFailed": "Your upload failed.",
```
matibat, line 1387: Are we sure we do want to apply this change to EVERY modal using it, or simply for file uploads? If the latter, then we should instead create a new ARB entry, while preserving this one intact.

I think this change would affect (apart from file uploads): manifests, snapshots, and maybe drive/folder creation as well.

lib/components/upload_form.dart
```diff
@@ -460,6 +463,39 @@ class UploadForm extends StatelessWidget {
...
+                TextButton(
+                  onPressed: () => Navigator.of(context).pop(false),
+                  child: Text(appLocalizationsOf(context).cancelEmphasized),
+                ),
+                ElevatedButton(
+                  onPressed: () => context
+                      .read<UploadCubit>()
+                      .prepareUploadPlanAndCostEstimates(),
+                  child: Text(
+                    appLocalizationsOf(context).proceed,
```
matibat, line 494: It's more a question for Lucas, but... Don't we want this text to be CAPITALIZED? We usually do so for buttons (e.i. CANCEL, OK).

If that's the case then I suggest renaming `proceedCongestionEmphasized` to `proceedEmphasized` and re-utilize it here.

test/blocs/upload/upload_file_checker_test.dart
```diff
@@ -0,0 +1,173 @@
...
+            ioFile: ioFileUnderSafeLimit,
+            parentFolderId: 'parentFolderId',
+          ),
+        ],
+      );
+
+      expect(result, []);
+    });
+
+    // test empry list
```
matibat, line 119: Nit
```suggestion
```

test/blocs/upload_cubit_test.dart
```diff
@@ -415,7 +581,7 @@ void main() {
           ],
         );
       },
-      skip: 'File size limit is too high to generate a file for',
+      // skip: 'File size limit is too high to generate a file for',
```
matibat, line 584: Nit: If we're no longer skipping we can remove the comment
```suggestion
```

matibat (commented):
Left some comments/questions but no blockers. It's looking good overall!

lib/blocs/upload/upload_file_checker.dart
thiagocarvalhodev: Because we only need to know if we have a file above the safe limit, then we'll show a generic warning.

lib/l10n/app_en.arb
thiagocarvalhodev: We only use it for file uploads - drag and drop and selecting the file.

lib/components/upload_form.dart
thiagocarvalhodev: We're using the text defined by Ali, but in any case we'll change it on redesign, so I think it's ok to move with this

lib/blocs/upload/limits.dart
```diff
@@ -1,3 +1,7 @@
+import 'package:ardrive/utils/data_size.dart';
+
 const privateFileSizeLimit = 104857600;
```
karlprieb, line 3: can we use `MiB(100).size` here?
and `GiB(1).size` for `mobilePrivateFileSizeLimit`

lib/blocs/upload/limits.dart
```diff
@@ -1,3 +1,7 @@
...
-const publicFileSizeLimit = 1288490189 * 100; // TODO: Decide a sane limit for stream transctions
+
 const mobilePrivateFileSizeLimit = 1073741823;
+// 5GiB
```
karlprieb, line 6: we don't need this comment `GiB(5).size` already makes a lot of sense

lib/blocs/upload/limits.dart
thiagocarvalhodev: [e6f901d](https://github.com/ardriveapp/ardrive-web/pull/908/commits/e6f901d3610dc491ed0a51c8f024d52f2491607e)

lib/blocs/upload/limits.dart
thiagocarvalhodev: [e6f901d](https://github.com/ardriveapp/ardrive-web/pull/908/commits/e6f901d3610dc491ed0a51c8f024d52f2491607e)

karlprieb (approved):

karlprieb (approved):

PR #912, opened 2023-02-23
PE-3003: Stay on Page ArConnect Warning on snapshot creation

matibat:
Changes: 
- Makes the prepare and sign phases to retry it -if the tab wasn't focused- after focused again when utilizing ArConnect.
- Adds an extra phase to the modal UI: "Finishing things up!" where it's preparing/singing.
  - In the case of ArConnect, this phase will have a warning asking the user to do not leave the page til' finished.
  - In the case of a JSON wallet, there's no warning.
  - Hides the CANCEL button in this stage, as the action is not cancellable at this point.
- Adds unit tests for this new behavior on the cubit.
- Refactors the focus methods at html_util.dart to be more meaningful on its names
  - The change for `isBrowserTabHidden` to be named `isTabFocused` implied a negation on the logic (when true, it's now false and vice-versa)

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5g5vs07v6oir8
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/2c2ah2lubqmc0


matibat:
Here I show the UI for the new modal stage: `"Finishing things up!"` where we prepare and sign the transaction, which comes after `"Determining size and cost of snapshot"`.
- **JSON wallet** (to the left)
  - Description doesn’t change: `"This may take a while…"`
- **ArConnect wallet** (to the right)
   - Description: `"Your Web3 wallet is signing your transactions. Please stay on this page while we finish things up!"`
- For both cases the `CANCEL` button gets hidden as the prepare/sign stage is not cancellable.

https://user-images.githubusercontent.com/22564541/220997958-2efd86c3-db57-4457-9964-d55a4c3d3283.mov

PR #955, opened 2023-03-16
feat(localization spanish): nit fix on punctuation marks PE-3171

matibat:


agsuy (approved):
Cheap translation update

PR #956, opened 2023-03-17
PE-3132: ArConnect logout race condition

matibat:


lib/services/arconnect/arconnect.dart
```diff
@@ -1,20 +1,123 @@
...
+          hasPermissions = value;
+          return value;
+        },
+      );
+    });
+
+    return hasPermissions;
+  }
+
+  Future<bool> safelyCheckPermissions({
```
matibat, line 47: **For the reviewer** See here the relevant changes

lib/services/arconnect/arconnect.dart
```diff
@@ -1,20 +1,123 @@
...
+        return true;
+      }
+
+      print(
+          '[ArConnectService::safelyCheckPermissions] Retrying in ${cooldownDuration.inMilliseconds} ms ... ($triesLeft tries left)');
+      await Future.delayed(cooldownDuration);
+    }
+
+    if (!tabVisibility.isTabFocused()) {
+      throw FocusError('Tab is not focused');
```
arielmelendez, line 78: Could be worth logging here that we unexpectedly encountered an out of focus condition.

matibat:
Replaced by #972

PR #957, opened 2023-03-20
Stream download + Encryption/Decryption

elliotsayes:
Private T1 file content transaction are now uploaded using `AES256-CTR` encryption. This is an "unauthenticated" encryption mode. Inauthentic data could only conceivable be downloaded if there was a malicious gateway or GraphQL endpoint. All other private transaction types (metadata, file content dataitems) continue to use AES-GCM.

To replace the authentication from `AES256-GCM`, the new stream downloader also authenticates the data against the transaction itself. This actually provides stronger guarantees that the data is not only encrypted with the same key, but also corresponds to the expected transaction id / owner!

Process:
```
                    ┌───────────────┐
                 ┌─►│Authentication ├─────────┐
┌────────────────┤  └───────────────┘         │
│Downloaded Data │                            ▼
└────────────────┤  ┌───────────────┐  ┌──────────────┐
                 └─►│Decryption     ├─►│Saving        │
                    └───────────────┘  └──────────────┘
``` 

Authentication happens in parallel with decryption/saving pipeline. If the data is inauthentic it will interrupt the saving process and do its best to clean up the downloaded data. Note that the authentication process is also carried out on public downloads for web.

Download progress is also shown in the UI, and upon completion the user is shown the authenticated owner.

Incomplete:
- The verified transaction owner is not compared to the address of the drive owner - I couldn't work out where to get that so I just display it to the user for them to check themselves.
- The download test is broken: `personal_file_download_cubit_test.dart`. I tried but couldn't work out how to do this properly sorry :(


lib/core/download_service.dart
```diff
@@ -1,28 +1,61 @@
...
+  
+  Stream<Uint8List> downloadStream(String fileTxId, int fileSize, {Completer<String>? cancelWithReason});
 }
 
 class _DownloadService implements DownloadService {
   _DownloadService(this._arweave);
 
   final ArweaveService _arweave;
 
+  Uri txSubdomainGateway(String fileTxId) {
```
karlprieb, line 23: why not use https://github.com/ardriveapp/arweave-dart/blob/224bfe4a84d0bfe622f28b0a2305e060a70d08c4/lib/src/api/sandbox.dart#L6 ?

lib/blocs/file_download/personal_file_download_cubit.dart
```diff
@@ -111,10 +112,30 @@ class ProfileFileDownloadCubit extends FileDownloadCubit {
...
+    final authStream = splitStream.split();
 
-    if (drive.drivePrivacy == DrivePrivacy.private) {
+    // Calling `close()` indicates that no further streams will be created,
+    // causing splitStream to function without an internal buffer.
+    // The future will be completed when both streams are consumed, so we
+    // shouldn't await it here.
+    unawaited(splitStream.close());
+
+    Stream<Uint8List> saveStreamDecrypted;
```
karlprieb, line 135: Add `Late` here to make it explicit

lib/services/crypto/ciphers.dart
```diff
@@ -0,0 +1,45 @@
+import 'dart:async';
```
karlprieb, line 1: I guess `cipherName` is not optional in all these functions

lib/services/arweave/arweave_service.dart
```diff
@@ -111,6 +111,10 @@ class ArweaveService {
         );
   }
 
+  Future<T?> getTransaction<T extends Transaction>(String txId) async {
+    return client.transactions.get<T>(txId);
```
karlprieb, line 115: we can use `getSandboxedTx` instead

lib/services/crypto/ciphers.dart
```diff
@@ -0,0 +1,45 @@
...
+    Cipher.aes256ctr: AesCtrStream.fromKeyData,
+  };
+  final ctr = ctrs[cipherName];
+  if (ctr == null) throw ArgumentError();
+  final impl = await ctr(keyData);
+  return impl;
+}
+
+FutureOr<EncryptStream> cipherStreamEncryptImpl(
+  String? cipherName, {
```
karlprieb, line 35: `chipherName` is not optional

test/blocs/personal_file_download_cubit_test.dart
```diff
@@ -50,6 +53,45 @@ class MockARFSDrive extends ARFSDriveEntity {
   });
 }
 
+class MockIOFile extends IOFile {
```
matibat, line 56: We usually put the mocks into this file here: `test/test_utils/mocks.dart` (already imported in this file)

test/services/crypto/stream_aes_test.dart
```diff
@@ -0,0 +1,67 @@
+// Fails to compile when testing on chrome
+// @TestOn('vm')
+@Skip('Discrepancies between app and test environment with package:webcrypto')
```
matibat, line 3: What do we need to do to enable this test?

matibat (commented):
It's looking good to me overall! Really nice improvements, thanks 🥇 

There are a couple of unresolved comments. And I think none of mine are blockers, just nits.

elliotsayes:
Waiting on some feedback for other comments. I also notices that the following PR was closed: https://github.com/ardriveapp/arweave-dart/pull/41 - changes therein are used by this PR

PR #961, opened 2023-03-20
PE-3132: ArConnect logout

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/10skt42tvh8a0
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/12slse0le89lo


matibat:
GH Actions failing because of a bad deppendency

lib/blocs/sync/utils/arconnect_connectivity.dart
matibat, line 1: Oops! Unintended file committed! - TODO: remove

lib/utils/html/html_util.dart
```diff
@@ -7,13 +7,30 @@ class TabVisibilitySingleton {
   static final TabVisibilitySingleton _singleton =
       TabVisibilitySingleton._internal();
 
+  /// TODO: Cancel these subscriptions
```
matibat, line 10: TODO: create a tech debt "Correctly handle Stream Subscriptions"

lib/utils/html/implementations/html_web.dart
matibat, line 1: TODO: Cleanup the newly introduced methods

matibat (commented):
Self-review checkpoint

lib/blocs/sync/utils/arconnect_connectivity.dart
matibat: d5be4a470778e323e0f4edba193f363bf54c9f11

lib/utils/html/implementations/html_web.dart
matibat: 597779f404fdd15a23aa0f70b535fc05069d921d

lib/blocs/profile/profile_cubit.dart
```diff
@@ -66,7 +66,18 @@ class ProfileCubit extends Cubit<ProfileState> {
     }
 
     // ArConnect connected to expected wallet - route to login screen
-    if (await arconnect.checkPermissions() &&
+    late bool hasPermissions;
```
arielmelendez, line 69: Using `late` leaves you responsible for making sure the assignment takes place before it's used and Dart will let you shoot yourself in the foot if you're so inclined. I tend to use the following syntax for type-safe conditional assignments (an anonymous closure declared and executed in place):
```
    bool hasPermissions = await (() async {
      try {
        return await arconnect.safelyCheckPermissions();
      } catch (e) {
        print(
          '[ProfileCubit] ArConnect permissions check failed - waiting for tab focus',
        );
        return await arconnect.safelyGetPermissionsWhenTabFocused();
      }
    })();
```

lib/blocs/profile/profile_cubit.dart
```diff
@@ -66,7 +66,18 @@ class ProfileCubit extends Cubit<ProfileState> {
...
+    try {
```
arielmelendez, line 70: This pattern does not look so user-friendly. It implies that anyone who ever calls `safelyCheckPermissions` should wrap with `try` then call `safelyGetPermissionsWhenTabFocused` upon exception. Remember one of the golden rules of programming:
`Make interfaces easy to use correctly and hard to use incorrectly.`
The pattern you're using here does not seem to fit those criteria. If I were just reading the API names, I'd have a hard time choosing which to call. They both have `safely` in their name, so they appear "safe". Despite that appearance, at least one of them `safelyCheckPermissions` might actually THROW! That doesn't seem so `safe`. This needs work.

lib/blocs/sync/sync_cubit.dart
```diff
@@ -151,10 +151,15 @@ class SyncCubit extends Cubit<SyncState> {
   }
 
   Future<void> arconnectSync() async {
-    if (_tabVisibility.isTabFocused() &&
-        await _profileCubit.logoutIfWalletMismatch()) {
-      emit(SyncWalletMismatch());
-      return;
+    final isTabFocused = _tabVisibility.isTabFocused();
+    if (isTabFocused) {
+      if (await _profileCubit.logoutIfWalletMismatch()) {
```
arielmelendez, line 156: To the point about making interfaces easy to use correctly and hard to use incorrectly, might there be arguments for or against simply doing the focus check inside of `checkIfWalletMismatch`?

lib/services/arconnect/arconnect.dart
```diff
@@ -12,6 +15,65 @@ class ArConnectService {
...
+        print(
+          '[ArConnectService::safelyCheckPermissions] Permissions not granted on try ${maxRetries - triesLeft}',
+        );
+        await Future.delayed(cooldownDuration);
+      }
+    }
+
+    if (!tabVisibility.isTabFocused()) {
+      print('[ArConnectService::safelyCheckPermissions] Tab is not focused');
+      throw FocusError('Tab is not focused');
```
arielmelendez, line 67: Not a great idea to throw from a function with `safe` in the name.

lib/utils/html/implementations/html_stub.dart
```diff
@@ -1,13 +1,23 @@
...
-bool isTabFocused() {
+bool isTabVisible() {
   return true;
 }
 
 Future<void> onTabGetsFocusedFuture(FutureOr<Function> onFocus) async {
   return;
 }
 
+StreamSubscription onTabBlurred(Function onBlur) {
```
arielmelendez, line 11: Do we still need these if we have access to `hasFocus` now? They're complementary approaches, but if we don't need the event based piece right now, then it's just less code to review, test, and support.

lib/utils/html/implementations/html_web.dart
```diff
@@ -1,30 +1,54 @@
...
+  return onVisibilityChangeStream;
+}
+
+bool isTabFocused() {
+  final hasFocus = context.callMethod('hasFocus');
+  print('[isTabFocused] hasFocus: $hasFocus');
+  return hasFocus;
+}
+
+void onTabGetsFocused(Function onFocus) {
```
arielmelendez, line 49: What's the point of this function? It doesn't look like anyone is using it. Also it leaks StreamSubscriptions each time it's called subsequently....

web/index.html
```diff
@@ -38,4 +39,4 @@
 
 </body>
 
-</html>
\ No newline at end of file
+</html>
```
arielmelendez, line 42: Did the code formatter remove trailing whitespace?

lib/utils/html/implementations/html_web.dart
```diff
@@ -1,30 +1,54 @@
...
 import 'package:universal_html/html.dart';
 
-bool isTabFocused() {
+bool isTabVisible() {
   return window.document.visibilityState == 'visible';
 }
 
 late StreamSubscription _onVisibilityChangeStream;
 
 Future<void> onTabGetsFocusedFuture(FutureOr<Function> onFocus) async {
```
arielmelendez, line 13: I'm not a fan of this function for a handful of reasons:
1) It's both a Future allowing you to await for it to complete if you wanted to AND it has a callback function. That's a strange mix of paradigms
2) It's not clear as a caller what it will do if I'm focused already... will it call the callback? Will it not fire the callback until I've gone blurred and then returned to focus? It's not clear in the function name. From my reading of the code, it appears like it will wait until the NEXT time we become focused. This seems like the kind of behavior that you might want to be configurable.
3) Since the passed closure is a `FutureOr<Function>`, it's not clear whether the stream handler function will (or should?) await on it or simply call it (i.e. fire and forget).
4) It's not clear what the error propagation pattern is or should be for the callback function.

lib/services/arconnect/arconnect.dart
```diff
@@ -12,6 +15,65 @@ class ArConnectService {
   /// requests permissions.
   Future<void> connect() => implementation.connect();
 
+  Future<bool> safelyGetPermissionsWhenTabFocused() async {
+    late bool hasPermissions;
+
+    // FIXME: inject this value for testing purposes
```
arielmelendez, line 21: Sure, and a new instance could be the default value if not provided.

lib/utils/html/implementations/html_stub.dart
```diff
@@ -1,13 +1,23 @@
 import 'dart:async';
 
-bool isTabFocused() {
+bool isTabVisible() {
```
arielmelendez, line 3: With this change I would have expected every caller around the app to have changed the called name to `isTabVisible` as well and for `isTabFocused` to only be used around the ArConnect stuff we are updating. Have you inadvertently caused all the callers of `isTabFocused` to have potentially changed their behavior?

arielmelendez (changes requested):
Needs a handful of changes for clarity.

lib/utils/html/implementations/html_web.dart
```diff
@@ -1,30 +1,54 @@
...
+      onBlur();
+    },
+    cancelOnError: false,
+  );
+  return onVisibilityChangeStream;
+}
+
+StreamSubscription onTabFocused(Function onFocus) {
+  final onVisibilityChangeStream = window.onFocus.listen(
+    (event) {
```
karlprieb, line 35: you don't need `event` here as it's not used

karlprieb: same thing for `onTabBlurred`

lib/utils/html/html_util.dart
matibat: It is for the html_web.dart source file - https://ardrive.atlassian.net/browse/PE-3295

lib/blocs/profile/profile_cubit.dart
matibat: Both methods throw, it is "safe" in the sense that it doesn't return a false negative (i.e. returning `false` when the race condition takes place). Anyway, your comment makes sense to me, so I'd make the `safelyGetPermissionsWhenTabFocused` method instantly return if we do have the permissions

lib/utils/html/implementations/html_web.dart
karlprieb: 1. Is there a place where you will not wait for the event? If so, we can have only a future that returns when you get focused and the caller can do whatever they need. If we still want to let a function handle the callback when it's ready, so the caller can move on with their code, the `onTabGetsFocused` can be used.
2. Agreed
3. I think it's clear that it will `await` on it or just call, depending on the callback that you pass.
4. 100% agreed

lib/services/arconnect/arconnect.dart
karlprieb: I really don't get the naming... why use safe? Does it mean that it's safe to call and it will work every time? Will it return something even when we get an error? I would just return a boolean and the caller decides what to do.

lib/services/arconnect/arconnect.dart
```diff
@@ -12,6 +15,65 @@ class ArConnectService {
   /// requests permissions.
   Future<void> connect() => implementation.connect();
 
+  Future<bool> safelyGetPermissionsWhenTabFocused() async {
```
karlprieb, line 18: I understand that `safelyGetPermissionsWhenTabFocused` should be used when your willing to wait for the next on focus event to check permissions and `safelyCheckPermissions` is for when you just want to check permissions whenever the focus state is.
I don't see a clear usage for `safelyCheckPermissions`, every time we need to check for permissions is because is important to have reliable data, so why we just dont have a function that will abstract both of these functions?
It could check for focus, if we have focus it tries to get permissions, if not it will wait for the next on focus event and try to get the permissions returning a boolean.

lib/blocs/profile_add/profile_add_cubit.dart
```diff
@@ -88,7 +88,7 @@ class ProfileAddCubit extends Cubit<ProfileAddState> {
       emit(ProfileAddUserStateLoadInProgress());
       _profileType = ProfileType.arConnect;
 
-      if (!(await arconnect.checkPermissions())) {
+      if (!(await arconnect.safelyCheckPermissions())) {
```
karlprieb, line 91: Here we should just use that abstraction that I commented before where it will return the boolean if it's focused or wait until it gets focused to check the permissions.
Without that the caller will have to do that if it really wants to pick the wallet.

PR #987, opened 2023-03-29
PE-3352: Adds the `owners` field to the GQL queries

matibat:
Changes
- Makes `DriveEntityHistory` and `SnapshotEntityHistory` queries to take the owner address.
- Refactors the callees to have the address injected into those.
  - For snapshot creation it comes from the logged in profile
  - For sync it comes from the DriveDao

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/7e9qf7is7vogo
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/555fgqh9pkid8

### DIFF
--- a/lib/blocs/create_snapshot/create_snapshot_cubit.dart
+++ b/lib/blocs/create_snapshot/create_snapshot_cubit.dart
@@ -37,7 +37,7 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
   bool returnWithoutSigningForTesting;
 
   late DriveID _driveId;
-  late String _driveOwner;
+  late String _ownerAddress;
   late Range _range;
   late int _currentHeight;
   late Transaction _preparedTx;
@@ -74,7 +74,8 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
       return;
     }
 
-    _driveOwner = await _arweave.getOwnerForDriveEntityWithId(driveId) ?? '';
+    final profileState = _profileCubit.state as ProfileLoggedIn;
+    _ownerAddress = profileState.walletAddress;
 
     _setTrustedRange(range);
 
@@ -150,7 +151,7 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
       _driveId,
       minBlockHeight: _range.start,
       maxBlockHeight: _range.end,
-      ownerAddress: _driveOwner,
+      ownerAddress: _ownerAddress,
     );
 
     // transforms the stream of arrays into a flat stream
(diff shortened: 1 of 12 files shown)


matibat:
TODO: add logging

lib/blocs/create_snapshot/create_snapshot_cubit.dart
```diff
@@ -73,6 +74,8 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
       return;
     }
 
+    _driveOwner = await _arweave.getOwnerForDriveEntityWithId(driveId) ?? '';
```
karlprieb, line 77: we can get the owner from the wallet

lib/services/arweave/arweave_service.dart
```diff
@@ -119,8 +119,9 @@ class ArweaveService {
 
   Stream<SnapshotEntityTransaction> getAllSnapshotsOfDrive(
     String driveId,
-    int? lastBlockHeight,
-  ) async* {
+    int? lastBlockHeight, {
+    required String ownerAddress,
```
karlprieb, line 123: what happens to the response if the `ownerAddress` is an empty string?

lib/blocs/sync/utils/sync_drive.dart
```diff
@@ -19,6 +19,9 @@ Stream<double> _syncDrive(
   final drive = await driveDao.driveById(driveId: driveId).getSingle();
   final startSyncDT = DateTime.now();
 
+  final String owner =
```
arielmelendez, line 22: I think this is fine as a fallback (lazy load). But it would be nice if this function accepted the drive owner and only did the lazy load if necessary.

lib/services/arweave/arweave_service.dart
matibat: There are no transactions with `owner == ''`. So when you try that it returns a valid empty response.

lib/blocs/sync/utils/sync_drive.dart
karlprieb: For sure

lib/blocs/create_snapshot/create_snapshot_cubit.dart
matibat: f4eb4b08ee7a743676123d805e674615e144304f

lib/blocs/sync/utils/sync_drive.dart
matibat: We don't actually need to lazily initialize: 2702d8b77cc50cc1ed3d0ede5a826d99934ecfa9

matibat (commented):
Self-review checkpoint.

karlprieb (approved):

thiagocarvalhodev (approved):
LGTM.

PR #1001, opened 2023-04-07
PE-3437: Big drive sync improvements PoC

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/3o918k5fnrca0
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/3ouoongkt4va8

### DIFF
--- a/lib/blocs/sync/sync_cubit.dart
+++ b/lib/blocs/sync/sync_cubit.dart
@@ -252,34 +252,36 @@ class SyncCubit extends Cubit<SyncState> {
       _syncProgress = _syncProgress.copyWith(drivesCount: drives.length);
       logSync('Current block height number $currentBlockHeight');
       final driveSyncProcesses = drives.map(
-        (drive) => _syncDrive(
-          drive.id,
-          driveDao: _driveDao,
-          arweave: _arweave,
-          ghostFolders: ghostFolders,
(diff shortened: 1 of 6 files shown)


lib/blocs/sync/sync_cubit.dart
```diff
@@ -252,34 +252,36 @@ class SyncCubit extends Cubit<SyncState> {
...
-              ? 0
-              : calculateSyncLastBlockHeight(drive.lastBlockHeight!),
-          currentBlockHeight: currentBlockHeight,
-          transactionParseBatchSize:
-              200 ~/ (_syncProgress.drivesCount - _syncProgress.drivesSynced),
-          ownerAddress: drive.ownerAddress,
-        ).handleError(
-          (error, stackTrace) {
+        (drive) async* {
+          try {
```
matibat, line 256: It previously utilized Stream.handleError, now replaced with a try/catch block that will fail to the first error delivered by the stream. Also, as it's an async* method and we're try-catch-ing from the yielded stream,  the stream automatically finishes as there's no other instruction after the try..catch block.

PR #1003, opened 2023-04-07
PE-3439: Request transaction confirmations in bigger batches

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/49jicdc8k585o
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/6j246ofh5rgtg

### DIFF
--- a/lib/services/arweave/arweave_service.dart
+++ b/lib/services/arweave/arweave_service.dart
@@ -209,8 +209,8 @@ class ArweaveService {
     List<DriveEntityHistory$Query$TransactionConnection$TransactionEdge$Transaction>
         entityTxs,
     SecretKey? driveKey,
-    String? owner,
     int lastBlockHeight, {
+    required String ownerAddress,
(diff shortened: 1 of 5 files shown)


lib/services/arweave/arweave_service.dart
```diff
@@ -749,7 +749,7 @@ class ArweaveService {
 
     // Chunk the transaction confirmation query to workaround the 10 item limit of the gateway API
```
karlprieb, line 750: Can you update the comment and verify with GW which is the actual limit?

matibat: I removed the comment directly. The limit of maximum queried items for GQL is 100 (you can verify it by asking for 101 items while getting 100).

karlprieb (dismissed):

karlprieb (approved):

PR #1042, opened 2023-04-24
PE-2786: Utilize sandboxed url for Quick Sync

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5fu46ovqnj81g
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/0mvg6hsnuavro

### DIFF
--- a/lib/utils/snapshots/snapshot_item.dart
+++ b/lib/utils/snapshots/snapshot_item.dart
@@ -138,7 +138,6 @@ class SnapshotItemOnChain implements SnapshotItem {
   String? _cachedSource;
   int _currentIndex = -1;
 
-  @override
   final ArweaveService _arweave;
 
   static final Map<String, Cache<Uint8List>> _jsonMetadataCaches = {};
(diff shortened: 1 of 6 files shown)


lib/utils/snapshots/snapshot_item.dart
```diff
@@ -135,10 +135,12 @@ abstract class SnapshotItem implements SegmentedGQLData {
 class SnapshotItemOnChain implements SnapshotItem {
   final int timestamp;
   final TxID txId;
-  final String arweaveUrl;
   String? _cachedSource;
   int _currentIndex = -1;
 
+  @override
```
thiagocarvalhodev, line 141: Why the override here? `_arweave` seems not to be a property from `SnapshotItem` interface.

thiagocarvalhodev (changes requested):
Everything looking good but the `@override` on the `_arweave`.

lib/utils/snapshots/snapshot_item.dart
matibat: 1e5f436b770134a7ac6bb26a3da79663e330aa23

PR #1049, opened 2023-04-26
PE-3559: skip sync for arconnect if tab is not focused

matibat:


thiagocarvalhodev (approved):
LGTM. Thanks for the clean up

thiagocarvalhodev (approved):

PR #1068, opened 2023-05-01
Localizely: Translations update

matibat:
Exported the most recent translations from Localizely

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -48,7 +48,7 @@
   "@anyoneCanAccessThisFile": {
     "description": "Warning for the user that is about to share a file"
   },
-  "appVersion": "Version {version}.",
+  "appVersion": "Version {version}",
   "@appVersion": {
     "description": "App version number",
     "placeholders": {
@@ -61,6 +61,10 @@
   "arConnectRemainOnThisTab": "CAUTION: Your Web3 wallet is signing your transactions. Please remain on this tab until it has completed.",
   "@arConnectRemainOnThisTab": {
     "description": "Warning for the user to do not leave the tab"
+  },
+  "arConnectWalletDoestNotMatchArDriveWallet": "Your ArConnect wallet does not match your ArDrive wallet. Please try again.",
+  "@arConnectWalletDoestNotMatchArDriveWallet": {
+    "description": "Your ArConnect wallet does not match your ArDrive wallet. Please try again."
   },
   "ardriveIsntJustAnotherCloudSyncApp": "ArDrive isn’t just another cloud sync app. It’s the beginning of a permanent hard drive.",
   "@ardriveIsntJustAnotherCloudSyncApp": {
@@ -89,6 +93,10 @@
   "back": "Back",
   "@back": {
     "description": "Go back"
+  },
+  "backButtonOnboarding": "Back",
+  "@backButtonOnboarding": {
+    "description": "Text of the back button onboarding"
   },
   "backEmphasized": "BACK",
   "@backEmphasized": {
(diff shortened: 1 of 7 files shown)


karlprieb:
With the new Flutter ARB syntax, we lost some plurals for Mandarin and Cantonese. Knowing that the follow six keys had their translations simplified and more generic for Mandarin and Cantonese:
- filesWillBeUploadedPublicly
- filesTooLarge
- foldersWithTheSameNameAlreadyExists
- duplicateFolders
- duplicateFiles
- filesWithTheSameNameAlreadyExists

lib/l10n/app_en.arb
```diff
@@ -90,6 +94,10 @@
   "@back": {
     "description": "Go back"
   },
+  "backButtonOnboarding": "Back",
+  "@backButtonOnboarding": {
+    "description": "Text of the back button onboarding"
+  },
```
matibat, line 100: I would make `back` to be `goBack`, and `backButtonOnBoarding` to be `back`. So we would have `goBack`, `back`, and `backEmphasized`. Although it's not so easy to apply: we would have to change every ARB file and update the Localizely Dashboard.

Why: we don't care about WHERE in the app the text is being used, but WHAT is it for.

lib/l10n/app_en.arb
```diff
@@ -240,6 +258,10 @@
   "@copyRootFolderTxID": {
     "description": "Copy transaction ID of Root Folder tooltip"
   },
+  "copyTooltip": "Copy",
+  "@copyTooltip": {
+    "description": "Tooltip for copy button"
+  },
```
matibat, line 264: Same comment: we don't have to mention that it is gonna be used in a tooltip. I would call it `copy` alone and the description would explain what "copy" means: "To copy a text"

lib/l10n/app_en.arb
```diff
@@ -258,6 +280,10 @@
   "@create": {
     "description": "Create an drive"
   },
+  "createAndConfirmPassword": "Please create and confirm your password. You will use this password to log in.",
+  "@createAndConfirmPassword": {
+    "description": "Form title to create a password"
```
matibat, line 285: Same: it's talking about WHERE and not WHAT
```suggestion
    "description": "Asks the user to create a password"
```

lib/l10n/app_en.arb
```diff
@@ -308,6 +334,10 @@
   "@creatingFolderEmphasized": {
     "description": "Folder is being created. Emphasized with upper case"
   },
+  "darkMode": "DARK MODE",
+  "@darkMode": {
+    "description": "DARK MODE"
+  },
```
matibat, line 340: The standard we chose for any text that's CAPITALIZED is to call it "emphasized" and to explain in the description: "Emphasized with upper case".
```suggestion
  "darkModeEmphasized": "DARK MODE",
  "@darkModeEmphasized": {
    "description": "Refers to the dark theme of the app. Emphasized with upper case"
  },
```

lib/l10n/app_en.arb
```diff
@@ -592,6 +642,10 @@
   "@fileFailedToDownloadFileAboveLimit": {
     "description": "Download failure message when a file is too big"
   },
+  "fileFailedToDownloadFileAbovePublicLimit": "The file(s) you have selected is too large to download",
+  "@fileFailedToDownloadFileAbovePublicLimit": {
+    "description": "Download failure message when a file is too big"
+  },
```
matibat, line 648: Can't we use plurals here?

lib/l10n/app_en.arb
```diff
@@ -956,6 +1014,10 @@
   "@letsGetStartedEmphasized": {
     "description": "Let's start setting up the account. Emphasized with upper case"
   },
+  "lightMode": "LIGHT MODE",
+  "@lightMode": {
+    "description": "LIGHT MODE"
+  },
```
matibat, line 1020: Same comment than `darkMode`.

matibat (commented):
Left some comments, none of them are blockers but maybe tech debts

PR #1069, opened 2023-05-01
PE-3773: upgrade Flutter to version 3.10.0

karlprieb:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/2e4oj2aqf49k0
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/1p09l7o93lru0


flake.nix
```diff
@@ -0,0 +1,24 @@
...
+      in
+      {
+        devShells.default = pkgs.mkShell {
+          packages = with pkgs; [
+            ruby
+            cocoapods
+            fastlane
+          ];
+          shellHook = ''
+            export PATH="$PWD/.fvm/flutter_sdk/bin:$PATH"
```
matibat, line 19: Should it be $PWD or $HOME?

localizely.yml
matibat, line 1: Nice!

matibat (approved):
LGTM - Left one single question

matibat:
BTW there are minor conflicts here

flake.nix
karlprieb: PWD :)

matibat (approved):
Nice!

flake.nix
karlprieb: > So then flake is for automation only, and we're not supposed to use it manually. Is it right?
flake is part of the nix package manager. You can use it manually if you want to
Currently I'm the only who uses it, but we can use it to automate stuff on our pipeline when we feel so
> I was asking because my .fvm directory is inside my $HOME - $PWD refers to the current directory
you should have both... on .fvm/flutter_sdk is a synlink to your $HOME/fvm/...

matibat: Ahh great! Thanks!

thiagocarvalhodev (approved):

pubspec.yaml
```diff
@@ -31,15 +31,15 @@ dependencies:
...
   ardrive_io:
     git:
       url: https://github.com/ar-io/ardrive_io.git
-      ref: v1.2.0
+      ref: PE-3773-upgrade-flutter-to-v-3-10-0
   ardrive_ui:
     git:
       url: https://github.com/ar-io/ardrive_ui.git
-      ref: v1.1.0
+      ref: PE-3773-upgrade-flutter-to-v-3-10-0
```
matibat, line 42: I don't actually recall, but I think we do update this after we release both. Reminder to do it :)

matibat (approved):
LGTM

pubspec.yaml
karlprieb: You're right @matibat

matibat (approved):
LGTM

PR #1073, opened 2023-05-01
PE-3604: Modularize DB migration logic

matibat:
Changes
- Implements the `resolveMigration` method.
- Utilizes the newly introduced method in the callback for  `onUpgrade`,  so the latter is now only responsible of calling `resolveMigration`.
- Unit tests the method.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/32cjvq449ggh8
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/2980n3g4hmuhg


matibat (commented):
Self review checkpoint

lib/models/database/database.dart
```diff
@@ -20,17 +23,22 @@ class Database extends _$Database {
   @override
   MigrationStrategy get migration => MigrationStrategy(
         onCreate: (Migrator m) {
+          debugPrint('Creating all tables');
```
karlprieb, line 26: use logger

karlprieb, line 26: use logger instead

lib/models/database/database.dart
```diff
@@ -20,17 +23,22 @@ class Database extends _$Database {
...
-            for (final table in allTables) {
-              await m.deleteTable(table.actualTableName);
-            }
-
-            await m.createAll();
+          CustomOnUpgrade migration = resolveMigration(from, to);
+          try {
+            await migration(allTables, m, from, to);
+          } catch (e, s) {
+            debugPrint('Database migration failed (v$from -> v$to): $e\n$s');
```
karlprieb, line 34: use logger

karlprieb (commented):
I don't get why you splitter that into so many files. They can all live together in `database.dart`. Even with all the helpers and migrations methods you created, the file will be small and easy to read.

lib/models/database/migration_strategy/resolve_migration.dart
```diff
@@ -0,0 +1,32 @@
...
+
+  if (from == 16 && to == 17) {
+    debugPrint('Migrating schema from v16 to v17');
+    return onUpgradeV16ToV17;
+  }
+
+  return _fallbackToDefault(from, to);
+}
+
+CustomOnUpgrade _fallbackToDefault(
```
karlprieb, line 23: Give it a proper name, like `resetDB`

lib/models/database/database.dart
```diff
@@ -20,17 +23,22 @@ class Database extends _$Database {
...
         onUpgrade: (Migrator m, int from, int to) async {
-          print('schema changed from $from to $to ');
-          if (from >= 1 && from < schemaVersion) {
-            // Reset the database.
-            for (final table in allTables) {
-              await m.deleteTable(table.actualTableName);
-            }
-
-            await m.createAll();
+          CustomOnUpgrade migration = resolveMigration(from, to);
```
karlprieb, line 30: In reality `resolveMigration` is so simple that we could have those 2 `if` statements and the fallback `return` here inside of `onUpgrade`

lib/models/database/migration_strategy/strategies/migration_default.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:drift/drift.dart';
+
+Future<void> onUpgradeDefault(
+  Iterable<TableInfo<Table, dynamic>> allTables,
+  Migrator m,
+  int from,
+  int to,
+) async {
+  if (from >= 1 && from < to) {
+    // Reset the database.
```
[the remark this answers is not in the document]
matibat: The log is at resolveMigration:
```dart
debugPrint(
    'WARNING: Fallbacking to default DB migration: drop and re-create all'
    ' tables - from v$from to v$to',
  );
```

lib/models/database/migration_strategy/resolve_migration.dart
matibat: This is actually a private member of the module so in this context it indeed represents we're fallback-ing to the default DB Migration Strategy. Also, this method doesn't contain any code related to the DB.

That said I would maybe rename `onUpgradeDefault` instead

lib/models/database/migration_strategy/strategies/migration_default.dart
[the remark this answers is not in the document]
karlprieb: can you move the log to this function as the db clearing happens here?

lib/models/database/database.dart
karlprieb: We just need separated functions (that can live in the same file). I don't see the complexity of testing these functions that live in the same file, in fact it will have the same tests that you already have.

lib/models/database/migration_strategy/strategies/migration_default.dart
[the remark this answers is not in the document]
matibat: Moves the log statements from resolveMigration into the migration strategy methods and use logger
7d11ba9a3b2daeae3d9a2d0e06846cb4768e384f

lib/models/database/database.dart
matibat: 3e3943569b51d3d6f0a0877b753c694e5b6c22bb

lib/models/database/database.dart
matibat: 3e3943569b51d3d6f0a0877b753c694e5b6c22bb

lib/models/database/migration_strategy/resolve_migration.dart
matibat: I removed that method at 7d11ba9a3b2daeae3d9a2d0e06846cb4768e384f. Renames the default migration method at 8bd44299a5c9905ada1bdfaa4e5484d50320b0ce.

karlprieb (approved):
Approved, but I still don't like and I think it doesn't make much sense to split this in several tiny files.
I will approve because we already spent much more time than I expected on this PR.

PR #1075, opened 2023-05-02
PE-3561: prep for turbo

javdhu:
--- Releases ---
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/568taid4jmns0
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/44s815brkt7h0

### DIFF
--- a/lib/services/turbo/payment_service.dart
+++ b/lib/services/turbo/payment_service.dart
@@ -14,24 +14,23 @@ class PaymentService {
     required this.httpClient,
   });
 
-  Future<int> getPrice({
-    required int dataItemSize,
-    required Wallet wallet,
+  Future<BigInt> getPrice({
+    required int byteSize,
   }) async {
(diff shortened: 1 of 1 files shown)


lib/services/turbo/payment_service.dart
```diff
@@ -0,0 +1,106 @@
...
+    );
+
+    if (result.data == 'User not found') {
+      return BigInt.zero;
+    }
+
+    return BigInt.parse(result.data);
+  }
+
+  Future topUp({
```
karlprieb, line 58: maybe `getPaymentSession` or `getPaymentInfo` would be a better name, what do you think?

karlprieb (approved):

PR #1076, opened 2023-05-02
PE-3580: Adds the custom_metadata field to the revision tables

matibat:
Changes
- Implements the migration strategy (from v16 to v17)
- Adds the custom_metadata column for drive_revisions, folder_revisions, and file_revisions
- Tests the DB interaction with the new column
- Bumps the DB schema version to 17

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6dd5gejkcdvv0
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/4rhjr82s9rmfg


matibat (commented):
Self-review checkpoint

matibat (commented):
Self-review checkpoint

test/models/custom_metadata_test.dart
```diff
@@ -0,0 +1,47 @@
+import 'package:ardrive/models/custom_metadata_type.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('CustomMetadata class', () {
```
karlprieb, line 5: you could add a comparison from `CustomMetadata` `toJson` with computed json and the same to `fromJson`.
Does it make sense?

lib/models/custom_metadata_type.dart
```diff
@@ -0,0 +1,39 @@
+import 'dart:collection';
+
+class CustomMetadata extends MapMixin<String, dynamic> {
```
karlprieb, line 3: why not use `json_annotation` and `json_serializable` from dart?
I don't think we're dealing with a special case here

lib/models/custom_metadata_type.dart
```diff
@@ -0,0 +1,39 @@
...
+    _data![key] = value;
+  }
+
+  @override
+  void clear() {
+    _data?.clear();
+  }
+
+  @override
+  Iterable<String> get keys => _data?.keys ?? [];
```
karlprieb, line 25: should we return an empty list when the key doesn't exist? why?

lib/models/database/migration_strategy/strategies/migration_v16_to_v17.dart
```diff
@@ -7,6 +7,28 @@ Future<void> onUpgradeV16ToV17(
...
+      (element) => element.actualTableName == 'drive_revisions',
+    );
+
+    await m.alterTable(
+      TableMigration(
+        driveRevisionsTable,
+        newColumns: [
+          GeneratedColumn<String?>(
+            'customJsonMetaData',
+            'drive_revisions',
```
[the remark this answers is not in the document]
matibat: 0478fe4bc9145531ee01dfe6b00c2a1002c8a925

lib/models/custom_metadata_type.dart
matibat: It's returning an empty array when `_data` is null, or when it doesn't have any keys. It does return the expected array of keys when there are some.

test/models/custom_metadata_test.dart
karlprieb: exactly :D So if we change `CustomMetadata` behavior we have a way to check it on this test

lib/models/custom_metadata_type.dart
matibat: I ended up removing that class as I'm not using it. I'll introduce it again if needed later on 👍

test/models/database/custom_metadata_column/custom_metadata_column_test.dart
```diff
@@ -0,0 +1,208 @@
+import 'package:ardrive/models/database/database.dart';
```
karlprieb, line 1: need to remove this file

matibat: It was actually at DriveDAO - Now the tests are passing: 1413aa6ec33cd78817e2c6a0dd3dff839279d7ec

matibat: Ohh nah, wait: this file is OK - Here I'm testing read/write operations to the new column

PR #1085, opened 2023-05-10
PE-3041: Create Snapshots out of cached Json Metadata

matibat:
## Changes
Refactors the migration strategy so now we can implement custom ones.
- Now we have better logging.
- All previous migrations will follow the same logic, no changes are expected.
- The migration from DB Schema v16 to v17 has now a custom strategy: it will introduce a new column with null values by default.

At sync, each newly synced entity (when you're already on DB Schema v17) will now write the custom metadata of it into the DB.
- Now we have better logging.
- Any entity already synced with DB Schema v16 will not be affected.

At snapshot creation, there will be a new phase just before "Computing snapshot data": "Preparing for snapshot creation".
- This stage will take care of updating any custom metadata not updated at sync time (the ones created at v16).
- This stage is not (yet) cancelable - See the tech_debt: PE-3739
- For private drives this behavior is disabled - See the tech_debt: PE-3738

Then at snapshot creation, after succeeding in updating the missing pieces of data we're gonna read the DB to re-construct the data of entities.
- So now the "Computing snapshot data" stage will use as much of the cache as possible.
- Will fallback to the network for any missing piece of data.
- For private drives this behavior is disabled - See the tech_debt: PE-3738

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/7t90tcfqhk1q8
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/03p2gvjpodjs8


matibat:
Won't do - replaced by #1109

PR #1102, opened 2023-05-17
PE-3692: feature(turbo): adds the turbo button on the profile dropdown

jdaev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/26qmis0k3vml8
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/46gqdti0ludt0


thiagocarvalhodev (changes requested):
@jdaev , the UI doesn't match the design, a few points:

- Hover effect at the first item and the two rows (the separators)
- Spacing between components
- Turbo SVG is weird.

If the component doesn't satisfy the needs of the UI, feel free to create a new one - you can do it on the ardrive-web side, since it is a specific component from ardrive app, not a core component.


https://github.com/ardriveapp/ardrive-web/assets/32248947/9039fe36-8ff1-49e2-8605-17d88fc6b956

image

thiagocarvalhodev:
@jdaev , still missing some spacing adjustments: 

- The hover effect does not good on the wallet address item.
- The padding from the Balance to Turbo
- The padding from Turbo to logout.

It could be because the dropdown uses the same height for each item. I strongly suggest implementing a new component

image

image

image

image

thiagocarvalhodev:
> image image Updated the item size for more spacing. There is no design for what the hover should look like. If you have something in mind, please add to the ticket.

Nice update!

Another small thing: the font color for the `0.0 AR` and `0.5 credits` are not the same as the design. Can you update it?

lib/components/turbo_balance_widget.dart
```diff
@@ -0,0 +1,108 @@
...
+
+class TurboBalance extends StatelessWidget {
+  const TurboBalance({
+    Key? key,
+    required this.paymentService,
+    required this.wallet,
+  }) : super(key: key);
+
+  final Wallet wallet;
+  final PaymentService paymentService;
```
thiagocarvalhodev, line 18: We're using a service directly in the UI. Please implement a Bloc (or even a ChangeNotifier) to handle the business logic.

thiagocarvalhodev (changes requested):

thiagocarvalhodev (commented):
LGTM. It's only missing the tests for the new Cubit. When done, I can approve

thiagocarvalhodev (dismissed):

thiagocarvalhodev (approved):

PR #1117, opened 2023-05-26
Mnemonic wallets

kunstmusik:
Implementation of mnemonic wallet login and generation. 

* User can login using seed phrase generated with Arweave.app
* User can generate a 12-word mnemonic wallet and user is prompted to download wallet as well as verify mnemonic via memory check before proceeding
* User can use Ardrive-generated wallet mnemonic to login into Arweave.app
* User can use Ardrive-generated wallet file to login into Arweave.app 

Uses custom arweave-wallet.js file when on web platform to generate wallet efficiently (compared to native Dart code via PointyCastle) from mnemonic.

### DIFF
--- a/lib/authentication/login/views/login_page.dart
+++ b/lib/authentication/login/views/login_page.dart
@@ -1,6 +1,4 @@
 import 'dart:convert';
-import 'dart:html';
-import 'dart:io' as io;
 
 import 'package:animations/animations.dart';
 import 'package:ardrive/authentication/ardrive_auth.dart';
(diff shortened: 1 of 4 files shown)


lib/authentication/login/views/login_page.dart
```diff
@@ -1250,3 +1294,429 @@ class MaxDeviceSizesConstrainedBox extends StatelessWidget {
...
+  }
+
+  void _onDownload() async {
+    final jsonTxt = jsonEncode(widget.wallet!.toJwk());
+
+    final ArDriveIO io = ArDriveIO();
+    final bytes = Uint8List.fromList(utf8.encode(jsonTxt));
+
+    await io.saveFile(await IOFile.fromData(bytes,
+        name: "ardrive-wallet.json", lastModifiedDate: DateTime.now()));
```
karlprieb, line 1720: maybe we can use the public address as name here, but it's something we can do later as this is just a POC.

karlprieb (approved):
Great POC. Thanks for adding this feature to the app!

thiagocarvalhodev (approved):
LGTM

PR #1118, opened 2023-05-31
PE-3040: Create snapshot out of cached GQL Nodes

matibat:


lib/utils/gql_nodes_cache.dart
```diff
@@ -0,0 +1,242 @@
...
+      ..sort(
+        (a, b) {
+          final indexRegexp = RegExp(r'.+_(\d+)$');
+          final aIndex = int.parse(indexRegexp.firstMatch(a)!.group(1)!);
+          final bIndex = int.parse(indexRegexp.firstMatch(b)!.group(1)!);
+          return aIndex.compareTo(bIndex);
+        },
+      );
+
+    // final firstKey = sortedKeysForDriveId.first;
```
matibat, line 78:
```suggestion
```

lib/utils/gql_nodes_cache.dart
```diff
@@ -0,0 +1,242 @@
...
+    final lastKey = sortedKeysForDriveId.last;
+    // final firstValue = (await _persistingCache.get(firstKey))!;
```
matibat, line 80:
```suggestion
```

lib/utils/metadata_cache.dart
matibat, line 1: TODO: do the same for get

matibat (commented):
Self-review checkpoint.

lib/blocs/sync/utils/sync_drive.dart
```diff
@@ -155,6 +158,13 @@ Stream<double> _syncDrive(
           calculatePercentageBasedOnBlockHeights() * fetchPhaseWeight;
       yield percentage;
     }
+
+    final isDeepSync = lastBlockHeight == 0;
+    if (isDeepSync) {
+      // TODO: re-visit this - any way of avoiding the conditional?
+      // FIXME: add only items 15+ blocks before current block height
```
matibat, line 165: BUMP HERE - It's not useful to track blocks that are too recent because of chain forks

lib/utils/gql_nodes_cache.dart
```diff
@@ -0,0 +1,227 @@
...
+
+  Future<bool> put(String driveId, DriveHistoryTransaction data) async {
+    if (await isFull) {
+      return false;
+    }
+
+    final nextIndex = await nextIndexForDriveId(driveId);
+    final key = '${driveId}_$nextIndex';
+
+    // FIXME: check for quota before attempting to write to cache
```
matibat, line 127: We need a workaround for this, at least for the get operation

matibat (commented):
Self-review checkpoint

matibat:
Also, don't forget to add items to the cache in the Snapshot Creation process when there's a cache miss.

lib/utils/gql_nodes_cache.dart
matibat: We've set a narrower maxEntries limit

lib/blocs/sync/utils/sync_drive.dart
matibat: d95d1229b9f38ffad16af07497c6f19cca32a75d

lib/blocs/create_snapshot/create_snapshot_cubit.dart
```diff
@@ -138,6 +150,16 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
...
+        ? Range.nullRange
+        : Range(
+            start: rangeInCache.start,
+
+            // We ignore the latest block for now, as it may be inclomplete.
+            end: rangeInCache.end - 1,
+          );
+
     // ignore: avoid_print
     print(
```
karlprieb, line 164: As you modified this file, can you change all print to logger?

lib/blocs/sync/utils/sync_drive.dart
karlprieb, line 1: change print to logger here as well

lib/blocs/sync/utils/sync_drive.dart
```diff
@@ -165,6 +168,14 @@ Stream<double> _syncDrive(
           calculatePercentageBasedOnBlockHeights() * fetchPhaseWeight;
       yield percentage;
     }
+
+    final isDeepSync = lastBlockHeight == 0;
+    final isTrustedBlock =
+        t.block != null && t.block!.height <= currentBlockHeight - 15;
+    if (isDeepSync && isTrustedBlock) {
+      // TODO: re-visit this - any way of avoiding the conditional?
```
karlprieb, line 176: I don't think there's a way to avoid that as we need to ensure that we're syncing from genesis and the block is trusted.

lib/utils/gql_nodes_cache.dart
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  Future<Range> range(DriveID driveId) async {
+    final keys = await _persistingCache.keys;
+    final keysForDriveId = keys.where((key) => key.startsWith(driveId));
+    if (keysForDriveId.isEmpty) {
+      return Range.nullRange;
+    }
+
+    final sortedKeysForDriveId = keysForDriveId.toList()
```
karlprieb, line 55: What are we sorting here? I don't think I understood

lib/utils/gql_nodes_cache.dart
```diff
@@ -0,0 +1,227 @@
...
+      ..sort(
+        (a, b) {
+          final indexRegexp = RegExp(r'.+_(\d+)$');
+          final aIndex = int.parse(indexRegexp.firstMatch(a)!.group(1)!);
+          final bIndex = int.parse(indexRegexp.firstMatch(b)!.group(1)!);
+          return aIndex.compareTo(bIndex);
+        },
+      );
```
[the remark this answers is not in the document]
karlprieb: As you use this sort more than once you could create function for that.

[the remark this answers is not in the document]
karlprieb: Why don't we just sort the filtered list? Like `final sortedKeysForDriveId = keysForDriveId.toList()..sort()`

lib/utils/gql_nodes_cache.dart
```diff
@@ -0,0 +1,227 @@
...
+
+  Future<DriveHistoryTransaction> _forceGet(String key) async {
+    logger.d('Asked to force get $key from GQL Nodes cache');
+    final maybeValue = await _persistingCache.get(key);
+    if (maybeValue == null) {
+      logger.e('Could not find $key in GQL Nodes cache');
+      throw Exception('Could not find $key in GQL Nodes cache');
+    }
+    return maybeValue;
+  }
```
karlprieb, line 152: Why do we need this? what's the difference between this and  the `get` below?

karlprieb: At least it should have the same params as `get`

lib/utils/gql_nodes_cache.dart
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    final nextIndex = _nextIndexForDriveId![driveId]! + 1;
+    return nextIndex;
+  }
+
+  Future<Map<String, int>> get currentIndexesPerDriveId async {
+    final Map<String, int> currentIndexesPerDriveId = {};
+
+    final regexp = RegExp(r'^(.+)_(\d+)$');
```
karlprieb, line 211: you can use split here that's waaaaay more performant

test/utils/gql_nodes_cache_test.dart
```diff
@@ -0,0 +1,386 @@
...
+    test('can be constructed out of a Cache Store', () async {
+      final CacheStore cacheStore = await newMemoryCacheStore();
+
+      expect(
+        await GQLNodesCache.fromCacheStore(cacheStore),
+        isInstanceOf<GQLNodesCache>(),
+      );
+    });
+
+    test('can take as many entries as the maxEntries limit', () async {
```
karlprieb, line 26: This test is not testing `maxEntries`... with a store that has `maxEntries` = 10 you should also try to add 11 to see what happens.

test/utils/gql_nodes_cache_test.dart
```diff
@@ -0,0 +1,386 @@
...
+      );
+
+      final mockData = generateMockData(1)[0];
+
+      await metadataCache.put(mockDriveId, mockData);
+
+      expect(await metadataCache.size, 1);
+    });
+
+    test('isFull indicates whether the cache is at max capacity', () async {
```
karlprieb, line 67: is `isFull` returning `false` when the store is not full? Add a test for that

karlprieb: Suggestion: create a store with maxEntries = 10. Check `isFull` before adding entrirs and at each `put` test `isFull`.

lib/utils/gql_nodes_cache.dart
```diff
@@ -0,0 +1,227 @@
...
+  Map<String, int>? _nextIndexForDriveId;
+
+  GQLNodesCache(
+    this._persistingCache, {
+    int maxEntries = defaultMaxEntries,
+  }) : _maxEntries = maxEntries;
+
+  static Future<GQLNodesCache> fromCacheStore(
+    CacheStore store, {
+    int maxEntries = defaultMaxEntries,
```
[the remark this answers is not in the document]
matibat: This field is not being used anywhere in the codebase. Just for testing purposes (to have a small cache in unit tests).

lib/utils/gql_nodes_cache.dart
matibat: I made it to be in the private API of the class, so the parameters shouldn't matter (it takes the cache key directly because it's a low-level method). This method is being used ONLY when WE'RE SURE there's a certain item in the cache and we want to read it again, that way it will be non-nullable.

test/utils/gql_nodes_cache_test.dart
matibat: Nice catch! Thanks

lib/blocs/create_snapshot/create_snapshot_cubit.dart
matibat: 350e8cda04f0b2b9742b94a519bc593f12055586

lib/blocs/sync/utils/sync_drive.dart
matibat: Should we get rid of `lib/blocs/sync/utils/log_sync.dart` as well?

```dart
part of 'package:ardrive/blocs/sync/sync_cubit.dart';

void logSync(String message) {
  // ignore: avoid_print
  print('sync: $message');
}

void logSyncError(Object error, StackTrace stackTrace) {
  logSync(
    '''
    an error occurred during sync. \n
    error: ${error.toString()} \n
    stacktrace: ${stackTrace.toString()}
    ''',
  );
}
```

karlprieb: If the last `logSync` usage are on these files you can delete it, if not let it to another PR.

lib/blocs/sync/utils/sync_drive.dart
karlprieb: Can you remove the comment?

lib/blocs/sync/utils/sync_drive.dart
matibat: Removes prints: 669ef6c91f1ac1b04d87d3b498476adaa1665715

lib/utils/gql_nodes_cache.dart
karlprieb: What's the difference between forceGet and get?

lib/blocs/sync/utils/sync_drive.dart
matibat: It's part of sync_cubit. I will end up editing some extra files, but everything is related

lib/utils/gql_nodes_cache.dart
karlprieb: drive-id uses uuid... they don't have underscores

lib/blocs/sync/utils/sync_drive.dart
matibat: Gets rid of logSync: f86941edb19006ef582ba11fe29f2d51c1c9f614

lib/blocs/sync/utils/sync_drive.dart
matibat: 7569eb93927d3236a171fa867584c66c409953c0

lib/utils/gql_nodes_cache.dart
matibat: Uhh Drive IDs don't have underscores.

```dart
final driveId = _uuid.v4();
```

matibat: There's a catch tho - https://stackoverflow.com/a/61435170

matibat: I did this: a9cf9acfb6707033dfdb5a55b6055fddd6edc837

matibat: We discussed about the possibility of using [compareNatural](https://pub.dev/documentation/collection/latest/collection/compareNatural.html) but we're OK with the current approach

matibat:
Now unit testing the class for the cache key - 0b2d6a0198d7f15bfbb30d1529b0db6fab225dbe

test/utils/gql_nodes_cache_test.dart
```diff
@@ -0,0 +1,388 @@
...
+      );
+
+      expect(await metadataCache.isFull, false);
+
+      final mockData = generateMockData(10);
+
+      for (int i = 0; i < mockData.length; i++) {
+        final wasPut = await metadataCache.put(mockDriveId, mockData[i]);
+        expect(wasPut, true);
+      }
```
karlprieb, line 80:
```suggestion
      for (int i = 0; i < mockData.length; i++) {
        final wasPut = await metadataCache.put(mockDriveId, mockData[i]);
        if (i < mockData.length) {
        	expect(metadataCache.isFull, false);
        }
        expect(wasPut, true);
      }
```
I would add this `isFull` expect as well

matibat: Applied - f461436e5d5f585b4cba58b7de53b045a4b60528

matibat:
Missing to wipe cache on logout!

lib/utils/gql_nodes_cache.dart
matibat: I did this: 4d9a680fb9561eee3abbba46d85f4b2386308836

karlprieb (approved):

PR #1130, opened 2023-06-07
PE-3977: Before-release preparations

matibat:
Changes
- Android release notes (next prod deploy number is 45)
- App version bump

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/1fue73b2crtdo
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/14oalufdvedho

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -3,7 +3,7 @@ description: Secure, permanent storage
 
 publish_to: 'none'
 
-version: 2.0.2
+version: 2.2.0
 
 environment:
   sdk: '>=2.18.5 <3.0.0'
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Secure, permanent storage
 
 publish_to: 'none'
 
-version: 2.0.2
+version: 2.2.0
```
matibat, line 6: The version at master seems to be v2.1.0. I don't understand why there is v2.0.2 in dev 😵‍💫 

Also, there's no [diff](https://github.com/ardriveapp/ardrive-web/blob/master/pubspec.yaml) in between master and dev - suspicious 🤨

thiagocarvalhodev: It was due this PR: #1102

thiagocarvalhodev (approved):
LGTM

PR #1143, opened 2023-06-13
PE-3954: refactor(snapshot_item_to_be_created.dart): replace asyncMap with StreamController to batch calls

jdaev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/1no0bu37kutpg
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/0tju1g4mas970

### DIFF
--- a/lib/blocs/sync/utils/sync_drive.dart
+++ b/lib/blocs/sync/utils/sync_drive.dart
@@ -46,22 +46,17 @@ Stream<double> _syncDrive(
 
   if (configService.config.enableSyncFromSnapshot) {
     logger.i('Syncing from snapshot');
-    try {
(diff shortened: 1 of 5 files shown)


matibat:
The code seems good to me. Before approving I wanna ensure that we're preserving the order of the TXs, so I'll give it a try

matibat (changes requested):
There was a missing requirement in the ticket: preserve the same order as GQL.

karlprieb (dismissed):

matibat (dismissed):
Tests are now passing. Thanks! <3

matibat (approved):

PR #1157, opened 2023-06-16
PE-4034: unable to open turbo for users without previous credit history

thiagocarvalhodev:


lib/turbo/turbo.dart
```diff
@@ -233,7 +233,18 @@ class TurboBalanceRetriever {
...
 
   Future<BigInt> getBalance(Wallet wallet) async {
-    return paymentService.getBalance(wallet: wallet);
+    try {
+      return paymentService.getBalance(wallet: wallet);
+    } catch (e, s) {
+      logger.e('Error getting balance', e, s);
+
+      if (e is TurboUserNotFound) {
+        logger.e('User not found, returning 0 balance');
```
karlprieb, line 242:
```suggestion
        logger.w('User not found, returning 0 balance');
```

karlprieb (approved):

PR #1159, opened 2023-06-19
PE-2932: for testing corrupted snapshot case

javdhu:


matibat:
Merged the base in order to lower the diff - dev was merged previously (not the base).

matibat:
The failing test here is expected as we're forcing it to fail for testing purposes. The original branch is not failing, as expected. All good :)

PR #1168, opened 2023-06-23
refactor: use logger, update deprecated widgets

javdhu:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6cr5n7fdovr40
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/7d3dsglrqn4l8

### DIFF
--- a/lib/utils/logger/logger.dart
+++ b/lib/utils/logger/logger.dart
@@ -7,7 +7,7 @@ final logger = Logger(
     lineLength: 120,
     colors: true,
     printEmojis: true,
-    printTime: false,
+    printTime: true,
     noBoxingByDefault: true,
   ),
 );


lib/utils/logger/logger.dart
```diff
@@ -1,3 +1,13 @@
...
 
-final logger = Logger();
+final logger = Logger(
+  printer: PrettyPrinter(
+    methodCount: 0,
+    errorMethodCount: 8,
+    lineLength: 120,
+    colors: true,
+    printEmojis: true,
+    printTime: false,
```
karlprieb, line 10: I think it make sense to set `true` here

karlprieb (approved):

PR #1169, opened 2023-06-23
PE-4003: Implementing Balance and Price Estimate View for Turbo Top-up - Only USD - MOBILE

matibat:
[WIP]

Changes
- Replaces Row with Column - This makes the modal to render even that it overflows.

--- Releases ---
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/4ddtvbl620aoo
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/176thdderqbt0


lib/components/top_up_dialog.dart
matibat, line 1: This is the relevant change: Replaces `Column` with `Row`

lib/components/profile_card.dart
```diff
@@ -107,17 +107,14 @@ class _ProfileCardState extends State<ProfileCard> {
           ),
           _buildBalanceRow(context, state),
           if (context.read<PaymentService>().useTurboPayment) ...[
-            Padding(
```
matibat, line 110: Unnecessary change / can be reverted

lib/components/turbo_balance_widget.dart
```diff
@@ -64,7 +64,7 @@ class _TurboBalanceState extends State<TurboBalance> {
     logger.d('Building turbo balance widget');
 
     return Padding(
-      padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 4),
+      padding: const EdgeInsets.only(left: 16, right: 16, top: 24, bottom: 4),
```
matibat, line 67: Unnecessary change / can be reverted

PR #1190, opened 2023-07-05
fix(stripe key): use stripe prod stripe key

thiagocarvalhodev:


vilenarios (approved):
This looks good.  Sets staging to use the production stripe api key.

karlprieb (approved):

PR #1191, opened 2023-07-05
fix(config service): load based on the env

thiagocarvalhodev:


lib/pages/app_router_delegate.dart
karlprieb, line 1: Not sure about this change

karlprieb (approved):

PR #1195, opened 2023-07-07
PE-4146: disable automatic sync when user is on the topup modal

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/7jvrjtlnaap48
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/1emqn45sgg96o

### DIFF
--- a/lib/blocs/sync/sync_cubit.dart
+++ b/lib/blocs/sync/sync_cubit.dart
@@ -5,6 +5,7 @@ import 'package:ardrive/blocs/activity/activity_cubit.dart';
 import 'package:ardrive/blocs/blocs.dart';
 import 'package:ardrive/blocs/constants.dart';
 import 'package:ardrive/blocs/sync/ghost_folder.dart';
+import 'package:ardrive/core/activity_tracker.dart';
 import 'package:ardrive/entities/entities.dart';
 import 'package:ardrive/entities/string_types.dart';
 import 'package:ardrive/models/models.dart';
@@ -59,6 +60,7 @@ class SyncCubit extends Cubit<SyncState> {
   final Database _db;
   final TabVisibilitySingleton _tabVisibility;
   final ConfigService _configService;
+  final ActivityTracker _activityTracker;
 
   StreamSubscription? _restartOnFocusStreamSubscription;
   StreamSubscription? _restartArConnectOnFocusStreamSubscription;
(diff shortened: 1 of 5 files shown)


lib/blocs/sync/sync_cubit.dart
```diff
@@ -183,6 +187,11 @@ class SyncCubit extends Cubit<SyncState> {
   var ghostFolders = <FolderID, GhostFolder>{};
 
   Future<void> startSync({bool syncDeep = false}) async {
+    if (!_activityTracker.isSyncAllowed) {
+      logger.i('Activity tracker is not allowing sync');
```
karlprieb, line 191:
```suggestion
      logger.d('Activity tracker is not allowing sync');
```

karlprieb (dismissed):
I'll approve so we can move one with Turbo, but I don't like the implementation. I was expecting something more generic that can track activities and decide if sync is allowed or not.
It's a good start, but too Turbo specific for now

lib/core/activity_tracker.dart
```diff
@@ -0,0 +1,14 @@
+class ActivityTracker {
+  bool _isToppingUp = false;
+  // getters
+  bool get isToppingUp => _isToppingUp;
+
+  // setters
+  void setToppingUp(bool value) {
```
matibat, line 7: The setter should be defined as so:
```suggestion
  set setToppingUp(bool value) {
```

lib/turbo/topup/views/topup_modal.dart
```diff
@@ -54,6 +55,8 @@ void showTurboModal(BuildContext context, {Function()? onSuccess}) {
 
   initializeStripe(context.read<ConfigService>().config);
 
+  context.read<ActivityTracker>().setToppingUp(true);
```
matibat, line 58: Will change if you replace the method approach with the setter approach
```suggestion
  context.read<ActivityTracker>().setToppingUp = true;
```

lib/turbo/topup/views/topup_modal.dart
```diff
@@ -82,6 +85,8 @@ void showTurboModal(BuildContext context, {Function()? onSuccess}) {
       onSuccess?.call();
     }
 
+    context.read<ActivityTracker>().setToppingUp(false);
```
matibat, line 88: Will change if you replace the method approach to the setter approach
```suggestion
    context.read<ActivityTracker>().setToppingUp = false;
```

matibat (commented):
Left some comments. It's also lacking unit tests, and maybe bloc tests for sync (not sure where)

karlprieb (approved):

PR #1215, opened 2023-07-13
PE-4173: uploading process stop when switching tabs after free file upload via ar connect

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/7gfqift7gpueo
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/5dtvhu90j4j7g


matibat:
Tests are broken - There I propose a possible fix: #1223

lib/core/arconnect/safe_arconnect_action.dart
matibat, line 1: Let's get rid of all logging messages not related to "safe ArConnect actions"

matibat (changes requested):
Reviewed only the safeArConnectAction method. Also, there's a PR with a potential fix for the tests (fixes injection, removes calls to the JS environment).

I liked the idea of abstracting the code into a separate method, and also having a specific signer for arconnect (that way we don't have to check for arconnect from inside the new method). Looking good so far

matibat:
I'm gonna take care of this PR now, with the permission of @thiagocarvalhodev.

I'm gonna
1. Address my comment;
2. review the changes I haven't,
  a. and address any comments I make -if any-;
4. make a quick test to ensure I didn't change the behavior before merging my changes;
5. approve, and let @thiagocarvalhodev know.

lib/core/arconnect/safe_arconnect_action.dart
matibat: 717920064c112d8234f45d2e690b05144f2ab456

matibat (commented):
Finished reviewing. I'll approve after a quick test

lib/blocs/profile/profile_cubit.dart
```diff
@@ -89,13 +93,31 @@ class ProfileCubit extends Cubit<ProfileState> {
...
-        return true;
-      }
-      final currentPublicKey = await arconnect.getPublicKey();
-      final savedPublicKey = profile.walletPublicKey;
-      if (currentPublicKey != savedPublicKey) {
-        return true;
+      try {
+        if (!(await arconnect.checkPermissions())) {
+          logger.i('ArConnect permissions changed');
+          throw Exception('ArConnect permissions changed');
```
matibat, line 99: "no check arconnect permissions" = "true, wallet mismatch", right?

lib/blocs/profile/profile_cubit.dart
matibat, line 1: IMO we shouldn't be touching this method, but the callers of it: that way we would be wrapping the call to this method in the safeArConnectAction

matibat (commented):
Left one comment - I need to check this with the team because I'm not sure

lib/blocs/profile/profile_cubit.dart
matibat: This change was intended, in order to re-try it on the catch part

lib/services/turbo/upload_service.dart
```diff
@@ -59,10 +63,22 @@ class TurboUploadService {
...
-      nonce: nonce,
-      wallet: wallet,
+    final publicKey = await safeArConnectAction<String>(
+      _tabVisibility,
+      (_) async {
+        logger.d('Getting public key with safe ArConnect action');
+        return wallet.getOwner();
+      },
+    );
+    final signature = await safeArConnectAction<String>(
```
matibat, line 73: It's calling the method even for non-arconnect wallets

matibat (commented):
Re-reviewed

PR #1217, opened 2023-07-13
PE-4192: disable the Add button for native android and ios

matibat:
Changes
- Disables the button for native apps, so it's only enabled on browser


thiagocarvalhodev (dismissed):
LGTM

karlprieb:
The ticket said to disable the button and that's what you did, but in fact we want to hide it

matibat:
Change in requirements. Now hiding the button instead of disabling it 👍

karlprieb (dismissed):

karlprieb (approved):

PR #1218, opened 2023-07-13
PE-4203: fix(turbo upload sizes): fix logic to display turbo sizes

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/0bravu0cj5dpo
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/5s5fsacidooi8

### DIFF
--- a/lib/core/upload/uploader.dart
+++ b/lib/core/upload/uploader.dart
@@ -332,9 +332,11 @@ class UploadPaymentEvaluator {
 
     BigInt turboBalance;
 
-    bool isTurboAvailable = _appConfig.useTurboUpload;
+    /// Even if this feature flag is off, it will be possible to upload using turbo
+    /// for free files
+    bool isTurboAvailableToUploadAllFiles = _appConfig.useTurboUpload;
 
-    if (isTurboAvailable) {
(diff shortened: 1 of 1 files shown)


lib/core/upload/uploader.dart
```diff
@@ -363,7 +365,7 @@ class UploadPaymentEvaluator {
     int turboBundleSizes = 0;
 
     /// Calculate the upload with Turbo if possible
-    if (isTurboAvailable && isUploadEligibleToTurbo) {
+    if (isUploadEligibleToTurbo) {
```
matibat, line 368: I don't see `isUploadEligibleToTurbo` changing here, but we're removing this conditional. Is this intended?

thiagocarvalhodev: Yes! We need because even if the turbo is not available (for all uploads) we can upload free files.

matibat (approved):
LGTM

PR #1223, opened 2023-07-14
PE-4173: inject tab visibility

matibat:
TODO - we have to cleanup html_utils.dart in order to avoid this kind of confusion: we cannot use the JS implementations directly in the code because they will throw in non-JS platforms.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6dq855tgj7v1g
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/086ubcdd5qfbg

### DIFF
--- a/lib/core/upload/bundle_signer.dart
+++ b/lib/core/upload/bundle_signer.dart
@@ -75,17 +75,21 @@ class ArweaveBundleTransactionSigner implements BundleTransactionSigner {
 }
 
 class SafeArConnectSigner<T> extends BundleSigner<T> {
-  final TabVisibilitySingleton tabVisibilitySingleton =
-      TabVisibilitySingleton();
+  final TabVisibilitySingleton _tabVisibilitySingleton;
 
   final BundleSigner bundleSigner;
 
-  SafeArConnectSigner(this.bundleSigner);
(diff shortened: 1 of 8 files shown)


lib/core/upload/bundle_signer.dart
```diff
@@ -75,17 +75,21 @@ class ArweaveBundleTransactionSigner implements BundleTransactionSigner {
...
-  final TabVisibilitySingleton tabVisibilitySingleton =
-      TabVisibilitySingleton();
+  final TabVisibilitySingleton _tabVisibilitySingleton;
 
   final BundleSigner bundleSigner;
 
-  SafeArConnectSigner(this.bundleSigner);
+  SafeArConnectSigner(this.bundleSigner,
+      {TabVisibilitySingleton? tabVisibility})
+      : _tabVisibilitySingleton = tabVisibility ?? TabVisibilitySingleton();
```
matibat, line 84: We may end up having to make it required if we change the unit tests in the future

matibat:
@thiagocarvalhodev gave me permission to merge these changes :)

PR #1225, opened 2023-07-14
feat(topup dialog): remove leading zeroes from the custom amount

matibat:
Suggestion to avoid typing as many zeros as they want in the Custom Amount field.

### Before
https://github.com/ardriveapp/ardrive-web/assets/22564541/f39b67fa-45eb-4d0b-8d0a-2ddc6cc8f490

### After

https://github.com/ardriveapp/ardrive-web/assets/22564541/d1400301-f99c-4d86-a11d-389c6bfa4266


matibat:
Close because of hard-to-fix conflict

PR #1249, opened 2023-07-20
PE-3985: Create new pin modal

matibat:
**IF YOU ARE HERE WANTING TO TRY OUT ArFS PINS** ~then try it at your own risk: I suggest using a  testing wallet. You can also wait until the feature gets QAed. BUT, if you're still interested in testing it~ then here you have a File-Id to one of my favorite artworks: `f724d2ee-a68a-4a1b-a738-65089355c3d2`. 

**Changes**
1. Adds an option to the `New` menu: New file pin
  a. Please don't mind the ArConnect icon :b
2. There you'll have a new modal that asks you for an ID and a name...
  a. If you provide a File-ID, then the Name field will be auto-populated with the name of the file
  b. If you provide a Transaction ID, then you'll have to be creative by choosing a wise name

**Disclaimer -** 
- This feature is enabled for PUBLIC drives only (for both the file to pin and the drive where to pin it at).
- ~There's no error feedback.~

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/2hpa8aqkphu68
iOS release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:ios:06ea33a95a2e0d42ffce07/releases/6jl5866t5ke28


lib/blocs/pin_file/pin_file_state.dart
```diff
@@ -0,0 +1,165 @@
...
+    this.maybeLastModified,
+    required this.dateCreated,
+    required this.size,
+    required this.dataTxId,
+    required this.pinnedDataOwnerAddress,
+  }) : super(
+          id: id,
+          name: name,
+          nameValidation: nameValidation,
+          idValidation: IdValidationResult.validEntityId,
```
matibat, line 97: It's not affecting the logic RN, but let's not assume it's a "valid entity id" when it can be a transaction id

lib/components/new_button/new_button.dart
```diff
@@ -342,6 +343,17 @@ class NewButton extends StatelessWidget {
...
             icon: ArDriveIcons.iconCreateSnapshot(size: defaultIconSize),
           ),
+        if (context.read<ConfigService>().config.enablePins &&
+            driveDetailState is DriveDetailLoadSuccess &&
+            drive != null &&
+            drive?.privacy == 'public')
+          _buildDriveDropdownItem(
+            name: 'Pin a file',
+            // TODO: put the actial icon
+            icon: ArDriveIcons.arconnectIcon1(size: defaultIconSize),
```
matibat, line 353: TODO: put the icon

lib/components/pin_file_dialog.dart
```diff
@@ -0,0 +1,172 @@
...
+  Widget build(BuildContext context) {
+    final pinFileBloc = context.read<PinFileBloc>();
+
+    return BlocConsumer<PinFileBloc, PinFileState>(
+      listener: (context, state) {
+        logger.d('PinFileBloc state: $state');
+        if (state is PinFileAbort ||
+            state is PinFileSuccess ||
+            state is PinFileError) {
+          // FIXME: give some feedback on error
```
matibat, line 68: Tech debt: error modal

lib/components/pin_file_dialog.dart
```diff
@@ -0,0 +1,172 @@
...
+                      final validation = pinFileBloc.validateName(p0);
+                      if (validation == NameValidationResult.invalid) {
+                        return 'Name is invalid';
+                      }
+                    }
+                    return null;
+                  },
+                  controller: pinFileBloc.nameTextController,
+                ),
+                if (customErrorMessage != null)
```
matibat, line 139: Tech debt: design for these messages

lib/components/pin_file_dialog.dart
```diff
@@ -0,0 +1,172 @@
...
+                  isFieldRequired: true,
+                  onChanged: (value) {
+                    pinFileBloc.add(
+                      FieldsChanged(id: value, name: state.name),
+                    );
+                  },
+                  validator: (p0) {
+                    if (p0 != null) {
+                      final validation = pinFileBloc.validateId(p0);
+                      if (validation == IdValidationResult.invalid) {
```
matibat, line 111: What about required? Same for the name field

matibat: Also, it's sad that I can't set a custom error from the state

lib/components/pin_file_dialog.dart
```diff
@@ -0,0 +1,172 @@
...
+          actions: [
+            ModalAction(
+              action: () => pinFileBloc.add(const PinFileCancel()),
+              title: 'Cancel',
+            ),
+            ModalAction(
+              action: () => pinFileBloc.add(const PinFileSubmit()),
+              title: 'Create',
+              // FIXME: "isEnabled"
+              isEnable: state is PinFileFieldsValid &&
```
matibat, line 164: Nit and out of scope

lib/services/config/app_config.dart
```diff
@@ -32,6 +33,7 @@ class AppConfig {
     this.enableSyncFromSnapshot = true,
     this.enableSeedPhraseLogin = true,
     required this.stripePublishableKey,
+    this.enablePins = false,
```
matibat, line 36: Maybe it's because this line that the dev tools are marking it as disabled by default, even that I've set the JSON config?

lib/utils/debouncer.dart
matibat, line 1: Yet unused, should I remove it?

test/blocs/pin_file_bloc_test.dart
```diff
@@ -0,0 +1,419 @@
+// Unit tests for the PinFileBloc class.
+
```
matibat, line 2:
```suggestion
```

matibat (commented):
Self-review checkpoint.

lib/blocs/pin_file/file_id_resolver.dart
```diff
@@ -0,0 +1,143 @@
...
+
+    return fileInfo;
+  }
+
+  @override
+  Future<FileInfo> requestForTransactionId(String dataTxId) async {
+    final hostname =
+        (configService.config.defaultArweaveGatewayUrl!).split('://')[1];
+
+    final uri = Uri.https(
```
thiagocarvalhodev, line 65: Use the Uri.parse instead of Uri.https:

```dart
Uri.parse('${configService.config.defaultArweaveGatewayUrl!}/$dataTxId')
```

lib/blocs/pin_file/file_id_resolver.dart
```diff
@@ -0,0 +1,143 @@
...
+
+  const NetworkFileIdResolver({
+    required this.arweave,
+    required this.httpClient,
+    required this.configService,
+  });
+
+  @override
+  Future<FileInfo> requestForFileId(String fileId) async {
+    late FileEntity? fileEntity;
```
thiagocarvalhodev, line 18: You can remove the `late` keyword as this property can be null

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+            rethrow;
+          }
+        }
+      }
+
+      if (!syncValidationResult.isIdValid) {
+        if (state is PinFileFieldsValidationError) {
+          final stateAsPinFileFieldsValidationError =
+              state as PinFileFieldsValidationError;
+          emit(PinFileFieldsValidationError(
```
thiagocarvalhodev, line 140: Consider implementing a private method of each `emit` method. It will improve the readability

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+              isArFsEntityValid:
+                  stateAsPinFileFieldsValidationError.isArFsEntityValid,
+              isArFsEntityPublic:
+                  stateAsPinFileFieldsValidationError.isArFsEntityPublic,
+              doesDataTransactionExist:
+                  stateAsPinFileFieldsValidationError.doesDataTransactionExist,
+            ),
+          );
+        } else {
+          logger.d('Unexpected state $state');
```
thiagocarvalhodev, line 240: What's the feedback to the user if we reach this point? Consider adding a generic error state, then show a generic message.

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+        _arweave = arweave,
+        _driveDao = driveDao,
+        _turboUploadService = turboUploadService,
+        _profileCubit = profileCubit,
+        _driveId = driveID,
+        _parentFolderId = parentFolderId,
+        super(const PinFileInitial()) {
+    nameTextController.text = '';
+
+    on<FieldsChanged>((event, emit) async {
```
thiagocarvalhodev, line 50: I suggest implementing a private method for each event handle. You can see an example on the `login_bloc.dart`

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+          await _turboUploadService.postDataItem(
+            dataItem: fileDataItem,
+            wallet: profileState.wallet,
+          );
+          newFileEntity.txId = fileDataItem.id;
+        } else {
+          final fileDataItem = await _arweave.prepareEntityTx(
+            newFileEntity,
+            profileState.wallet,
+            null, // TODO: key
```
thiagocarvalhodev, line 303: Do we have a ticket to handle private drives?

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+          id: state.id,
+          name: state.name,
+          nameValidation: state.nameValidation,
+          idValidation: state.idValidation,
+        ));
+      });
+    });
+  }
+
+  Future<bool> doesNameConflicts(String name) async {
```
thiagocarvalhodev, line 337: Consider changing this method to be private

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+        name: name,
+        driveId: _driveId,
+        parentFolderId: _parentFolderId,
+      );
+
+      logger.d('entityWithSameNameExists: $entityWithSameNameExists');
+
+      return entityWithSameNameExists;
+    } catch (err) {
+      // TODO: do somethin'
```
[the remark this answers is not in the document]
thiagocarvalhodev: If we can't check if an entity with the same doesn't exists, we should show a feedback error to the user. Let's add it as a tech debt.

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+    if (idValidation == IdValidationResult.validTransactionId) {
+      resolveFuture = _fileIdResolver.requestForTransactionId(id);
+    } else {
+      resolveFuture = _fileIdResolver.requestForFileId(id);
+    }
+
+    return resolveFuture;
+  }
+
+  NameValidationResult validateName(String value) {
```
thiagocarvalhodev, line 392: I really liked your approach to handling the name validation. We have other methods that does the same thing. 
Please create a new ticket to replace the already implemented methods with this one.

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+      return NameValidationResult.required;
+    } else if (!nameRegex.hasMatch(value) ||
+        !trimTrailingRegex.hasMatch(value)) {
+      return NameValidationResult.invalid;
+    }
+
+    return NameValidationResult.valid;
+  }
+
+  IdValidationResult validateId(String value) {
```
thiagocarvalhodev, line 406: The same for my previous comment.

lib/blocs/pin_file/types.dart
```diff
@@ -0,0 +1,39 @@
...
+}
+
+enum IdValidationResult {
+  required,
+  invalid,
+  validEntityId,
+  validTransactionId,
+}
+
+class FileInfo {
```
thiagocarvalhodev, line 17: `FileInfo` is pretty generic. Consider changing the name of this class

lib/components/pin_file_dialog.dart
thiagocarvalhodev: We must show at least generic feedback to the user here.

lib/components/pin_file_dialog.dart
thiagocarvalhodev: >Also, it's sad that I can't set a custom error from the state

Can you explain it?

lib/services/config/config_fetcher.dart
```diff
@@ -31,10 +31,12 @@ class ConfigFetcher {
     final gatewayUrl = localStore.getString('arweaveGatewayUrl');
     final enableQuickSyncAuthoring =
         localStore.getBool('enableQuickSyncAuthoring');
+    final enablePins = localStore.getBool('enablePins');
```
thiagocarvalhodev, line 34: `enablePins` is defined on the env config file. We don't need to update it here.

lib/utils/debouncer.dart
thiagocarvalhodev: Yes!

lib/utils/debouncer.dart
```diff
@@ -0,0 +1,19 @@
+import 'dart:async';
+
+class Debouncer {
```
thiagocarvalhodev, line 3: 🚀

lib/services/config/app_config.dart
thiagocarvalhodev: If the `enablePins` is provided, it will use the value passed to this constructor

lib/components/pin_file_dialog.dart
thiagocarvalhodev: Do we have the designs? If not, please reach Alison for provide those

thiagocarvalhodev (changes requested):
Looking pretty good! Some refactors are not required, but we need:
- Update the translations
- Add error feedback
- Update the pin icon

lib/components/pin_file_dialog.dart
matibat: What I mean here is: I can't set a "custom validation error" to the text field, so instead, I had to put a text at the bottom as if it were an error message, but it's actually a simple Text element

matibat: BTW we decided to put this into the scope of "Add Error Handling and Feedback to the User"

lib/components/pin_file_dialog.dart
matibat: See this comment: https://github.com/ardriveapp/ardrive-web/pull/1249#discussion_r1277724976

TL/DR - The ArDriveTextField component doesn't have a "validationErrorMessage" property, so we can't display one. We can just run an input validation.

lib/blocs/pin_file/pin_file_bloc.dart
matibat: Sorry, I didn't get it. What would that private method do?

lib/blocs/pin_file/pin_file_bloc.dart
matibat: I actually copy/pasted them, and refactored them to be pure & de-coupled methods. The other methods are coupled to the UI because they depend on the context for localizations. What I did instead is to return an enum, so the UI can decide what to display based on the enum value.

So yes, it would be great to replace them with this more generic one :)

lib/services/config/config_fetcher.dart
matibat: So then this is the line that sets the value to false even when it is set to true in the config. Thanks!

lib/utils/debouncer.dart
matibat: ddc789415d2430d6bfd40c456f272b5cb816c11b

lib/services/config/config_fetcher.dart
matibat: bf56f33f8e8fb845a32499893a40f01d3e9f8004

lib/blocs/pin_file/pin_file_bloc.dart
matibat: 0a0fd8bb0b5c981b378d5fb69adc27d795bbef3c

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+      idValidation == IdValidationResult.validEntityId ||
+      idValidation == IdValidationResult.validTransactionId;
+
+  bool get isValid => isNameValid && isIdValid;
+
+  @override
+  String toString() {
+    return 'SynchronousValidationResult { nameValidation: $nameValidation, '
+        'idValidation: $idValidation, isValid: $isValid }';
+  }
```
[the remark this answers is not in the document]
matibat: 76113eb4ac649b5fbfffddac4dbfa589120d1c56

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+    } else if (transactionIdHasMatch) {
+      return IdValidationResult.validTransactionId;
+    } else {
+      return IdValidationResult.invalid;
+    }
+  }
+
+  @override
+  Future<void> close() {
+    // idTextController.dispose();
```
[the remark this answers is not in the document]
matibat: 76113eb4ac649b5fbfffddac4dbfa589120d1c56

lib/blocs/pin_file/pin_file_state.dart
```diff
@@ -0,0 +1,165 @@
...
+          idValidation: IdValidationResult.required,
+          nameValidation: NameValidationResult.required,
+        );
+
+  @override
+  List<Object> get props => [];
+}
+
+class PinFileFieldsValidationError extends PinFileState {
+  // network check
```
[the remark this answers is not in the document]
matibat: b9924b726f56f31db14c7c5d6f3aea73f601bb1b

lib/blocs/pin_file/pin_file_state.dart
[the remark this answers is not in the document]
matibat: DrivePrivacy: 147dd8d81b331266b6292075a3be7bab4852d1ea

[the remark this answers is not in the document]
matibat: Transaction ID and File ID types: 83fb9f426093cb4914e46929be6f90346c1232d1

lib/blocs/pin_file/pin_file_state.dart
matibat: ba3178616fe79c9ff176b4056fd6744b60553b03

lib/components/pin_file_dialog.dart
matibat: Required field validation: 6e4df16e7d432b2d88da9cb8e069964550e0d901

test/blocs/pin_file_bloc_test.dart
matibat: 3bfaca75f86b42f6ef276f60060b34163f220307

lib/blocs/pin_file/file_id_resolver.dart
matibat: f9b6a7f9f4a8c16052ebd192050d3741f01a7ac1

lib/blocs/pin_file/file_id_resolver.dart
matibat: 448ecbc87695e923f4a8439fbb166e20fe29ff63

lib/blocs/pin_file/types.dart
matibat: If I made it private I couldn't access it from the unit tests, so I renamed to `ResolveIdResult`: b93318fe53d8e49f0d02846e4546e6f8153250e0

matibat:
Addressed the most relevant and non-tech-debt-related comments. Will now setup localization

lib/blocs/pin_file/pin_file_bloc.dart
thiagocarvalhodev: For example:

```dart
emitPinFileFieldsValidationError() 
```

which emits the state inside

matibat:
Localizations in place!

lib/l10n/app_en.arb
```diff
@@ -310,7 +310,7 @@
   "@create": {
     "description": "Create an drive"
   },
-  "createAndConfirmPassword": "Please create and confirm your password. Your password cannot be changed or retrieved, so please keep this safe.",
+  "createAndConfirmPassword": "Please create and confirm your password.",
```
matibat, line 313: It seems someone edited this locally but not in Localizely

matibat:
Re-enabled turbo uploads!

I tested them again and they are working just fine. My guess is that what happened is that a chain fork made my files go away when testing yesterday.

matibat:
Also, I realized that in the case you provide a File-Id, the owner it was setting is the owner of the File METADATA (which can be a pin), and not the owner of the actual DATA. Here it is the fix: [7417978](https://github.com/ardriveapp/ardrive-web/pull/1249/commits/7417978be2cc124d93c233d20de363d361639b78)

lib/components/pin_file_dialog.dart
thiagocarvalhodev: The input validation returns the text with the validation message and displays it at the bottom of the ArDriveTextField. If it is not what is happening, so it's an error on the component.

lib/l10n/app_en.arb
thiagocarvalhodev: It was edited on localizelly but after you start this feature, so for you it was the old text

lib/blocs/pin_file/file_id_resolver.dart
```diff
@@ -0,0 +1,164 @@
...
+    required this.id,
+    required this.cancelled,
+    required this.networkError,
+    required this.isArFsEntityValid,
+    required this.isArFsEntityPublic,
+    required this.doesDataTransactionExist,
+  });
+}
+
+class OwnerAndPrivacy {
```
thiagocarvalhodev, line 156: This class name is very generic. One good question is: owner and privacy of what? In this case, an transaction.
Consider: `TransationOwnerAndPrivacy`.

Or, we can make this class only visible inside this file, so the reader know exactly the context and won't get confused seeing this class in another context.

lib/blocs/pin_file/pin_file_bloc.dart
thiagocarvalhodev: That's a pretty good solution. Don't forget to create a tech debt to update this!

lib/blocs/pin_file/pin_file_bloc.dart
thiagocarvalhodev: @matibat , consider doing the refactor I suggested

thiagocarvalhodev (commented):
Left just a few more comments.

lib/components/pin_file_dialog.dart
matibat: Yes. That's the only way to make a text field display an error: it's triggered when the user manually edits the field. There's no way of setting a custom error message that came from the state, let's say.

In my case, I cannot set the errors that came from the network, such as for non-existing transactions.

That's why I'm displaying the error message at the bottom.

lib/blocs/pin_file/file_id_resolver.dart
matibat: I made it be private: de31ebb864b4b801e8d3dd4a3eb99bd105ca5444

IMO it's not the "owner and privacy of a transaction" because transactions are always public, privacy is an ArFS concept. The class has a generic name because it is actually a generic pair of owner address and privacy: it can be the owner and privacy of a drive, a folder, a file, or a snapshot.

lib/blocs/pin_file/pin_file_bloc.dart
matibat: Here: ca94bd75218d11780a1f24013ce73e4d1e252256

lib/blocs/pin_file/pin_file_bloc.dart
```diff
@@ -0,0 +1,457 @@
...
+
+        // TODO: re-enable
+        /// It's disabled becaues the uploads are suceeding, but then it doesn't
+        /// appear on chain
+        const forceDisableTurbo = true;
+        if (_turboUploadService.useTurboUpload && !forceDisableTurbo) {
+          final fileDataItem = await _arweave.prepareEntityDataItem(
+            newFileEntity,
+            profileState.wallet,
+            // TODO: key
```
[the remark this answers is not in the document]
matibat: Not a tech debt: it's out of scope

lib/components/new_button/new_button.dart
matibat: Out of scope for this story

lib/blocs/pin_file/pin_file_bloc.dart
matibat: The bloc already has an error state, there's the other story that will handle the UI

lib/l10n/app_en.arb
matibat: Hmm, I exported these changes on Friday. Was it changed on late Friday? If not then we have to update it on Localizely

thiagocarvalhodev (dismissed):
Thanks for the hard work!

lib/l10n/app_en.arb
thiagocarvalhodev: Checked on localizelly and the Strings match

karlprieb (approved):

PR #1257, opened 2023-07-26
chore(play store): stop updating app metadata

karlprieb:
Currently Fastlane is updating the app metadata (description, images, etc) on every deploy. We want to only manage changelogs through Fastlane and the app metadata should be managed directly on Google Play Store console


thiagocarvalhodev (approved):
LGTM

PR #1260, opened 2023-07-31
PE-3986: Show original owner in Details Panel

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/2rh3k465asjmg

### DIFF
--- a/lib/components/details_panel.dart
+++ b/lib/components/details_panel.dart
@@ -4,6 +4,7 @@ import 'package:ardrive/components/components.dart';
 import 'package:ardrive/components/dotted_line.dart';
 import 'package:ardrive/components/drive_rename_form.dart';
 import 'package:ardrive/components/sizes.dart';
+import 'package:ardrive/components/truncated_address.dart';
 import 'package:ardrive/core/arfs/entities/arfs_entities.dart';
 import 'package:ardrive/core/crypto/crypto.dart';
 import 'package:ardrive/entities/string_types.dart';
@@ -344,6 +345,9 @@ class _DetailsPanelState extends State<DetailsPanel> {
       metadataTxId = (widget.item as FileDataTableItem).metadataTx?.id;
     }
 
+    String? pinnedDataOwnerAddress =
+        (widget.item as FileDataTableItem).pinnedDataOwnerAddress;
+
     return [
       DetailsPanelItem(
         leading: CopyButton(text: widget.item.id),
@@ -413,6 +417,13 @@ class _DetailsPanelState extends State<DetailsPanel> {
         ),
         itemTitle: appLocalizationsOf(context).dataTxID,
       ),
+      if (pinnedDataOwnerAddress != null) ...[
+        sizedBoxHeight16px,
+        DetailsPanelItem(
+          leading: TruncatedAddress(walletAddress: pinnedDataOwnerAddress),
+          itemTitle: 'Uploaded by',
+        ),
+      ]
     ];
   }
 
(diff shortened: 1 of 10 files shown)


lib/components/truncated_address.dart
```diff
@@ -0,0 +1,39 @@
+import 'package:ardrive/utils/open_url.dart';
+import 'package:ardrive_ui/ardrive_ui.dart';
+import 'package:flutter/material.dart';
+
+class TruncatedAddress extends StatelessWidget {
```
thiagocarvalhodev, line 5: Thanks for abstracting the component @matibat !

lib/components/truncated_address.dart
```diff
@@ -0,0 +1,39 @@
...
+  final String walletAddress;
+  final double? fontSize;
+
+  const TruncatedAddress({Key? key, required this.walletAddress, this.fontSize})
+      : super(key: key);
+
+  String get _croppedAddress {
```
thiagocarvalhodev, line 12: Consider moving this method to a utils file, we may use the same logic in the future

lib/components/details_panel.dart
```diff
@@ -413,6 +417,13 @@ class _DetailsPanelState extends State<DetailsPanel> {
         ),
         itemTitle: appLocalizationsOf(context).dataTxID,
       ),
+      if (pinnedDataOwnerAddress != null) ...[
+        sizedBoxHeight16px,
+        DetailsPanelItem(
+          leading: TruncatedAddress(walletAddress: pinnedDataOwnerAddress),
+          itemTitle: 'Uploaded by',
```
thiagocarvalhodev, line 424: Missing localizations

thiagocarvalhodev (changes requested):
LGTM! After moving the wallet address logic and add the localizations I'll approve

lib/components/details_panel.dart
matibat: Exported the localizations: f5b61f0598230ff60471c81ed1e7b2f8e5123ffb

lib/components/truncated_address.dart
matibat: I was gonna say that we better wait for the moment when that happens before applying an unnecessary change, but then I realized that we're currently using the same logic with duplicated code in two different places. So it's worth doing it RN 🤓 

Factors out the method, plus unit tests: 0a5e285d3c8eba082fb35d9287f3589f7fc89aa6
Uses it on the TruncatedAddress component and ProfileCard: 4ea84e1606387119415ef235eb28ceebec8dba3d

lib/utils/truncate_string.dart
```diff
@@ -0,0 +1,13 @@
+// Truncate a string by adding ... in the middle
+String truncateString(String text, int offsetStart, int offsetEnd) {
```
thiagocarvalhodev, line 2: Nit: add named parameters: 

```dart
Sring truncateString({
  required String text, 
  required int offsetStart, 
  required int offsetEnd,
}) {

matibat: 87c280b45b157ea9c05adc4929ae6c37dc6bcc0c

lib/components/truncated_address.dart
matibat: I didn't use this component in the ProfileCard because of tiny differences (just the onClick action). But we could if we wanted to 😎

PR #1271, opened 2023-08-07
PE-4354: File not found for an existing file

matibat:


lib/services/arweave/arweave_service.dart
```diff
@@ -128,6 +128,14 @@ class ArweaveService {
     return query.data?.transaction;
   }
 
+  Future<InfoOfTransactionToBePinned$Query$Transaction?> getInfoOfTxToBePinned(
+    String txId,
+  ) async {
+    final query = await _gql.execute(InfoOfTransactionToBePinnedQuery(
+        variables: InfoOfTransactionToBePinnedArguments(txId: txId)));
+    return query.data!.transaction;
```
matibat, line 136:
```suggestion
    return query.data?.transaction;
```

lib/blocs/pin_file/file_id_resolver.dart
```diff
@@ -62,50 +64,28 @@ class NetworkFileIdResolver implements FileIdResolver {
 
   @override
   Future<ResolveIdResult> requestForTransactionId(TxID dataTxId) async {
```
karlprieb, line 66: dataTxId is not a good name as it can be any transaction. Let's call it txId

thiagocarvalhodev (approved):

PR #1277, opened 2023-08-09
PE-4352: Maybe fixes carrying over the custom metadata and custom gql tags for folder and drive

matibat:
**Note** I say "maybe fixes" because there's no easy way to create folders and drives with custom metadata as for today

### DIFF
--- a/lib/models/drive.dart
+++ b/lib/models/drive.dart
@@ -1,4 +1,7 @@
+import 'dart:convert';
+
 import 'package:ardrive/entities/entities.dart';
+import 'package:arweave/arweave.dart';
 
 import './database/database.dart';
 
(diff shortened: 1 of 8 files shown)


thiagocarvalhodev:
> Note I say "maybe fixes" because there's no easy way to create folders and drives with custom metadata as for today

Would not be possible to mock those tags in a unit test?

lib/models/drive.dart
```diff
@@ -1,18 +1,32 @@
...
+          ? DriveAuthMode.password
+          : DriveAuthMode.none,
+    );
+
+    drive.customJsonMetadata =
+        customJsonMetadata != null ? jsonDecode(customJsonMetadata!) : null;
+    drive.customGqlTags = customGQLTags != null
+        ? (jsonDecode(customGQLTags!) as List<dynamic>)
+            .map((maybeTag) => Tag.fromJson(maybeTag))
+            .toList()
```
thiagocarvalhodev, line 28: Consider implementing a new method to parse the tags. I saw the same implementation on `FileEntity` so it's worth implementing a helper.

lib/models/folder_entry.dart
```diff
@@ -1,12 +1,25 @@
...
+      parentFolderId: parentFolderId,
+      name: name,
+    );
+    folder.customJsonMetadata =
+        customJsonMetadata != null ? jsonDecode(customJsonMetadata!) : null;
+    folder.customGqlTags = customGQLTags != null
+        ? (jsonDecode(customGQLTags!) as List<dynamic>)
+            .map((maybeTag) => Tag.fromJson(maybeTag))
+            .toList()
+        : null;
```
thiagocarvalhodev, line 22: The same of the previous comment.

thiagocarvalhodev (changes requested):
Requested only to extract a method that is duplicated. Also, even if the custom metadata on Drives and Folders are something we don't have a nice way to implement, I'm pretty confident that this implementation fixes the problem. The unit tests for this seems to be right on the other PR for metadata. For graphQL I don't see any reason to don't work if it works for files.

matibat:
> > Note I say "maybe fixes" because there's no easy way to create folders and drives with custom metadata as for today
>
> Would not be possible to mock those tags in a unit test?

Yes! It is, I'll do that at the end of the Story. But what I mean is in regard to manual testing.

lib/models/drive.dart
matibat: DRY: 215b408ef5b05a1acc239fccc47d2d7b2e760295
Test: 0311c6579d11b38a551cd66895956e89682dfbd0

thiagocarvalhodev (approved):
LGTM

thiagocarvalhodev:
The PR title and description are out of date. Can you update it?

PR #1286, opened 2023-08-11
PE-4389: Makes disabled items not to run the onTap method

matibat:


thiagocarvalhodev (approved):
LTGM! Thanks for the fix!

PR #1301, opened 2023-08-23
PE-2586: New/Plus button redesign

matibat:


lib/components/app_bottom_bar.dart
matibat, line 1: **For the reviewer -** This is because I made the field to be required. I moved the default value to here.

lib/components/new_button/new_button.dart
```diff
@@ -21,19 +21,19 @@ class NewButton extends StatelessWidget {
...
     required this.driveDetailState,
-    this.child,
+    required this.child,
     this.isBottomNavigationButton = false,
     this.anchor = const Aligned(
-      follower: Alignment.bottomCenter,
-      target: Alignment.topCenter,
+      follower: Alignment.bottomRight,
+      target: Alignment.topRight, // Top Right
     ),
```
matibat, line 29: I think we can make it either required or nullable.

lib/components/new_button/new_button.dart
```diff
@@ -78,30 +78,34 @@ class NewButton extends StatelessWidget {
...
+                              final isLastItem = index == items.length - 1;
+
                               return Column(
                                 children: [
                                   GestureDetector(
+                                    behavior: HitTestBehavior.translucent,
                                     onTap: () {
-                                      Navigator.pop(context);
-                                      items[index].onClick();
+                                      if (!item.isDisabled) {
```
matibat, line 89: I'm pretty sure the Tree Component does this check under the hood. Double-check and remove if so.

lib/components/new_button/new_button.dart
```diff
@@ -120,21 +124,12 @@ class NewButton extends StatelessWidget {
...
-            child: ArDriveFAB(
-              backgroundColor:
-                  ArDriveTheme.of(context).themeData.colors.themeAccentBrand,
-              child: ArDriveIcons.plus(
-                color: Colors.white,
-              ),
-            ),
-          ),
+      rootNode: _buildDropdownTree(context),
+      // nestedPortal: nestedPortal,
```
matibat, line 131:
```suggestion
```

matibat (commented):
Self-review checkpoint

matibat:
TODO: fix the conflicts

lib/components/new_button/new_button.dart
matibat: Just double-checked - You can remove it and do a manual test just to confirm

lib/components/new_button/new_button.dart
matibat: Made it be required - bdd868d4c9569e5f1697272eba779ea609a81ac8

matibat:
Replaced by: #1303

PR #1302, opened 2023-08-23
PE-4480: clean up logs

thiagocarvalhodev:
### Description:
This PR addresses a concern related to logging potentially sensitive information such as drive, folder, and file names. With the introduction of the export log feature, we are now taking measures to sanitize these logs by removing or obfuscating any sensitive content.

#### Previous Behavior: 
Sensitive data like drive, folder, and file names were being logged onto the console.

#### New Behavior: 
This information is now appropriately handled to ensure no sensitive data is exposed.

### Note:
This change is mainly preventive, as we were not logging this sensitive information in the production environment

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/03opd85l57pvg


lib/services/arweave/arweave_service.dart
```diff
@@ -485,7 +489,7 @@ class ArweaveService {
           () async => await Future.wait(
                 driveTxs.map((e) => client.api.getSandboxedTx(e.id)),
               ), onRetry: (Exception err) {
-        logger.i(
+        logger.w(
           'Retrying for get unique user drive entities on Exception: ${err.toString()}',
```
kunstmusik, line 493: Should this pass err as second argument instead of inserting to the string?

lib/utils/graphql_retry.dart
```diff
@@ -26,22 +26,19 @@ class GraphQLRetry {
         maxAttempts: maxAttempts,
         onRetry: (exception) {
           onRetry?.call(exception);
-          logger.i('''
-          Retrying Query: ${query.toString()}\n
-          On Exception: ${exception.toString()}
-          ''');
+          logger.i(
+              'Retrying Query: ${query.toString()}. On Exception: ${exception.toString()}');
```
kunstmusik, line 30: Should this pass exception as second argument rather than insert into string?

lib/services/arweave/arweave_service.dart
thiagocarvalhodev: The `w` doesn't have the error parameter. Do you think we need to add it?

lib/utils/graphql_retry.dart
thiagocarvalhodev: The `i` doesn't have the error parameter

kunstmusik (approved):

PR #1316, opened 2023-08-29
PE-4502: remove arfs vesion filter

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6h56afqgf9f00


matibat (commented):
We should carefully discuss this with the team before applying it.

My only concern is regarding support:
Currently, the Flutter App is aware of which ArFS version it supports. If we removed this filter, then we'd be letting it fetch Transactions form an unsupported version (e.g. future releases).

matibat (dismissed):
LGTM

matibat (dismissed):

matibat (approved):
LGTM.

PR #1328, opened 2023-08-30
PE-4502: filters the arfs transactions in the client side

matibat:
### Changes
- Filters the ArFS transactions on the client side
- Make the single-item-queries request pages of 10 items until the first one matches.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6ukjoa02tga8o

### DIFF
--- a/lib/services/arweave/arweave_service.dart
+++ b/lib/services/arweave/arweave_service.dart
@@ -651,31 +651,50 @@ class ArweaveService {
   /// Returns `null` if no valid file is found.
 
   Future<Privacy?> getFilePrivacyForId(String fileId) async {
-    final firstOwnerQuery = await _gql.execute(FirstFileEntityWithIdOwnerQuery(
-        variables: FirstFileEntityWithIdOwnerArguments(fileId: fileId)));
-
-    if (firstOwnerQuery.data!.transactions.edges.isEmpty) {
+    final fileOwner = await getOwnerForFileEntityWithId(fileId);
+    if (fileOwner == null) {
       return null;
     }
 
-    final fileOwner =
-        firstOwnerQuery.data!.transactions.edges.first.node.owner.address;
+    String cursor = '';
 
-    final latestFileQuery = await _gql.execute(LatestFileEntityWithIdQuery(
-        variables:
-            LatestFileEntityWithIdArguments(fileId: fileId, owner: fileOwner)));
+    while (true) {
+      final latestFileQuery = await _gql.execute(
+        LatestFileEntityWithIdQuery(
+          variables: LatestFileEntityWithIdArguments(
+            fileId: fileId,
+            owner: fileOwner,
+            after: cursor,
+          ),
+        ),
+      );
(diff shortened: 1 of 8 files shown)


matibat:
Follow-up work of #1316

lib/services/arweave/arweave_service.dart
```diff
@@ -430,14 +436,26 @@ class ArweaveService {
     String userAddress, {
     int maxRetries = defaultMaxRetries,
   }) async {
+    // FIXME: make me run all pages
```
matibat, line 439: This comment refers to #1330 
TODO: either integrate that PR into this one, or remove the comment

lib/services/arweave/graphql/queries/FirstFileEntityWithIdOwner.graphql
```diff
@@ -1,17 +1,24 @@
-query FirstFileEntityWithIdOwner($fileId: String!) {
+query FirstFileEntityWithIdOwner($fileId: String!, $after: String) {
   transactions(
     first: 1
```
matibat, line 3:
```suggestion
    first: 10
```

matibat (commented):
Self-review checkpoint

lib/utils/arfs_txs_filter.dart
```diff
@@ -0,0 +1,17 @@
...
+final supportedArFSVersions = ['0.10', '0.11', '0.12', '0.13'];
+
+bool doesTagsContainValidArFSVersion(List<Tag> tags) {
+  return tags.any(
+    (tag) =>
+        tag.name == EntityTag.arFs && supportedArFSVersions.contains(tag.value),
+  );
+}
+
+class Tag {
```
thiagocarvalhodev, line 12: We already have a class `Tag` on our `arweave-dart` package, let's use the same class here

thiagocarvalhodev (changes requested):
Just a minor change

lib/utils/arfs_txs_filter.dart
matibat: Yes. And we also have another one from the auto-generated code of GQL as well, so I opted to use a custom one. I'd indeed prefer to simply describe the shape of the tags so I could take any implementation but didn't know how to.

thiagocarvalhodev: >I'd indeed prefer to simply describe the shape of the tags so I could take any implementation but didn't know how to.

So we can re-use the tag from the `arweaved-dart`, this way we can avoid some confusion in the future because of 3 class that represents the same entity, does that make sense?

matibat: Makes sense. Applied - 2557e91b6f4d309b4833a54e33ed513bf9099006

thiagocarvalhodev (approved):
LGTM! thanks mati!

PR #1330, opened 2023-08-30
PE-4502: makes UserDriveEntities query use pagination

matibat:
I encountered we're requesting a maximum of 100 individual drives (perhaps intended). This PR removes that limit.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/28u73hu6uvk80


matibat:
Related to: #1328

matibat (commented):
Self-review checkpoint

thiagocarvalhodev (approved):

PR #1331, opened 2023-08-30
PE-4515: fix background color on divider of the new button

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/4c7j6hopc2jrg


matibat (approved):
LGTM

matibat:
The Android Build is failing because it is using the initial PR Title we've set. I tried re-running it but it's still using the old title as a reference. I'm gonna push one more commit, perhaps it makes it to run again.

PR #1335, opened 2023-09-05
PE-4454: Add Promo Code Field to Turbo Modal

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/0ivtcq3m4nu38

### DIFF
--- a/lib/turbo/topup/blocs/payment_review/payment_review_state.dart
+++ b/lib/turbo/topup/blocs/payment_review/payment_review_state.dart
@@ -45,7 +45,6 @@ class PaymentReviewPaymentModelLoaded extends PaymentReviewState {
     String? subTotal,
     String? credits,
     DateTime? quoteExpirationDate,
-    // PaymentUserInformation? userInformation,
     String? promoDiscount,
   }) {
     return PaymentReviewPaymentModelLoaded(
--- a/lib/turbo/topup/blocs/payment_review/payment_review_state.dart
+++ b/lib/turbo/topup/blocs/payment_review/payment_review_state.dart
@@ -17,6 +17,7 @@ class PaymentReviewLoading extends PaymentReviewPaymentModelLoaded {
     required super.total,
     required super.subTotal,
     required super.credits,
+    required super.promoDiscount,
   });
 }
 
@@ -28,12 +29,14 @@ class PaymentReviewPaymentModelLoaded extends PaymentReviewState {
   final String total;
   final String subTotal;
   final String credits;
+  final String? promoDiscount;
   final DateTime quoteExpirationDate;
 
   const PaymentReviewPaymentModelLoaded({
     required this.total,
     required this.subTotal,
     required this.credits,
+    required this.promoDiscount,
     required this.quoteExpirationDate,
   });
 
@@ -42,13 +45,15 @@ class PaymentReviewPaymentModelLoaded extends PaymentReviewState {
     String? subTotal,
     String? credits,
     DateTime? quoteExpirationDate,
-    PaymentUserInformation? userInformation,
+    // PaymentUserInformation? userInformation,
+    String? promoDiscount,
   }) {
     return PaymentReviewPaymentModelLoaded(
       total: total ?? this.total,
       subTotal: subTotal ?? this.subTotal,
       credits: credits ?? this.credits,
       quoteExpirationDate: quoteExpirationDate ?? this.quoteExpirationDate,
+      promoDiscount: promoDiscount ?? this.promoDiscount,
     );
   }
 }
@@ -59,6 +64,7 @@ class PaymentReviewLoadingQuote extends PaymentReviewPaymentModelLoaded {
     required super.total,
     required super.subTotal,
     required super.credits,
+    required super.promoDiscount,
   });
 }
 
@@ -68,6 +74,7 @@ class PaymentReviewQuoteLoaded extends PaymentReviewPaymentModelLoaded {
     required super.total,
     required super.subTotal,
     required super.credits,
+    required super.promoDiscount,
   });
 }
 
@@ -79,6 +86,7 @@ class PaymentReviewQuoteError extends PaymentReviewPaymentModelLoaded {
     required super.subTotal,
     required super.credits,
     required super.quoteExpirationDate,
+    required super.promoDiscount,
   });
 }
 
@@ -88,6 +96,7 @@ class PaymentReviewPaymentSuccess extends PaymentReviewPaymentModelLoaded {
     required super.total,
     required super.subTotal,
     required super.credits,
+    required super.promoDiscount,
   });
 }
 
@@ -107,6 +116,7 @@ class PaymentReviewPaymentError extends PaymentReviewPaymentModelLoaded {
     required super.subTotal,
     required super.credits,
     required this.errorType,
+    required super.promoDiscount,
   });
 }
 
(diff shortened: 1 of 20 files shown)


matibat:
@matibat, the tests are not passing!

thiagocarvalhodev (commented):
@matibat , the promo field doesn't apply the promo, even though the text field shows the green border indicating the code entered is valid, but no promotion or any other feedback is shown

https://github.com/ardriveapp/ardrive-web/assets/32248947/99443d0f-b544-4fa5-9950-8ffd8166a7b5

thiagocarvalhodev (changes requested):
Left a couple of comments and some required changes.

lib/turbo/topup/blocs/payment_review/payment_review_state.dart
```diff
@@ -42,13 +45,15 @@ class PaymentReviewPaymentModelLoaded extends PaymentReviewState {
     String? subTotal,
     String? credits,
     DateTime? quoteExpirationDate,
-    PaymentUserInformation? userInformation,
+    // PaymentUserInformation? userInformation,
```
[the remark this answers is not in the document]
matibat: Ahh yes. I was pretty sure we were not using that piece of data. IMO we can remove that.

matibat:
RE:
> @matibat , the promo field doesn't apply the promo, even though the text field shows the green border indicating the code entered is valid, but no promotion or any other feedback is shown

You have to click the Apply button, although it seems it's not visible on the light theme.

Also, the only three valid Codes I've enabled are `TURBO`, `ARDRIVE`, and `MATI`.

matibat:
Added an extra code: `ERROR` that will simulate a network error.

lib/turbo/turbo.dart
```diff
@@ -186,6 +192,43 @@ class Turbo extends Disposable {
...
+  //       // setState(() {
+  //       //   _promoCode = '';
+  //       //   _promoCodeInvalid = true;
+  //       //   _promoCodeController.clear();
+  //       // });
+  //     }
+  //   }
+  // }
+
+  Future<double?> getPromoDiscountFactor(String promoCode) async {
```
thiagocarvalhodev, line 217: The Turbo class should use an abstraction for the logic of the promo code and implement a method using an abstraction. We can see similar examples on the other methods. 

What I suggest is to implement this method in a new class to retrieve the promo code and that also validates the code, example: `TurboPromoDiscountValidator`, that will get a promo code, fetch the API and validate if it's valid or not depending on the service response.

We can also implement a method on the `PaymentService` class to mock the http call, and use this class on the `TurboPromoDiscountValidator`, so, when the backend is done, we can just replace the call on the payment service and we won't change anything else in the code.

thiagocarvalhodev: With that, we can also implement the unit tests validating a mocked response from the `PaymentService`

lib/turbo/turbo.dart
```diff
@@ -186,6 +192,43 @@ class Turbo extends Disposable {
 
     logger.d('Turbo disposed');
   }
+
+  // Future<double?> _getPromoDiscountFactor(String promoCode) async {
```
thiagocarvalhodev, line 196: Commented code

test/turbo/blocs/payment_form/payment_form_bloc_test.dart
```diff
@@ -16,7 +16,11 @@ void main() {
     setUp(() {
       mockTurbo = MockTurbo();
       initialPriceEstimate = PriceEstimate(
-          credits: BigInt.from(10), priceInCurrency: 10, estimatedStorage: 1);
+        credits: BigInt.from(10),
```
thiagocarvalhodev, line 19: We are not testing the case when the promo discount is different than 0.
We need to test:
 - promo discount different from 0
 - promo discount has an error

thiagocarvalhodev (commented):
We will need just more one refactor on the `Turbo` class, and also to update the unit tests on the PaymentFormBloc

thiagocarvalhodev:
Now the feature is looking great! Thanks for the fix on Theme, @matibat

lib/turbo/turbo.dart
matibat: Is this what you meant? - 5912114a34c596c110d74bf223d66bd60dd65cef

lib/turbo/topup/blocs/payment_review/payment_review_state.dart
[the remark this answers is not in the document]
matibat: 26aa7a40c25351c08ec4d70cc89cf88290d07a96

lib/turbo/topup/views/topup_payment_form.dart
```diff
@@ -574,6 +596,178 @@ class TurboPaymentFormViewState extends State<TurboPaymentFormView> {
...
+    });
+    await Future.delayed(const Duration(seconds: 2));
+    setState(() {
+      _promoCodeFetching = false;
+    });
+
+    final isValid = validCodes.keys.contains(textInPromoCode);
+
+    return isValid ? validCodes[textInPromoCode] : null;
+  }
```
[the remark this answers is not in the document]
matibat: Applied

lib/turbo/turbo.dart
matibat: Removed

test/turbo/blocs/payment_form/payment_form_bloc_test.dart
matibat: The bloc won't behave differently for a different promoDiscountFactor. So the zero value is not distinct (in behavior) from one.

Also, there's no error case for this bloc.

thiagocarvalhodev: >Also, there's no error case for this bloc.

I mean the test for emitting this state:

```dart
emit(
        PaymentFormLoaded(
          state.priceEstimate,
          _expirationTimeInSeconds(turbo.maxQuoteExpirationDate),
          (state as PaymentFormLoaded).supportedCountries,
          isPromoCodeInvalid: false,
          isFetchingPromoCode: false,
          errorFetchingPromoCode: true,
        ),
      );

lib/turbo/turbo.dart
thiagocarvalhodev: Exactly! We are now just missing the unit tests

test/turbo/blocs/payment_form/payment_form_bloc_test.dart
matibat: Makes the class be testable by injecting a method to make it be deterministic - d228ee900704c19d7bdf22a691f6d9ed08973d0d
Unit tests those cases - 2428e7394be807a3685c3819320be238072c1189

thiagocarvalhodev (approved):
Tested and it's working nicely! I just noticed that we can improve the UX by a simple change. @matibat , I think it would be a nice UX if we also could load the promo code by hitting enter on the field. We have already a property on the component that does that: `onFieldSubmitted`. It's a callback that will be triggered when you hit enter, so you can just call the same function you call on the `Apply` button.

I'm approving, if you think it makes sense just let me know so I'll approve again.

matibat:
Changes are carried to #1349

PR #1337, opened 2023-09-05
PE-4526: update ardrive_ui

thiagocarvalhodev:


pubspec.lock
```diff
@@ -79,11 +79,11 @@ packages:
     dependency: "direct main"
     description:
       path: "."
-      ref: PE-4515
-      resolved-ref: "11c4cb89beb769fa39b2a963ced2c6de4d3806f4"
+      ref: "v1.9.0"
```
matibat, line 82: Was this manually edited? Shouldn't we let Flutter re-generate it?

matibat (commented):
Left one question

pubspec.lock
thiagocarvalhodev: Flutter generated it

matibat (approved):

PR #1341, opened 2023-09-06
PE-4537: Enable Pins for private drives

matibat:


lib/blocs/file_download/personal_file_download_cubit.dart
```diff
@@ -117,6 +124,19 @@ class ProfileFileDownloadCubit extends FileDownloadCubit {
     if (drive.drivePrivacy == DrivePrivacy.private) {
       SecretKey? driveKey;
 
+      final isPinFile = _file.pinnedDataOwnerAddress != null;
+      if (isPinFile) {
```
thiagocarvalhodev, line 128: @kunstmusik , we must verify how the multi-files download behaves with Pins. The point here is that we don't decrypt the content - dataTx - of the Pin, but only its metadata, if the pin is created on a private drive.

We must not decrypt the `dataTx` of the pin when the pin is on a private drive, so we will have a similar validation on the multi-files download Bloc.

We also need to create a tech debt to unify all the downloads options in a single abstraction, and use that abstraction in different blocs or cubits.

lib/components/new_button/new_button.dart
```diff
@@ -405,9 +405,7 @@ class NewButton extends StatelessWidget {
             name: appLocalizations.newFolder,
             icon: ArDriveIcons.iconNewFolder1(size: defaultIconSize),
           ),
-          if (context.read<ConfigService>().config.enablePins &&
-              drive != null &&
-              drive?.privacy == 'public')
+          if (context.read<ConfigService>().config.enablePins && drive != null)
```
thiagocarvalhodev, line 408: We can remove this feature flag for now as we don't need it

thiagocarvalhodev (changes requested):
Looking good! Let's just fix the lint and remove the feature flag.

We'll need to adapt the multi-file downloads for private pins, @kunstmusik .

thiagocarvalhodev (approved):
LGTM!!

kunstmusik (approved):

PR #1360, opened 2023-09-14
PE-4608: lets turbo balance check to throw instead of assuming Zero balance 

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6n1sjq48khmv0


thiagocarvalhodev:
That's the expected behavior. Closing as we won't apply these changes.

PR #1369, opened 2023-09-19
PE-4632: Persistent loading issue when creating drives

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5uboc4dgpe1a0


test/blocs/drive_create_cubit_test.dart
```diff
@@ -1,12 +1,11 @@
 @Tags(['broken'])
```
matibat, line 1: I attempted to fix this, but it was taking too much time so I left if Skipped still

PR #1371, opened 2023-09-20
PE-4645: Unable to create Pins in both public and private drive

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/28tpsr3nah7u8


thiagocarvalhodev (approved):
lgtm

PR #1384, opened 2023-09-26
PE-4477: ArDrive logout issue when switching ArConnect wallets after seed-phrase/json login

matibat:


lib/app_shell.dart
```diff
@@ -31,17 +32,27 @@ class AppShellState extends State<AppShell> {
   bool _showProfileOverlay = false;
   bool _showWalletSwitchDialog = true;
   @override
-  Widget build(BuildContext context) => BlocBuilder<DrivesCubit, DrivesState>(
+  Widget build(BuildContext context) => BlocBuilder<ProfileCubit, ProfileState>(
```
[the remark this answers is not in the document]
matibat: I'd suggest reverting the change because the builder is running on every build of the bloc and we would be changing that behavior.

lib/authentication/ardrive_auth.dart
```diff
@@ -26,7 +26,7 @@ abstract class ArDriveAuth {
   Future<User> unlockWithBiometrics({required String localizedReason});
   Future<User> unlockUser({required String password});
   Future<void> logout();
-  User? get currentUser;
+  User get currentUser;
```
matibat, line 29: **For the reviewer -** This attribute is never accessed as nullable anywhere in the code, and the implementation is always returning a non-null value.

lib/authentication/ardrive_auth.dart
```diff
@@ -216,32 +216,33 @@ class ArDriveAuthImpl implements ArDriveAuth {
...
-
-        _secureKeyValueStore.remove('password');
-        _secureKeyValueStore.remove('biometricEnabled');
-
+        await _secureKeyValueStore.remove('password');
+        await _secureKeyValueStore.remove('biometricEnabled');
         currentUser = null;
         firstPrivateDriveTxId = null;
-
+        await _disconnectFromArConnecct();
```
matibat, line 223: **For the reviewer -** This is the trick: to trigger the wallet-switch event just after making currentUser be null. That way the class knows it's not logged in.

lib/authentication/ardrive_auth.dart
```diff
@@ -216,32 +216,33 @@ class ArDriveAuthImpl implements ArDriveAuth {
...
       (await _metadataCache).clear();
     } catch (e) {
       logger.e('Failed to logout user', e);
       throw AuthenticationFailedException('Failed to logout user');
     }
   }
 
+  Future<void> _disconnectFromArConnecct() async {
+    final hasArConnectPermissions = await _arConnectService.checkPermissions();
+    if (hasArConnectPermissions) {
```
matibat, line 237: **For the reviewer -** Now, instead of checking if the current profile is ArConnect as the condition to log out, we check if ArConnect is connected.

lib/authentication/login/blocs/login_bloc.dart
```diff
@@ -260,10 +274,6 @@ class LoginBloc extends Bloc<LoginEvent, LoginState> {
       await _arDriveAuth.logout();
     }
 
-    if (_isArConnectWallet()) {
```
matibat, line 263: **For the reviewer -** This call to disconnect was redundant because `_arDriveAuth.logout()` already does it.

lib/app_shell.dart
```diff
@@ -31,17 +32,27 @@ class AppShellState extends State<AppShell> {
...
-              );
-            }
-            //Used to prevent the dialog being shown multiple times.
-            _showWalletSwitchDialog = false;
+            context
+                .read<ProfileCubit>()
+                .isCurrentProfileArConnect()
+                .then((isCurrentProdileArConnect) {
+              if (_showWalletSwitchDialog) {
+                if (isCurrentProdileArConnect) {
```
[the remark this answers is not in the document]
matibat: 75991d6152d6bc8099a493a025c0c0bf873225fa

lib/authentication/ardrive_auth.dart
```diff
@@ -216,32 +216,33 @@ class ArDriveAuthImpl implements ArDriveAuth {
...
       await _databaseHelpers.deleteAllTables();
-
       (await _metadataCache).clear();
     } catch (e) {
       logger.e('Failed to logout user', e);
       throw AuthenticationFailedException('Failed to logout user');
     }
   }
 
+  Future<void> _disconnectFromArConnecct() async {
```
thiagocarvalhodev, line 235:
```suggestion
  Future<void> _disconnectFromArConnect() async {
```

thiagocarvalhodev (dismissed):
LGTM

lib/authentication/ardrive_auth.dart
thiagocarvalhodev: Thanks for implementing that, @matibat .

thiagocarvalhodev (approved):

PR #1387, opened 2023-09-27
PE-4687: Free snapshots aren't using Turbo as it's expected

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6mqpvsrn50m2o

### DIFF
--- a/lib/blocs/create_snapshot/create_snapshot_cubit.dart
+++ b/lib/blocs/create_snapshot/create_snapshot_cubit.dart
@@ -72,6 +72,9 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
   bool _sufficientArBalance = false;
   bool _isFreeThanksToTurbo = false;
 
+  bool get _useTruboUpload =>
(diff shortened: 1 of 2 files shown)


lib/blocs/create_snapshot/create_snapshot_cubit.dart
```diff
@@ -72,6 +72,9 @@ class CreateSnapshotCubit extends Cubit<CreateSnapshotState> {
   bool _sufficientArBalance = false;
   bool _isFreeThanksToTurbo = false;
 
+  bool get _useTruboUpload =>
```
kunstmusik, line 75: rename to _useTurboUpload

kunstmusik (changes requested):

lib/blocs/create_snapshot/create_snapshot_cubit.dart
matibat: Nice catch! Here it is: abf45c630258826398907a7d223b405761101161

kunstmusik (approved):

PR #1392, opened 2023-09-28
PE-4081: feat(share file): allow to share file when the file is pending

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/790aqo2pk0mjo

### DIFF
--- a/lib/components/file_share_dialog.dart
+++ b/lib/components/file_share_dialog.dart
@@ -58,9 +58,35 @@ class FileShareDialogState extends State<FileShareDialog> {
                 const Center(child: CircularProgressIndicator())
               else if (state is FileShareLoadedFailedFile)
                 Text(appLocalizationsOf(context).shareFailedFile)
-              else if (state is FileShareLoadedPendingFile)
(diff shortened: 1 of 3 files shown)


lib/components/file_share_dialog.dart
```diff
@@ -58,9 +58,35 @@ class FileShareDialogState extends State<FileShareDialog> {
...
+                              .themeData
+                              .colors
+                              .themeWarningEmphasis,
+                        ),
+                        const SizedBox(
+                          width: 8,
+                        ),
+                        Flexible(
+                          child: Text(
+                            'Warning: This file is currently pending and may not be immediately accessible.',
```
karlprieb, line 78: can you create the translation key?

thiagocarvalhodev: We have one already. We need to change it on the localizely. It's not possible now because changing there, we'll update the same translation on other PRs

karlprieb (approved):

PR #1393, opened 2023-09-28
PE-4544: lock multi select when a modal is open

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/25csvr5sv67i0

### DIFF
--- a/lib/core/activity_tracker.dart
+++ b/lib/core/activity_tracker.dart
@@ -1,5 +1,8 @@
-class ActivityTracker {
+import 'package:flutter/foundation.dart';
+
+class ActivityTracker extends ChangeNotifier {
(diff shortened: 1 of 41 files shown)


lib/core/activity_tracker.dart
```diff
@@ -8,7 +11,16 @@ class ActivityTracker {
     _isToppingUp = value;
   }
 
+  void setShowingAnyDialog(bool value) {
```
karlprieb, line 14: could be `_isShowingDialog` and `setShowingDialog`

karlprieb (approved):

PR #1405, opened 2023-10-04
PE-4732: video-preview-doesnt-accept-all-mov-and-mpg-formats

kunstmusik:
* Redesigned to accept all video mime types, but show error if format is unsupported by underlying video player

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5ov25e60c7a2o


thiagocarvalhodev (approved):
LGTM

PR #1410, opened 2023-10-05
PE-4738: fix: Unable to continue drive creation after creating the first drive as a new user

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/4jvfj00dbskkg


matibat (approved):
LGTM - Let's create a tech-debt ticket to make that component hard to use incorrectly.

PR #1413, opened 2023-10-06
PE-4752: Wallet address in login page PoC

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/1l2lgcfblkooo

### DIFF
--- a/lib/authentication/login/blocs/login_bloc.dart
+++ b/lib/authentication/login/blocs/login_bloc.dart
@@ -5,11 +5,13 @@ import 'package:ardrive/authentication/ardrive_auth.dart';
 import 'package:ardrive/entities/profile_types.dart';
 import 'package:ardrive/services/arconnect/arconnect.dart';
 import 'package:ardrive/services/arconnect/arconnect_wallet.dart';
+import 'package:ardrive/user/repositories/user_repository.dart';
 import 'package:ardrive/user/user.dart';
 import 'package:ardrive/utils/html/html_util.dart';
 import 'package:ardrive/utils/logger/logger.dart';
 import 'package:ardrive_io/ardrive_io.dart';
 import 'package:arweave/arweave.dart';
+import 'package:arweave/utils.dart';
 import 'package:bip39/bip39.dart' as bip39;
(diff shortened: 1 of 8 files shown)


lib/authentication/login/blocs/login_bloc.dart
```diff
@@ -34,11 +37,21 @@ class LoginBloc extends Bloc<LoginEvent, LoginState> {
   @visibleForTesting
   ProfileType? profileType;
 
+  Future<String?> getWalletAddress() async {
```
thiagocarvalhodev, line 40: Can we implement this method on the `ArDriveAuth` class?

lib/user/repositories/user_repository.dart
```diff
@@ -14,6 +14,7 @@ abstract class UserRepository {
     Wallet wallet,
   );
   Future<void> deleteUser();
+  Future<String?> getOwnerOfDefaultProfile();
```
thiagocarvalhodev, line 17: Can you add a comment here explaining why we can receive a null value? e.g. returns null if no profile is available

thiagocarvalhodev (changes requested):
Left a few changes that you can consider implementing

lib/authentication/login/blocs/login_bloc.dart
matibat: There we go - 7f293297c9285c708b5fdf29c9ed357c1d6f46c6

lib/user/repositories/user_repository.dart
matibat: Here it is - 756f7141bbff1adaa57dc4d6ebfa456d8449b3af

thiagocarvalhodev (changes requested):
Thanks for applying the changes! We are just missing the tests on the new method of the repository. I can approve after that

thiagocarvalhodev (dismissed):

thiagocarvalhodev (approved):

PR #1419, opened 2023-10-11
PE-4741: Track Plausible events

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/7sjsl6u5o82u0


matibat (approved):
Got permission to approve this one from @thiagocarvalhodev.

And I'm approving the latest commit by myself, which makes it not to use custom events but the standard `pageview` one. We can make it use custom events when we configure them in the private dashboard.

matibat:
I made a follow-up PR of this making the API of the class be prepared for adding plausible custom events for when we configure them - #1431

matibat:
This feature is lacking unit tests!

PR #1431, opened 2023-10-20
PE-4741: prepares the API of the Plausible event tracker to start using custom events 

matibat:
Custom events are being tracked as page views. This is a chore change.

This is not the feature PR - Please refer to #1419.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/0p3r77f599dd8

### DIFF
--- a/lib/authentication/login/views/login_page.dart
+++ b/lib/authentication/login/views/login_page.dart
@@ -47,7 +47,7 @@ class _LoginPageState extends State<LoginPage> {
   void initState() {
     super.initState();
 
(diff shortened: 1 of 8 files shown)


matibat:
Well... This feature is indeed changing the name of the "events" (the page view paths). I can revert the changes, but I'd choose to have them as it is (with no "Page" prefix).

I'd also say we better wait to have the custom events configured so we can have more reliable data. That way we'd be also testing changing the name of the page URL paths vs custom events.

PR #1435, opened 2023-10-23
PE-4821: Erroneous 'Preview' display in private drive audio playback

matibat:


lib/components/details_panel.dart
```diff
@@ -119,19 +120,20 @@ class _DetailsPanelState extends State<DetailsPanel> {
     required FsEntryInfoState infoState,
   }) {
     final tabs = [
-      if (!(widget.isSharePage && !mobileView)) ...[
+      if (!(widget.isSharePage && !mobileView) &&
```
thiagocarvalhodev, line 123: Can you extract this logic to a new variable?

lib/components/details_panel.dart
```diff
@@ -238,8 +241,11 @@ class _DetailsPanelState extends State<DetailsPanel> {
                     ],
                   ),
                 ),
-              if (previewState is FsEntryPreviewSuccess &&
-                  !(widget.isSharePage))
+              if ((previewState is FsEntryPreviewSuccess &&
```
thiagocarvalhodev, line 244: Same comment here

thiagocarvalhodev (changes requested):

lib/components/details_panel.dart
matibat: ef1e3215f3bb5ff448603bbcafb699e5fd87adcf

lib/pages/drive_detail/components/fs_entry_preview_widget.dart
```diff
@@ -21,6 +21,11 @@ class _FsEntryPreviewWidgetState extends State<FsEntryPreviewWidget> {
   @override
   Widget build(BuildContext context) {
     switch (widget.state.runtimeType) {
+      case FsEntryPreviewUnavailable:
+        return const Center(
+          child: Text('Preview unavailable'),
```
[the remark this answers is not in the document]
matibat: Indeed this text will never be shown. This change can be reverted.

thiagocarvalhodev (approved):
LGTM

PR #1436, opened 2023-10-23
PE-4851: upload with wakelock

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/701rmnt1vo4ho


lib/blocs/upload/upload_cubit.dart
```diff
@@ -708,6 +709,9 @@ class UploadCubit extends Cubit<UploadState> {
   void retryTask(UploadController controller, UploadTask task) {}
 
   Future<void> _uploadUsingArDriveUploader() async {
+    await WakelockPlus.enable();
```
matibat, line 712: Shouldn't we disable it once we're done with the upload?

matibat (commented):
LGTM - Left a single question.

PR #1439, opened 2023-10-24
PE-4866: implement a contract reader that reads our own cache

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/55gh4lgm6mbs0


packages/pst/lib/src/implementations/pst_web.dart
```diff
@@ -3,7 +3,8 @@ library pst;
 
 import 'dart:convert';
 
-import 'package:ardrive/types/transaction_id.dart';
+import 'package:ardrive_utils/ardrive_utils.dart';
+// ignore: depend_on_referenced_packages
```
kunstmusik, line 7: Shouldn't we add a dependency rather than use this ignore?

lib/turbo/utils/utils.dart
```diff
@@ -1,10 +1,18 @@
+import 'package:arweave/utils.dart' as utils;
+
 String convertCreditsToLiteralString(BigInt credits) {
   final creditsAsAr = convertWinstonToAr(credits);
   final creditsString = creditsAsAr.toStringAsFixed(4);
 
   return creditsString;
 }
 
+String convertARToLiteralString(BigInt ar) {
```
kunstmusik, line 10: Using substring looks a little awkward, may need rework (from Slack huddle discussion)

kunstmusik: rename to convertWinstonToARLiteralString and rename ar argument to winston

packages/ardrive_uploader/lib/src/ardrive_uploader.dart
```diff
@@ -44,17 +45,32 @@ abstract class ArDriveUploader {
...
+    Arweave? arweave,
+    PstService? pstService,
   }) {
     metadataGenerator ??= ARFSUploadMetadataGenerator(
       tagsGenerator: ARFSTagsGenetator(
         appInfoServices: AppInfoServices(),
       ),
     );
 
+    arweave ??= Arweave();
```
kunstmusik, line 57: Make non-optional and remove default

packages/pst/lib/src/ardrive_contract_oracle.dart
```diff
@@ -31,12 +36,23 @@ class ArDriveContractOracle implements ContractOracle {
...
+              .map((e) async => await _getContractWithRetries(e))
+              .toList());
 
-    return contract;
+      return contract;
+    } catch (e) {
+      debugPrint('Could not read contract state from any of the oracles');
+      if (_fallbackContractOracle == null) {
+        throw const CouldNotReadContractState(
+          reason: 'Max retry attempts reached',
```
kunstmusik, line 50: Needs a better reason due to the code change

packages/pst/lib/src/ardrive_contract_oracle.dart
```diff
@@ -31,12 +36,23 @@ class ArDriveContractOracle implements ContractOracle {
...
-            .map((e) async => await _getContractWithRetries(e))
-            .toList());
+    try {
+      final contract = await getFirstFutureResult<CommunityContractData>(
+          _contractOracles
+              .map((e) async => await _getContractWithRetries(e))
+              .toList());
 
-    return contract;
+      return contract;
```
kunstmusik, line 45: Need to check if contract is empty (i.e., no results). Could add call to fallback contract oracle here, so that try/catch can catch any errors that might happen with it.

packages/pst/lib/src/ardrive_contract_oracle.dart
```diff
@@ -31,12 +36,23 @@ class ArDriveContractOracle implements ContractOracle {
...
+    } catch (e) {
+      debugPrint('Could not read contract state from any of the oracles');
+      if (_fallbackContractOracle == null) {
+        throw const CouldNotReadContractState(
+          reason: 'Max retry attempts reached',
+        );
+      }
+
+      return _fallbackContractOracle!.getCommunityContract();
```
kunstmusik, line 54: Maybe use _getContractWithRetries with the fallbackContractOracle

packages/pst/lib/src/pst.dart
```diff
@@ -33,4 +35,31 @@ class PstService {
...
+  Future<CommunityTip> getCommunityTip(BigInt uploadCost) async {
+    final currentTime = DateTime.now();
+
+    if (_communityTip != null &&
+        _lastFetchedTime != null &&
+        currentTime.difference(_lastFetchedTime!).inMinutes < 30) {
+      return _communityTip!;
+    }
+
+    _communityTip = CommunityTip(
```
kunstmusik, line 48: We should not cache PST holder, but can cache PST fee

kunstmusik: We can cache the list of holders and then recalculate for random holder each time.

lib/turbo/utils/utils.dart
thiagocarvalhodev: @kunstmusik , in the end, the wonky balance was due to a difference in the price of the upload adding the fee, and the upload not using the fee. So, I just changed the method name. 65c607da153f721c84d7c401574aaca3b0a88e4e

packages/ardrive_uploader/lib/src/ardrive_uploader.dart
thiagocarvalhodev: a1cbb70da5acc4d000a9844765abbdb8cbce8525

packages/pst/lib/src/ardrive_contract_oracle.dart
thiagocarvalhodev: a1cbb70da5acc4d000a9844765abbdb8cbce8525

packages/pst/lib/src/ardrive_contract_oracle.dart
thiagocarvalhodev: a2902134301050c13b3bc0adbe4537eb9525591c

packages/pst/lib/src/ardrive_contract_oracle.dart
thiagocarvalhodev: a1cbb70da5acc4d000a9844765abbdb8cbce8525

packages/pst/lib/src/pst.dart
thiagocarvalhodev: a1cbb70da5acc4d000a9844765abbdb8cbce8525

packages/pst/lib/src/ardrive_contract_oracle.dart
```diff
@@ -31,12 +31,27 @@ class ArDriveContractOracle implements ContractOracle {
...
+      return contract;
+    } catch (e) {
+      debugPrint('Could not read contract state from any of the oracles');
+      if (_fallbackContractOracle == null) {
+        throw const CouldNotReadContractState(
+          reason: 'No fallback contract reader provided',
+        );
+      }
+
+      return _getContractWithRetries(_fallbackContractOracle!);
```
kunstmusik, line 53: Could the fallback oracle throw an exception?

packages/pst/lib/src/community_oracle.dart
```diff
@@ -1,38 +1,45 @@
...
 
   CommunityOracle(ContractOracle contractOracle)
       : _contractOracle = contractOracle;
 
   Future<Winston> getCommunityWinstonTip(Winston winstonCost) async {
-    final tipPercentage = await _contractOracle.getTipPercentageFromContract();
+    final contractData = await _getCommunityContractData();
+
+    final CommunityTipPercentage tipPercentage =
+        contractData.settings.fee / 100;
```
kunstmusik, line 25: Suggest to divide by 100.0 to force a double division, since fee is defined as num and could be an integer.

kunstmusik (changes requested):

packages/pst/lib/src/ardrive_contract_oracle.dart
thiagocarvalhodev: Yes. If so, the client should handle it properly.

packages/pst/lib/src/community_oracle.dart
thiagocarvalhodev: I agree. 1b4f08080e04c6fde007187c5d5425ae72e94b69

packages/ardrive_uploader/lib/src/ardrive_uploader.dart
kunstmusik: Does not look addressed as ArDriveUploader still has optional args

kunstmusik (approved):

PR #1444, opened 2023-10-25
PE-4709: Image preview on Shared File Page

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/1l30b6e6ii060


lib/pages/drive_detail/components/fs_entry_preview_widget.dart
```diff
@@ -1028,6 +1027,199 @@ class _FullScreenVideoPlayerWidgetState
...
+      ],
+    );
+  }
+
+  String _getFileNameWithNoExtension() {
+    return widget.filename.substring(0, widget.filename.lastIndexOf('.'));
+  }
+
+  String _getFileExtension() {
+    return widget.contentType
```
thiagocarvalhodev, line 1197: You can use a helper method from `ardrive_io`:  `getFileExtension`

thiagocarvalhodev (changes requested):

lib/pages/drive_detail/components/fs_entry_preview_widget.dart
matibat: If I upload my PNG image (`application/png`) named `image.png.txt` it would return TXT instead of PNG.

matibat: Perhaps I can rename the method to "_getFileExtensionFromContentType"

thiagocarvalhodev: That's a good point. We must rename the `getFileExtension` on `ardrive_ui` for something like: `getFileExtensionFromPath` and we probably want a new method for getting the extension based on content type.

thiagocarvalhodev (approved):
LGTM. We will update the `ardrive_io` helper method in another ticket.

thiagocarvalhodev (approved):

PR #1458, opened 2023-10-31
PE-4910: Release App v2.23.0

thiagocarvalhodev:
### Release Notes
- Allows preview of audio files on the share page.
- Created a direct link to new user onboarding.
- Increases the file size limit for downloads via the share page.


matibat (approved):
LGTM

PR #1460, opened 2023-11-02
PE-4892: refactor the progress update uploading files

thiagocarvalhodev:
## Overview

This ticket will apply a few refactors to improve the upload progress performance. The main points on where the upload performance is impacted:

- Memory usage
- Algorithm for updating the progress
- Saving the files as they are uploaded individually
- UI updates while uploading

See Details & Requirements for more info.

### Saving the files as they are uploaded individually

Currently, since the v2.22.0 (large uploads release) we are batching all the files for then, when the upload finishes, save them on the data database. With that approach, we have some problems:

For large files, the modal will get stuck while saving all the files without any feedback from the user

Large amount of memory usage when saving all files at the same time

Now, we will save the files as each one finalizes. For example, if the user is uploading 1000 files, we will call the database 1000 times, a call for each file as soon as that file finishes the upload.

It fixes an issue mentioned by our user. See notes section.

### Algorithm for updating the progress

Refactor the UploadController and UploadCubit classes to improve the performance when updating the progress. It requires some refactors on the methods we use for updating the progress. For example, instead of iterating through the entire list of files to get the total size, start the total size at zero and increment as we add files to the upload. It’s very technical and the changes can be reviewed on the PR.

### Memory usage

We found a memory leak in our upload process. The stream passed for the Dio client was not being closed (we thought Dio would handle that, but it did not). Now, we are properly closing the stream. 

**Important**: This only happens on TURBO!

### UI updates while uploading

Before the large upload feature, each time we save a file on the database we update the entire drive tree. It’s expensive as we need to regenerate the entire drive tree multiple times while uploading the files. Now, it will update the drive explorer only once when the upload finishes, either with success or failure.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/1gurqhnqemr78


web/index.html
```diff
@@ -19,6 +19,7 @@
     <script defer src="js/arconnect.js"></script>
     <script defer src="js/is_document_focused.js"></script>
     <script defer src="js/StreamSaver.min.js"></script>
+    <script defer src="https://unpkg.com/web-streams-polyfill/dist/polyfill.min.js"></script>
```
karlprieb, line 22: why we're adding polyfills?

packages/ardrive_uploader/lib/src/data_bundler.dart
matibat, line 1: There's two commented print statements here.

packages/ardrive_uploader/lib/src/upload_controller.dart
```diff
@@ -485,57 +225,120 @@ class _UploadController implements UploadController {
...
   }
 
-  @override
-  void addTask(UploadTask task) {
-    tasks[task.id] = task;
+  Future<void> _updateTotalSize(
+      UploadTask task, UploadItem? uploadItem, UploadTask? existingTask) async {
+    if (existingTask?.uploadItem == null && uploadItem != null) {
+      if (task is FileUploadTask) {
+        _totalSize -= await task.file.length;
```
matibat, line 325: I don't fully understand this subtraction here. Any risk of ending up with negative numbers?

karlprieb: Only if _totalSize is smaller than all file lengths combined

thiagocarvalhodev: >Any risk of ending up with negative numbers?
There's not. When adding a new task, we always add the file.lenght, when we have the upload item (the bundle) we update with the real size.

karlprieb (approved):

matibat (approved):
Approved with a non-blocking comment.

PR #1475, opened 2023-11-09
PE-4984: Hide controls and status bar when full-screen

matibat:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5p9nmcib0jbbo


lib/main.dart
```diff
@@ -57,6 +58,8 @@ late PaymentService _turboPayment;
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
+  MobileStatusBar.show();
```
matibat, line 61: **For the reviewer -** this is important for when the system does memory trimming while the app is in full-screen, because it will load again and may possibly start with the status bar being hidden.

karlprieb (approved):

PR #1476, opened 2023-11-13
PE-5010: optimize pst retrieval by excluding verto cache on ar drive app

thiagocarvalhodev:
Currently, we call Verto and Redstone which is not up anymore. We must remove it and use the endpoint from our service in order to optimize the PST retrieval process. This call is only made when we are starting an upload.

We are removing the `VertoContractReader` and `RedstoneContractReader` from the `ArDriveContractOracle`, and we are also changing the endpoint on the class `ARNSContractReader` from `https://dev.arns.app/v1/contract` to `https://api.arns.app/v1/contract`.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/50lmriqut0qc0


packages/pst/test/src/contract_oracle_test.dart
```diff
@@ -78,13 +78,6 @@ void main() {
         myBrokenContractOracle = ContractOracle(myBrokenContractReader);
       });
 
-      test('throws if an empty array is passed', () {
-        expect(
```
thiagocarvalhodev, line 82: We are validating if the `ArDriveContractOracle` is empty when we call the method `_getContractFromOracles` which calls the fallback.

karlprieb (approved):

PR #1478, opened 2023-11-14
PE-5011: fix: missing u tags in direct to network bundles

thiagocarvalhodev:
We've identified an issue where $U tags are not being included in D2N uploads. This update ensures that each D2N transaction is accompanied by the necessary $U tags.

That’s an example of a bundle with the $U tags:
image

### Tests
This PR introduces tests for `metadata_generator.dart`, focusing on:

- Comprehensive testing of the `ARFSTagGenerator` class.
- `ARFSUploadMetadataGenerator` tests to validate metadata creation for files.

Additional tests for other aspects will follow in a subsequent PR.

### Workflow Update:
We've integrated unit tests for each internal package into the testing workflow. See the ticket: https://ardrive.atlassian.net/browse/PE-5018

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/582ptk4dgk3bg

### DIFF
--- a/scripts/build/.last_build_id
+++ b/scripts/build/.last_build_id
@@ -1,0 +1,1 @@
+d4fe773e5af8aea054184ce4dc69442d
--- a/scripts/build/.last_build_id
+++ b/scripts/build/.last_build_id
@@ -1,1 +1,0 @@
-d4fe773e5af8aea054184ce4dc69442d
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,6 @@ jobs:
         run: flutter analyze
 
       - name: Test app
-        run: scr test && scr test-ardrive-uploader && scr test-ardrive-utils && scr test-pst
+        run: scr test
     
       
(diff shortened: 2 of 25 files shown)


scripts/build/.last_build_id
```diff
@@ -0,0 +1 @@
+d4fe773e5af8aea054184ce4dc69442d
```
karlprieb, line 1: why do we have this id now?

karlprieb (commented):
PR is good but I don't think it make sense to create the dumb tests for arfs_test, ardrive_crypto_test and arconnect_test

scripts/build/.last_build_id
thiagocarvalhodev: Removed and added this to the git ignore.

thiagocarvalhodev:
> PR is good but I don't think it make sense to create the dumb tests for arfs_test, ardrive_crypto_test and arconnect_test

I created so it won't break the script I implemented.

karlprieb (approved):

PR #1489, opened 2023-11-22
PE-5087: download modal stays open in android for private files

thiagocarvalhodev:
- use the `decryptTransactionData` instead of `decryptTransactionDataStream` method to decrypt the file on Dart VM (Mobile).

Using the `decryptTransactionDataStream` we get the following exception: 
```
Error saving file: 'package:webcrypto/src/impl_ffi/impl_ffi.aesctr.dart': Failed assertion: line 155 pos 16: 'i == M': is not true.
```

Note that it seems to use the aes ctr even passing the right cipher: **GCM**.

For now, we are using the old method (current in prod) to decrypt the file using GCM.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5c7rp53c6c740

### DIFF
--- a/lib/download/ardrive_downloader.dart
+++ b/lib/download/ardrive_downloader.dart
@@ -69,7 +69,7 @@ class _ArDriveDownloader implements ArDriveDownloader {
       final isPrivateFile =
           fileKey != null && cipher != null && cipherIvString != null;
 
-      if (isPrivateFile && cipherIvString == Cipher.aes256gcm) {
+      if (isPrivateFile && cipher == Cipher.aes256gcm) {
         return _getDartVMGCMDecryptStream(
           cipher,
           cipherIvString,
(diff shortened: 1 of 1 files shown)


kunstmusik:
Test with a dev build from branch on Samsung S22 and found issues:

1. After first time install, I downloaded a single private file. The download modal appeared and then Android asked me to allow Ardrive to save files. I checked yes then saw the dialog show and get stuck at 62% and freeze there. In the logs I saw:

```E/flutter (14744): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Exception: Failed to save file
E/flutter (14744): #0      _ArDriveDownloader.downloadFile.<anonymous closure> (package:ardrive/download/ardrive_downloader.dart:145:31)```

2. I then tried again and again it was stuck at 62%.
3. I tried multi-file download by downloading a folder and no progress occurred and no files downloaded.  

Next I shutdown the app and restarted. 

1. Multi-file download worked.
2. Single-file download went to 100% but dialog was stuck. I checked and the file was in my downloads folder.  Log showed completion to 100% and no errors.

Seems there are issues for first-time use and needing to grant permissions, as well as in general for single-file download dialog still remaining after completion of download.

karlprieb (approved):

PR #1492, opened 2023-11-28
PE-5015: ArConnect logout leftover

matibat:


thiagocarvalhodev (approved):
LGTM! Thanks for fixing it.

PR #1500, opened 2023-11-30
PE-4966: pwa none

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6voi3coq2gsg0


.github/workflows/pr.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
         run: |
           scr setup
           flutter config --enable-web
-          flutter build web --dart-define=environment=development --release
+          flutter build web --dart-define=environment=development --release --pwa-strategy=none
```
karlprieb, line 45: add `--pwa-strategy=none` in all workflows and add `--no-web-resources-cdn` so we can serve all the assets like canvaskit instead of getting them from a cdn

lib/blocs/upload/limits.dart
```diff
@@ -13,7 +13,7 @@ int getBundleSizeLimit(bool isTurbo) =>
     isTurbo ? turboBundleSizeLimit : d2nBundleSizeLimit;
 
 final d2nBundleSizeLimit = const GiB(65).size;
-final turboBundleSizeLimit = const GiB(2).size;
+final turboBundleSizeLimit = const GiB(65).size;
```
karlprieb, line 16: check with @dtfiedler / @fedellen the latest bundle size supported by upload service

.github/workflows/pr.yaml
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1500/commits/69a1c457b1b518cf646ca942c8465961604314c5

thiagocarvalhodev:
These changes are being applied on the #1507

PR #1507, opened 2023-12-04
PE-5167: Fix Tab-Switch Upload Failure in Turbo Web App

thiagocarvalhodev:
This PR addresses a critical bug in the Turbo web application (version v2.22.0), where large file uploads were failing when users switched tabs. The root cause has been identified as the presence of `flutter_service_worker.js`.

Changes:
- Implemented the `--pwa-strategy=none` flag to disable the service worker.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/33k8hk0ajqfn0


thiagocarvalhodev:
We are opening a new one for tracking the changes in the right ticket.

We can check the previously opened PR for these changes: #1500

PR #1531, opened 2023-12-21
PE-5325: feat(sharing intent)

thiagocarvalhodev:
implement sharing intent feature

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/330em6smu8bko


lib/blocs/sync/sync_cubit.dart
```diff
@@ -85,6 +87,14 @@ class SyncCubit extends Cubit<SyncState> {
         _configService = configService,
         _tabVisibility = tabVisibility,
         super(SyncIdle()) {
+    // TODO: verify
```
karlprieb, line 90: verify what?

thiagocarvalhodev: Good catch. 
Everytime we open the app, the sync happens and it kills the sharing modal. This was added to prevent the sync from killing the sharing modal when the app is opened. We must provide a right solution here.

lib/sharing/blocs/sharing_file_bloc.dart
```diff
@@ -0,0 +1,66 @@
...
+part 'sharing_file_state.dart';
+
+class SharingFileBloc extends Bloc<SharingFileEvent, SharingFileState> {
+  List<SharedFile>? files;
+  final ActivityTracker activityTracker;
+
+  SharingFileBloc(
+    this.activityTracker,
+  ) : super(SharingFileInitial()) {
+    // For sharing images coming from outside the app while the app is in the memory
```
karlprieb, line 20: ??

thiagocarvalhodev: I'll remove this comment, but basically we have two ways for sharing files. The first one is when the app is active, and the last one is when the app is closed. 

The methods are:
`getMediaStream`
`getInitialSharing`

karlprieb (approved):

PR #1548, opened 2024-01-09
PE-5394: fix(D2N uploads)

thiagocarvalhodev:
- properly wait for the upload to finish and return the right status
- add the missing dataItem on D2N using bundle strategy

### DIFF
--- a/packages/ardrive_uploader/lib/src/d2n_streamed_upload.dart
+++ b/packages/ardrive_uploader/lib/src/d2n_streamed_upload.dart
@@ -7,7 +7,6 @@ import 'package:flutter/foundation.dart';
 
 class D2NStreamedUpload implements StreamedUpload<UploadItem> {
   UploadAborter? _aborter;
-  StreamedUploadResult? _result;
 
   @override
   Future<StreamedUploadResult> send(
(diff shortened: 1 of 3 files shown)


packages/ardrive_uploader/lib/src/upload_strategy.dart
```diff
@@ -245,6 +246,15 @@ class UploadFileUsingBundleStrategy extends UploadFileStrategy {
 
     final streamedUpload = _streamedUploadFactory.fromUploadType(task.type);
 
+    task = task.copyWith(
+      status: UploadStatus.inProgress,
+      cancelToken: UploadTaskCancelToken(
+        cancel: () => streamedUpload.cancel(task.uploadItem!),
```
matibat, line 252: Doesn't it have to be async? Just asking because cancel is a Future.

packages/ardrive_uploader/lib/src/d2n_streamed_upload.dart
```diff
@@ -43,20 +42,21 @@ class D2NStreamedUpload implements StreamedUpload<UploadItem> {
...
+        debugPrint('D2NStreamedUpload.send: error while uploading');
+        upload.complete(StreamedUploadResult(success: false));
+      }
     });
 
-    await upload.future;
-    return _result!;
+    final result = await upload.future;
+
+    return result;
```
matibat, line 59: We can directly return without awaiting. But either way is fine, just coding style.

matibat (approved):
LGTM - I didn't test, but the code looks good.

PR #1527, opened 2023-12-18
[Draft] ANS-105 Licenses (UDL + CC)

elliotsayes:
Changes:
- Support for ANS-105 licenses: 
  - "LicenseAssertion": a standalone transaction that retroactively applies a license to previously uploaded data 
  - "LicenseComposed": a license that is applied by adding tags to the same transaction as the uploaded data (i.e. dataTx)
- New UX flow for adding licenses to existing files
  - Buttons in file details, file/folder context menu, multi select
  - Configuration for UDL licenses
- New Revision action: `assertLicense`
- New (optional) field in `ArFS` `FileEntity`: `licenseTxId`
- New table in database: `licenses`
  - Updated several queries to join on this table when needed
  - References from `FileEntity`/`FileRevision` table
- New License column in drive view
- Added GQL queries to fetch Licenses (both standalone LicenseAssertions + those composed with the DataTx)
- Added sync logic to find licenses referenced in `file_revisions` table & populate the `licenses` table
- Updated the uploader to attach a file's existing license to the new dataTxs when doing `RevisionAction.uploadNewVersion` & update the `FileEntity` to reference the new dataTxId as the new license.
  - This will also make it easier once we add the UX flow to add a license during initial upload :)
- Integrated license fetching into `SharedFileCubit`
  - Bug fix: FsEntryInfoCubit was attempting to query the drive for the file entry/revision even when it might not exist causing an error. It now checks if it is a shared file and falls back to the data provided by `SharedFileCubit`
 
Notes:
- Licenses should not be added to pinned files
- Licenses not referenced in an ArFS File Entity will not be synced

<details>
<summary>Tasks</summary>
Working:
- `Add license` UI / logic
- Creating License Assertions
- License Assertion Entity
- License Assertion Model
- Additional UI buttons to open `Add license` UI
- Handling FsEntry folders
- Handling syncing
- Handling pinned files
- Custom add license button in file details
- UDL form styling
- Custom license popover in file details
- DB migration
- License column in folder view?
- Handling uploading new license when `RevisionAction.createNewVersion` (WIP: testing)
- Remove unneeded queries/conversions

Todo:
- ???
</details>

Issues:
- `build_runner` can sometimes fail, requiring clean and rerun
- `DriveDetailCubit.refreshDriveDataTable()` does not seem to work (same issue with `FsEntryRenameForm` on `dev`)
- Error loading `FsEntryInfoCubit` for shared files (same issue on `dev`)
  - Screenshot 2024-01-11 at 11 17 56 AM

PR #1551, opened 2024-01-12
PE-5412: ANS-105 Licenses (UDL + CC)

thiagocarvalhodev:
https://github.com/ardriveapp/ardrive-web/pull/1527

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6f237nm9oc7do


thiagocarvalhodev:
@elliotsayes , you can use this branch for your PRs! Feel free to change the description PR as needed, or anything you need let us know!

elliotsayes:
Fix here: https://github.com/ardriveapp/ardrive-web/pull/1552

elliotsayes:
More :) https://github.com/ardriveapp/ardrive-web/pull/1560

elliotsayes:
🤞 https://github.com/ardriveapp/ardrive-web/pull/1562

atticusofsparta:
License doesnt update on first add, need to add table refresh on workflow completion.

atticusofsparta:
In license column, UDL is missing the icon that goes with it, see the screenshot from figma below
image

lib/components/fs_entry_license_form.dart
```diff
@@ -0,0 +1,805 @@
...
+              width: kMediumDialogWidth,
+              contentPadding: EdgeInsets.zero,
+              content: Column(
+                mainAxisSize: MainAxisSize.max,
+                crossAxisAlignment: CrossAxisAlignment.stretch,
+                children: [
+                  Container(
+                    padding: const EdgeInsets.only(left: 16, right: 16),
+                    width: double.infinity,
+                    height: 77,
```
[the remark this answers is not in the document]
elliotsayes: Good point. I've used `fs_entry_move_form` / `create_manifest_form` as a reference here

lib/components/fs_entry_license_form.dart
```diff
@@ -0,0 +1,805 @@
...
+              height: 350,
+              width: kMediumDialogWidth,
+              contentPadding: EdgeInsets.zero,
+              content: Column(
+                mainAxisSize: MainAxisSize.max,
+                crossAxisAlignment: CrossAxisAlignment.stretch,
+                children: [
+                  Container(
+                    padding: const EdgeInsets.only(left: 16, right: 16),
+                    width: double.infinity,
```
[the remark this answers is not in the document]
elliotsayes: I'm not sure either. Using other components as a reference here.

lib/entities/license_assertion.dart
```diff
@@ -0,0 +1,85 @@
...
+      LicenseTag.appName: LicenseTag.appNameLicenseAssertion,
+      LicenseTag.originalTxId: dataTxId,
+      LicenseTag.licenseDefinitionTxId: licenseDefinitionTxId,
+    };
+
+    baseTags.forEach((key, value) {
+      licenseAssertionDataItem.addTag(key, value);
+    });
+
+    additionalTags.forEach((key, value) {
```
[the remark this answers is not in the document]
elliotsayes: Good catch, also present in LicenseComposedEntity

lib/services/license/licenses/udl.dart
```diff
@@ -0,0 +1,133 @@
...
+  UdlDerivation.allowedWithLicensePassthrough:
+      'Allowed-With-License-Passthrough',
+};
+
+class UdlTags {
+  static const String licenseFee = 'License-Fee';
+  static const String currency = 'Currency';
+  static const String commercialUse = 'Commercial-Use';
+  static const String derivations = 'Derivation';
+}
```
atticusofsparta, line 55: I think we may need a Payment-Address tag here, considering these tx's might be deployed via turbo's bundler. From the UDL license, Section 3.g : 

"If no ”Payment-Address” tag is present, the following applies: If a smart contract is attached to the Media, the Payment Address is the Addresses defined within the smart contract, otherwise the Payment Address is the Address which posted the media to the Network."

atticusofsparta: Having the payment address ensures the user gets paid with no misinterpretations.

elliotsayes:
Was there a comment here about adding the license Icon or am I going crazy? 
Screenshot 2024-01-31 at 10 45 17 AM

elliotsayes:
> License doesnt update on first add, need to add table refresh on workflow completion.

Table refresh is there but broken, bug is fixed by another PR from @thiagocarvalhodev 
Screenshot 2024-01-31 at 10 52 16 AM

elliotsayes:
Fixes in PR #1591

lib/services/license/licenses/udl.dart
elliotsayes: You're right, the wording makes it seem like payment might go to the bundling service. I've gone with your suggestion here

atticusofsparta:
@elliotsayes Hey there, yeah there was a comment re:icons that I deleted, I'm talking to the team to discuss what to do on PR https://github.com/ardriveapp/ardrive-web/pull/1591. Sorry for the miscommunication 😬

atticusofsparta:
@elliotsayes BTW re the refresh, i was confused why it wasnt working - looks like the state that was getting emitted was incorrect/overlapping - fixed in https://github.com/ardriveapp/ardrive-web/pull/1551/commits/ffdf5446c4c177468c8a15a807f4504b8dc5be11 
image

lib/entities/license_assertion.dart
```diff
@@ -0,0 +1,96 @@
...
+  LicenseTag.originalTxId,
+  LicenseTag.licenseDefinitionTxId,
+];
+
+final appInfo = AppInfoServices().appInfo;
+final ardriveTags = {
+  EntityTag.appVersion: appInfo.version,
+  EntityTag.appPlatform: appInfo.platform,
+  EntityTag.appName: appInfo.appName,
+  EntityTag.unixTime:
```
elliotsayes, line 25: Would this fix unixTime to when the app is launched?

lib/entities/license_assertion.dart
```diff
@@ -0,0 +1,96 @@
...
+  LicenseTag.appName,
+  LicenseTag.originalTxId,
+  LicenseTag.licenseDefinitionTxId,
+];
+
+final appInfo = AppInfoServices().appInfo;
+final ardriveTags = {
+  EntityTag.appVersion: appInfo.version,
+  EntityTag.appPlatform: appInfo.platform,
+  EntityTag.appName: appInfo.appName,
```
elliotsayes, line 24: Are we okay with multiple `App-Name`s? Does this play nice with GQL etc

lib/entities/license_assertion.dart
atticusofsparta: good question, i dont think so but i will double check.

lib/entities/license_assertion.dart
atticusofsparta: Good question, its *supposes to be* okay with gql. Currently there is a bug on the gateway behind arweave.net that returns multiple items on transactions with duplicate tags - client side code should filter for this.

🤔 

Now that you mention it, im uncertain how the app will handle this. It only happens on larger multipage gql responses, will need to test.

atticusofsparta: GQL seems fine, likely because of the way we store entities via a generated cache key - itll just put the duplicate into the same key-value as the duplicate entity, which is okay since all the data is the same.

lib/entities/license_assertion.dart
atticusofsparta: You were right, they were singletons, move to the function where they are used to instantiate on call, fixed now. Good catch!

atticusofsparta (dismissed):

atticusofsparta (approved):

atticusofsparta (approved):

PR #1552, opened 2024-01-13
Fix licensing config/review screen on small displays

elliotsayes:
Add scrollbar to License configuring/reviewing screens. This fixes visibility of back/next buttons on mobile chrome (or very tiny android screens?)


thiagocarvalhodev:
Merging in order to get the deployed branch

PR #1553, opened 2024-01-15
PE-5431: show appropriate icon on upload page for file with extension written in uppercase

thiagocarvalhodev:
- set content type to lower case for getting the right icon

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/59otoa42nu1g8


kunstmusik (approved):
Final release requires depending on released version of ardrive_io

PR #1554, opened 2024-01-15
PE-5437: add an option to the user update the file extension when renaming a file

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/2805j3fc6m12g


lib/components/fs_entry_rename_form.dart
```diff
@@ -105,6 +106,44 @@ class _FsEntryRenameFormState extends State<FsEntryRenameForm> {
...
+              context,
+              content: ArDriveStandardModal(
+                title: 'Do you want to change the file extension?',
+                description: 'The file extension will be changed from '
+                    '${state.previousExtension} to ${getFileExtension(name: state.entityName, contentType: state.newExtension)}',
+                actions: [
+                  ModalAction(
+                    action: () {
+                      Navigator.of(context).pop();
+                      Navigator.of(context).pop();
```
kunstmusik, line 120: Double-checking, do we need two pop's here?

lib/components/fs_entry_rename_form.dart
```diff
@@ -105,6 +106,44 @@ class _FsEntryRenameFormState extends State<FsEntryRenameForm> {
...
+                    },
+                    title: appLocalizationsOf(context).cancelEmphasized,
+                  ),
+                  ModalAction(
+                    action: () {
+                      context.read<FsEntryRenameCubit>().dontVerifyExtension();
+                      context
+                          .read<FsEntryRenameCubit>()
+                          .submit(newName: _nameController.text);
+                      Navigator.of(context).pop();
```
kunstmusik, line 130: Should this be submit().then() like in the following ModalAction?

lib/components/fs_entry_rename_form.dart
thiagocarvalhodev: One pop is for the modal that asks the user for updating the extension, and the other one is the regular rename modal

lib/components/fs_entry_rename_form.dart
thiagocarvalhodev: Good point. None of those needs to wait the submit.

lib/blocs/fs_entry_rename/fs_entry_rename_cubit.dart
```diff
@@ -161,6 +190,29 @@ class FsEntryRenameCubit extends Cubit<FsEntryRenameState> {
     }
   }
 
+  bool verifyExtensionAndReturnIfExtensionWasUpdated(
```
kunstmusik, line 193: This is one a case where I think having this in a separate function makes the code more difficult to read and understand, since it is mutating _newFileExtension and also emitting an event. Would prefer to see this in-lined above and this function removed.

lib/components/fs_entry_rename_form.dart
```diff
@@ -105,6 +106,42 @@ class _FsEntryRenameFormState extends State<FsEntryRenameForm> {
...
             );
+          } else if (state is UpdatingEntityExtension) {
+            showArDriveDialog(
+              context,
+              content: ArDriveStandardModal(
+                title: 'Do you want to change the file extension?',
+                description: 'The file extension will be changed from '
+                    '${state.previousExtension} to ${getFileExtension(name: state.entityName, contentType: state.newExtension)}',
+                actions: [
+                  ModalAction(
```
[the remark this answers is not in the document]
atticusofsparta: opened a PR against this to fix the navigation and remove the button https://github.com/ardriveapp/ardrive-web/pull/1575

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "editor.codeActionsOnSave": {
-    "source.organizeImports": true
+    "source.organizeImports": "explicit"
```
kunstmusik, line 3: Prefer to revert this back to true. Thoughts on this @thiagocarvalhodev ?

kunstmusik (approved):

atticusofsparta (approved):

PR #1559, opened 2024-01-15
PE-5387: chunk uploads to turbo

karlprieb:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/71ebjp23q03m8


lib/blocs/upload/limits.dart
```diff
@@ -1,20 +1,14 @@
...
-
-final mobilePrivateFileSizeLimit = const GiB(10).size;
+const maxBundleDataItemCount = 500;
+const maxFilesPerBundle = maxBundleDataItemCount ~/ 2;
+const maxFilesSizePerBundleUsingTurbo = 1;
 
-final publicFileSafeSizeLimit = const GiB(5).size;
-final nonChromeBrowserUploadSafeLimitUsingTurbo = const MiB(500).size;
+final d2nBundleSizeLimit = const GiB(65).size;
+final turboBundleSizeLimit = const GiB(10).size;
```
arielmelendez, line 11: It would be nice if Turbo could advertise this number so that it wouldn't require client updates every time we bump it.

lib/blocs/upload/upload_cubit.dart
```diff
@@ -480,13 +479,9 @@ class UploadCubit extends Cubit<UploadState> {
 
     if (_uploadMethod == UploadMethod.turbo) {
       await _verifyIfUploadContainsLargeFilesUsingTurbo();
-      if ((_containsLargeTurboUpload ?? false) &&
-          !hasEmittedWarning &&
-          kIsWeb &&
-          !await AppPlatform.isChrome()) {
+      if ((_containsLargeTurboUpload ?? false) && !hasEmittedWarning) {
```
arielmelendez, line 482: So is this now updated because the `web && chrome` guard was to avoid problems with large, non-chunked uploads to Turbo with Chrome?

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
```diff
@@ -0,0 +1,166 @@
+import 'dart:async';
+import 'dart:typed_data';
+
+import 'package:ardrive_utils/ardrive_utils.dart';
+import 'package:arweave/arweave.dart';
+import 'package:dio/dio.dart';
+import 'package:retry/retry.dart';
+
+class TurboUploadService<Response> {
```
arielmelendez, line 9: Why is this generic with a preset specialization if it's not implementing/extending anything?

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
```diff
@@ -0,0 +1,166 @@
...
+
+  final Uri turboUploadUri;
+  final retry = RetryOptions(maxAttempts: 8);
+  final CancelToken _cancelToken = CancelToken();
+
+  Future<Response> post({
+    required DataItemResult dataItem,
+    required Wallet wallet,
+    Function(double)? onSendProgress,
+    required int size,
```
arielmelendez, line 22: Nits: could be good to add specificity in the name around the context and the units, e.g. `rawDataItemByteCount`, which could be contrasted with `payloadByteCount`.

lib/blocs/upload/limits.dart
thiagocarvalhodev: Totally agree

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
thiagocarvalhodev: Previously, we had this class:

```dart
abstract class TurboUploadService<T> {
  Future<T> postStream({
    required DataItemResult dataItem,
    required Wallet wallet,
    Function(double p1)? onSendProgress,
    required int size,
    required Map<String, dynamic> headers,
  });

  Future<void> cancel();
}
``` 
Which we could infer a type for the response. Now, that we don't implement this, we can remove it.

lib/blocs/upload/upload_cubit.dart
karlprieb: Yes, now all platforms can follow the same rules

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
karlprieb: Good call! I'll refactor this.

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
arielmelendez: Also it looks like a DataItemResult (by the way I'm not a fan of that class name at all - it implies some completed workflow with a data item but it's more of a metadata-bearing class combined with some decorator functions for getting streams to the data) already has size information for both the data item AND its payload in it.

thiagocarvalhodev:
When we cancel the upload, it does not really cancels. The modal goes away, but we can see on the network tab that the upload still going.

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
karlprieb: You're right!

karlprieb:
@thiagocarvalhodev `cancelToken` now is being passed to `Dio` and it cancels all in progress requests. The uploader itself stop sending new requests when it gets cancelled as well.

@arielmelendez @thiagocarvalhodev There's more 6 file changes, but I'm only changing all those `debugPrint` to use our `logger`.

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
karlprieb: Addressed in my latest commit

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
karlprieb: Addressed in my latest commit

lib/blocs/upload/limits.dart
karlprieb: I discussed that with @fedellen a while ago and he also agrees that we can add this information in `/info` endpoint

lib/blocs/upload/upload_file_checker.dart
```diff
@@ -1,37 +1,51 @@
...
-      {required List<UploadFile> files}) async {
-    final filesAbovePrivateLimit = <String>[];
+    return filesAboveSizeLimit;
+  }
 
+  Future<bool> _hasFileAboveLimit({
+    required List<UploadFile> files,
+    required int limit,
+  }) async {
     for (final file in files) {
```
kunstmusik, line 42: Could replace this with list.any()

lib/blocs/upload/upload_file_checker.dart
```diff
@@ -1,37 +1,51 @@
...
+    required List<UploadFile> files,
+  }) =>
+      _hasFileAboveLimit(files: files, limit: _fileSizeWarning);
+
+  Future<List<String>> getFilesAboveSizeLimit({
+    required List<UploadFile> files,
+  }) async {
+    final filesAboveSizeLimit = <String>[];
+
     for (final file in files) {
```
kunstmusik, line 28: Could replace with a list.fold()

lib/blocs/upload/upload_cubit.dart
```diff
@@ -101,15 +100,15 @@ class UploadCubit extends Cubit<UploadState> {
     required ProfileCubit profileCubit,
     required DriveDao driveDao,
     required PstService pst,
-    required UploadFileChecker uploadFileChecker,
+    required UploadFileSizeChecker uploadFileChecker,
```
thiagocarvalhodev, line 103:
```suggestion
    required UploadFileSizeChecker uploadFileSizeChecker,
```

lib/blocs/upload/upload_cubit.dart
```diff
@@ -480,13 +479,9 @@ class UploadCubit extends Cubit<UploadState> {
...
     if (_uploadMethod == UploadMethod.turbo) {
       await _verifyIfUploadContainsLargeFilesUsingTurbo();
-      if ((_containsLargeTurboUpload ?? false) &&
-          !hasEmittedWarning &&
-          kIsWeb &&
-          !await AppPlatform.isChrome()) {
+      if ((_containsLargeTurboUpload ?? false) && !hasEmittedWarning) {
         emit(
           UploadShowingWarning(
-            reason: UploadWarningReason.fileTooLargeOnNonChromeBrowser,
```
thiagocarvalhodev, line 489: Removing this reason was intended?

lib/blocs/upload/upload_cubit.dart
```diff
@@ -756,11 +751,9 @@ class UploadCubit extends Cubit<UploadState> {
     if (_containsLargeTurboUpload == null) {
       _containsLargeTurboUpload = false;
 
-      for (var file in files) {
-        if (await file.ioFile.length >= largeFileUploadSizeThreshold) {
-          _containsLargeTurboUpload = true;
-          break;
-        }
+      if (await _uploadFileSizeChecker.hasFileAboveSizeLimit(files: files)) {
```
thiagocarvalhodev, line 754: I know this method name was defined before, but FileAboveSizeLimit does not tell about which limit. Can we change to something: 

```Dart
_uploadFileSizeChecker.hasFileSizeAboveUploadLimit()
```

```Dart
_uploadFileSizeChecker.hasFileSizeAboveUploadWarningLimit()
```

Something like this

lib/blocs/upload/upload_cubit.dart
```diff
@@ -855,8 +848,8 @@ class UploadCubit extends Cubit<UploadState> {
 
   Future<void> skipLargeFilesAndCheckForConflicts() async {
     emit(UploadPreparationInProgress());
-    final List<String> filesToSkip = await _uploadFileChecker
-        .checkAndReturnFilesAbovePrivateLimit(files: files);
+    final List<String> filesToSkip =
+        await _uploadFileSizeChecker.getFilesAboveSizeLimit(files: files);
```
thiagocarvalhodev, line 852: Same comment above about naming this method without providing full context

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
```diff
@@ -0,0 +1,182 @@
...
+import 'package:retry/retry.dart';
+import 'package:ardrive_uploader/src/utils/logger.dart';
+
+class TurboUploadService {
+  TurboUploadService({
+    required this.turboUploadUri,
+  });
+
+  final Uri turboUploadUri;
+  final r = RetryOptions(maxAttempts: 8);
```
thiagocarvalhodev, line 16:
```suggestion
  final retryOptions = RetryOptions(maxAttempts: 8);
```

lib/blocs/upload/upload_cubit.dart
karlprieb: Yes, there's no more different reason for UploadShowingWarning as the only reason is file too large

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
karlprieb: I thought about that, but now every retry call would be `retryOptions.retry`.
What do you think?

lib/blocs/upload/upload_file_checker.dart
kunstmusik: Ignore, issues with async will make this more complicated that iteration.

lib/blocs/upload/upload_file_checker.dart
kunstmusik: Ignore, issues with async will make this more complicated that iteration.

packages/ardrive_uploader/lib/src/turbo_upload_service.dart
thiagocarvalhodev: Got it. I don't have too much preferences on it. I'll resolve the comment

lib/blocs/upload/upload_cubit.dart
karlprieb: fixed [`e604581` (#1559)](https://github.com/ardriveapp/ardrive-web/pull/1559/commits/e6045818c8fafbb86c35f1d563e9ca78e8378a36)

thiagocarvalhodev (dismissed):
LGTM!

PR #1577, opened 2024-01-28
PE-5509: fix conceal items bug

atticusofsparta:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/40e089kk24ie0


thiagocarvalhodev (approved):
LGTM

atticusofsparta:
Found another edge case. If you have a drive with no hidden files and click "reveal hidden files" it hides the files that are not hidden.

If you have a drive with both, it works properly - only revealing the hidden files and not hiding the non-hidden files.

PR #1585, opened 2024-01-30
PE-5534: hidden file movement possible

thiagocarvalhodev:
dont move hidden files

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/0mknq6mfpv460

### DIFF
--- a/lib/pages/drive_detail/components/drive_detail_data_list.dart
+++ b/lib/pages/drive_detail/components/drive_detail_data_list.dart
@@ -3,6 +3,7 @@ part of '../drive_detail_page.dart';
 Widget _buildDataList(
   BuildContext context,
   DriveDetailLoadSuccess state,
+  Widget emptyState,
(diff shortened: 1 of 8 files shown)


lib/pages/drive_detail/components/drive_detail_data_list.dart
```diff
@@ -136,6 +139,10 @@ Widget _buildDataListContent(
     filteredItems = items.where((item) => !item.isHidden).toList();
   }
 
+  if (filteredItems.isEmpty) {
+    return emptyState;
```
atticusofsparta, line 143: This is good but need to add it to the mobile view as well.

atticusofsparta (changes requested):
Need to update the mobile view for this as well.

PR #1595, opened 2024-01-31
PE-5495: add turbo chunk uploader

karlprieb:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/1sbrcfn5s09ko


lib/blocs/upload/upload_cubit.dart
```diff
@@ -598,13 +599,27 @@ class UploadCubit extends Cubit<UploadState> {
 
     _activityTracker.setUploading(true);
 
+    final uploadType =
+        _uploadMethod == UploadMethod.ar ? UploadType.d2n : UploadType.turbo;
+
+    late TurboUploadType turboUploadType;
+    if (_containsLargeTurboUpload ?? false) {
```
atticusofsparta, line 606: nit:
```suggestion
    if (_containsLargeTurboUpload) {
```

atticusofsparta:
getting 404's on transaction finalization 
image

lib/blocs/upload/upload_cubit.dart
thiagocarvalhodev: In Dart we can't do it. An if statement must have a non-nullable boolean.

atticusofsparta:
Got the above issue on both prod and dev, talked with @karlprieb and we are lowering the max to 1gb which worked on both

thiagocarvalhodev (approved):

PR #1629, opened 2024-02-15
PE-5669

thiagocarvalhodev:


packages/ardrive_uploader/lib/src/turbo_upload_service.dart
```diff
@@ -97,6 +100,7 @@ class TurboUploadService {
           }
           return response;
         }, onError: (error) {
+          onSendProgressTimer?.cancel();
```
thiagocarvalhodev, line 103: Here is the main change: the timer was not being canceled

PR #1672, opened 2024-03-20
PE-5845: improve sync and drive explorer navigation performance by removing the usage of paths on folders and files

thiagocarvalhodev:
No features are being added on this PR. The main point is the improvements on the sync time and drive explorer navigation.

### Refactor Sync Process:

The generatePaths step, responsible for generating paths for folders and files, will be removed from the sync process. This method, as detailed in the provided code snippet, performed extensive computations to manage folder and file paths for navigation and manifest creation, which proved to be inefficient.

### Navigation Changes:

Replace path-based navigation with folderId-based navigation within the application. This change necessitates adjustments in how breadcrumbs and manifests are generated, ensuring that the app's navigation remains user-friendly and efficient without relying on the previous path system.

#### Methodology
The performance tests were conducted to compare the efficiency of the ArDrive application between the current production environment and a new branch containing proposed performance improvements. The following methodology was applied to ensure the accuracy and consistency of the tests:
- Test Environment Setup: Each test was conducted with only the ArDrive app tab open in the browser to prevent interference from other tabs or applications.
- Database Preparation: Before each test, the local database was cleared to ensure that no residual data could affect the sync times. This step guarantees that each test starts from the same baseline.
- Data Consistency: The same wallet, containing identical data, was used across all tests to ensure that the sync process was consistent and comparable between tests.
- Sync Process: The drives were synced using the "Snapshots" feature of the ArDrive product, a standardized method to ensure that each drive's state was accurately captured and restored.
- Repetition and Averaging: Each sync operation was performed multiple times for each browser in both the production and new branch environments. The average sync time was then calculated to provide a consistent basis for comparison.

The data reveals significant performance improvements in the new branch across all browsers. Chrome showed a 72.72% improvement, Firefox demonstrated an 80.20% improvement, and Safari exhibited a 77.86% improvement in synchronization times.
- Chrome: The average sync time decreased from 123,023ms in production to 33,573ms in the new branch, indicating a substantial improvement.
- Firefox: The average sync time decreased from 207,138ms in production to 41,014ms in the new branch.
- Safari: The average sync time decreased from 113,345ms in production to 25,090ms in the new branch.

image

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/7j4un71qcrddo


test/models/daos/drive_dao_test.dart
```diff
@@ -173,26 +172,6 @@ void main() {
         expect(file.path, equals(expectedTreeResults[i][1]));
       }
     });
-
```
kunstmusik, line 176: Should this unused test be removed? If not, should probably get fixed and re-enabled.

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -85,7 +88,7 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
   }
 
   void openFolder({
-    required String path,
+    String? folderId,
```
kunstmusik, line 91: Why is this an optional String? It's not clear why this should accept a null here.

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -193,6 +197,25 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
             isOwner: isDriveOwner(_auth, drive.ownerAddress),
           );
 
+          final List<BreadCrumbRowInfo> pathSegments = [];
+
+          if (folderContents.folder.id != drive.rootFolderId) {
+            await _getBreadcrumbsForFolder(
```
kunstmusik, line 203: Would prefer _getBreadcrumbsForFolder to handle a passed-in folder rather than see all this here. The function could be refactored to 1. create the pathSegments List and 2. do a while loop to keep adding to it while there is a parent folder for the current folder. Code here should be as simple as:

```
final pathSegments = _getBreadCrumbsForFolder(folderId, parentFolderId, rootFolderId, driveId);
```

lib/blocs/drive_detail/drive_detail_cubit.dart
```diff
@@ -237,6 +266,45 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
     }
   }
```
kunstmusik, line 268: This could be made a little simpler using using a while-loop and a 'currentFolder' variable. Also could instantiate the list here and return, rather than have something passed-in.

lib/sync/domain/repositories/sync_repository.dart
```diff
@@ -1149,111 +1114,112 @@ Future<Map<FolderID, GhostFolder>> _generateFsEntryPaths({
...
-
-        await addMissingFolder(
-          staleOrphanFile.parentFolderId.value,
-        );
-      }
-    }
-  }
-  return ghostFolders;
+  return {};
+  // final staleFolderTree = <FolderNode>[];
```
kunstmusik, line 1119: remove dead code

lib/sync/domain/repositories/sync_repository.dart
```diff
@@ -1149,111 +1114,112 @@ Future<Map<FolderID, GhostFolder>> _generateFsEntryPaths({
...
+  //       .updateFileById(driveId, staleFileId)
+  //       .write(FileEntriesCompanion(path: Value(filePath)));
+  // }
+
+  // for (final staleFolder in node.subfolders) {
+  //   await updateFolderTree(staleFolder, folderPath);
+  // }
 }
 
+// for (final treeRoot in staleFolderTree) {
```
kunstmusik, line 1172: remove dead code

lib/blocs/drive_detail/drive_detail_cubit.dart
thiagocarvalhodev: Agreed.

lib/blocs/drive_detail/drive_detail_cubit.dart
thiagocarvalhodev: If we pass a null variable here, it will open the root folder. That's the current behavior of the `watchFolderContents`.

lib/blocs/drive_detail/drive_detail_cubit.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1672/commits/daefceb6101aac5e1b79521585273d921f9a8138

test/models/daos/drive_dao_test.dart
thiagocarvalhodev: The `getRecursiveFiles` does not have the `maxDepth`, I guess that was the reason this test was commented.

We have this test in that file:

```dart
 test('getRecursiveFiles returns all the correct files', () async {
      final treeRoot = await driveDao.getFolderTree(driveId, rootFolderId);
      final filesInFolderTree = treeRoot.getRecursiveFiles();

      expect(filesInFolderTree.length, equals(10));
      for (var i = 0; i < filesInFolderTree.length; i++) {
        final file = filesInFolderTree[i];
        expect(file.id, equals(expectedTreeResults[i][0]));
        expect(file.path, equals(expectedTreeResults[i][1]));
      }
    });
 ```
 As we don't use the maxDepth, that test is sufficient.
 
That's the method implementation: 
image

lib/sync/domain/repositories/sync_repository.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1672/commits/15dd6bd6aafd3c812ec1fd6ec98e01b922192c47

lib/sync/domain/repositories/sync_repository.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1672/commits/15dd6bd6aafd3c812ec1fd6ec98e01b922192c47

lib/blocs/drive_detail/drive_detail_cubit.dart
kunstmusik: I still prefer the prior implementation for required path, as it seems cleaner, but I'd feel better with this if documentation was added to the openFolder() method that explained the behavior when folderId is null. I'll go ahead and approve the PR and leave it to you if you want to add that.

kunstmusik (dismissed):

karlprieb (dismissed):

karlprieb (approved):

PR #1702, opened 2024-04-08
PE-5938: update create wallet loading state UI

thiagocarvalhodev:
add lottie for drive animation

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/7grheggjsrv8g


kunstmusik (changes requested):
We should delete the animated gif version of the loading plates

kunstmusik (approved):

PR #1703, opened 2024-04-10
PE-5864: bento box

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/4vri3nsn3rlm0


lib/authentication/login/views/tiles/tiles_view.dart
```diff
@@ -7,8 +17,882 @@ class TilesView extends StatelessWidget {
...
+              ],
+            ),
+          ),
+        ),
+      ],
+    );
+  }
+}
+
+class _MilitareGradeEncryption extends StatelessWidget {
```
kunstmusik, line 159: Could you change to _MilitaryGradeEncryption ?

lib/authentication/login/views/tiles/tiles_view.dart
```diff
@@ -7,8 +17,882 @@ class TilesView extends StatelessWidget {
...
+        },
+        child: Container(
+          decoration: BoxDecoration(
+            border: Border.all(
+              color: colorTokens.strokeLow.withOpacity(0.08),
+              width: 1,
+            ),
+            color: colorTokens.containerL0,
+            borderRadius: const BorderRadius.all(
+              Radius.circular(8.5),
```
kunstmusik, line 383: Elsewhere there is use of radius 8, but here it is 8.5. Seems like this should be changed to 8 and also on line 394 below

lib/misc/resources.dart
```diff
@@ -105,6 +106,33 @@ class Login {
 
   final String metamask = 'assets/images/login/metamask_logo_flat.svg';
   final String walletUpload = 'assets/images/login/wallet_upload.svg';
+
+  final String particleSpace = 'assets/images/login/particle_space.png';
```
kunstmusik, line 110: These files look related to bento box; the keys here could be moved to BentoBox, and the image files for particle and data storage look like they could be moved to the bento_box folder and the values updated here.

kunstmusik (approved):

kunstmusik (approved):

PR #1710, opened 2024-04-15
PE-5955: flutter upgrade plus lint fixes

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/62mbmcniokhso


lib/components/details_panel.dart
```diff
@@ -1206,12 +1207,11 @@ class _CopyButtonState extends State<CopyButton> {
 
 class _DownloadOrPreview extends StatelessWidget {
   const _DownloadOrPreview({
-    Key? key,
```
kunstmusik, line 1209: should this be super.key ?

kunstmusik (changes requested):

lib/components/details_panel.dart
thiagocarvalhodev: `_DownloadOrPreview` is a private class, and we don't use the `key` anywhere, so we have a lint rule, `unused_element`, that prevents us from adding an extra parameter that we're not actually using.

kunstmusik (approved):

PR #1729, opened 2024-05-01
PE-5883: search feature

thiagocarvalhodev:


lib/search/search_modal.dart
```diff
@@ -2,317 +2,352 @@ import 'package:ardrive/authentication/components/login_modal.dart';
...
+                  state.results[index],
+                  typography,
+                  colorTokens,
+                ),
+              ),
+            );
+          } else if (state is SearchQueryEmpty) {
+            return Center(
+              child: Text(
+                'Please search for files',
```
kunstmusik, line 142: This seems like a strange message to show. Suggest something like "Enter text to search for files.".

lib/search/search_modal.dart
```diff
@@ -2,317 +2,352 @@ import 'package:ardrive/authentication/components/login_modal.dart';
...
+      ],
+    );
+  }
+
+  void _handleNavigation(BuildContext context, SearchResult searchResult) {
+    if (searchResult.result is FileEntry) {
+      final file = DriveDataTableItemMapper.fromFileEntryForSearchModal(
+        searchResult.result as FileEntry,
+      );
+      Future.delayed(const Duration(milliseconds: 300)).then((value) async {
```
kunstmusik, line 302: I don't know why there's an initial 300 ms delay? Also, I think it's not good to have these chained like this with timeouts which seems unreliable. Would rather see the cubit have an openFolderAndSelect() method added so that timing could be better controlled.

kunstmusik: Alternatively, why is this delay necessary? Are there problems with just calling these back-to-back?

lib/search/search_modal.dart
```diff
@@ -2,317 +2,352 @@ import 'package:ardrive/authentication/components/login_modal.dart';
...
+    } else if (searchResult.result is Drive) {
+      context
+          .read<DrivesCubit>()
+          .selectDrive((searchResult.result as Drive).id);
+      Navigator.of(context).pop();
+    }
+  }
+
+  void _handleTap(BuildContext context, SearchResult searchResult) {
+    if (searchResult.result is FileEntry) {
```
kunstmusik, line 333: This feels odd in the app. I would expect tapping the FileEntry row to navigate to the File, and only have the download icon on the right-side of the row be for downloading, due to the right navigation arrow being left of the download icon.

lib/search/search_modal.dart
thiagocarvalhodev: The first delay is necessary to wait for the 'Pop' animation of the modal. The second one, is needed because we need to wait until the selected folder is mounted and then automatically select the item

lib/search/search_modal.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1729/commits/abde480531365ca58047b4888f90658d4c850df0

lib/search/search_modal.dart
kunstmusik: Hmm, I still don't like that we're waiting estimated times vs. using something more robust. Would like to see this addressed at some point, but can live with it for this PR. (Please file a task to revisit.)

kunstmusik (approved):

lib/models/daos/drive_dao/drive_dao.dart
```diff
@@ -644,12 +654,16 @@ class DriveDao extends DatabaseAccessor<Database> with _$DriveDaoMixin {
 
 class SearchResult<T> {
   final T result;
-  final FolderEntry? folder;
+  final FolderEntry? parentFolder;
   final Drive drive;
+  final bool? isRootFolder;
```
kunstmusik, line 659: This isRootFolder was added but I don't see it used? Should remove it if not used.

kunstmusik (approved):

PR #1739, opened 2024-05-10
PE-6115: feat(upload manifest with turbo)

thiagocarvalhodev:
- implements the CreateManifestRepository to abstract the logic on cubit
- Refactored the CreateManifestForm widget
- enhanced api on FileRepository
- enhanced api on FolderRepository

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/5a51bbt5goou8

### DIFF
--- a/lib/blocs/create_manifest/create_manifest_cubit.dart
+++ b/lib/blocs/create_manifest/create_manifest_cubit.dart
@@ -208,8 +208,9 @@ class CreateManifestCubit extends Cubit<CreateManifestState> {
           existingManifestFileId: existingManifestFileId,
         ),
       );
-    } catch (err) {
-      addError(err);
+    } catch (e) {
+      logger.e('Failed to prepare manifest file', e);
+      addError(e);
     }
   }
 
--- a/lib/blocs/create_manifest/create_manifest_cubit.dart
+++ b/lib/blocs/create_manifest/create_manifest_cubit.dart
@@ -220,29 +220,32 @@ class CreateManifestCubit extends Cubit<CreateManifestState> {
     }
 
     if (state is CreateManifestUploadReview) {
-      final createManifestUploadReview = state as CreateManifestUploadReview;
-      final uploadType =
-          createManifestUploadReview.uploadMethod == UploadMethod.ar
-              ? UploadType.d2n
-              : UploadType.turbo;
+      try {
+        final createManifestUploadReview = state as CreateManifestUploadReview;
+        final uploadType =
+            createManifestUploadReview.uploadMethod == UploadMethod.ar
+                ? UploadType.d2n
+                : UploadType.turbo;
 
-      emit(CreateManifestUploadInProgress());
+        emit(CreateManifestUploadInProgress());
 
-      await _manifestRepository.uploadManifest(
-        params: ManifestUploadParams(
-          manifestFile: createManifestUploadReview.manifestFile,
-          driveId: _drive.id,
-          parentFolderId: createManifestUploadReview.parentFolder.id,
-          existingManifestFileId:
-              createManifestUploadReview.existingManifestFileId,
-          uploadType: uploadType,
-          wallet: _auth.currentUser.wallet,
-        ),
-      );
+        await _manifestRepository.uploadManifest(
+          params: ManifestUploadParams(
+            manifestFile: createManifestUploadReview.manifestFile,
+            driveId: _drive.id,
+            parentFolderId: createManifestUploadReview.parentFolder.id,
+            existingManifestFileId:
+                createManifestUploadReview.existingManifestFileId,
+            uploadType: uploadType,
+            wallet: _auth.currentUser.wallet,
+          ),
+        );
 
-      emit(CreateManifestSuccess());
-
-      return;
+        emit(CreateManifestSuccess());
+      } catch (e) {
+        logger.e('An error occured uploading the manifest.', e);
+        addError(e);
+      }
     }
   }
 
--- a/lib/blocs/create_manifest/create_manifest_cubit.dart
+++ b/lib/blocs/create_manifest/create_manifest_cubit.dart
@@ -1,80 +1,80 @@
 import 'dart:async';
 
+import 'package:ardrive/authentication/ardrive_auth.dart';
 import 'package:ardrive/blocs/blocs.dart';
-import 'package:ardrive/core/arfs/repository/file_repository.dart';
+import 'package:ardrive/blocs/upload/models/payment_method_info.dart';
 import 'package:ardrive/core/arfs/repository/folder_repository.dart';
-import 'package:ardrive/entities/entities.dart';
-import 'package:ardrive/entities/manifest_data.dart';
+import 'package:ardrive/manifest/domain/manifest_repository.dart';
 import 'package:ardrive/models/models.dart';
-import 'package:ardrive/services/services.dart';
-import 'package:ardrive/turbo/services/upload_service.dart';
 import 'package:ardrive/utils/logger.dart';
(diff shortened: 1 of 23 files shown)


lib/components/progress_dialog.dart
```diff
@@ -26,49 +26,61 @@ class ProgressDialog extends StatelessWidget {
...
+    this.useNewArDriveUI = false,
   });
 
   final String title;
   final List<ModalAction> actions;
   final Widget? progressDescription;
   final Widget? progressBar;
   final Widget? percentageDetails;
-
+  final bool useNewArDriveUI;
```
thiagocarvalhodev, line 37: To make it easier while we migrate the new modals

lib/manifest/domain/manifest_repository.dart
```diff
@@ -0,0 +1,248 @@
...
+
+  ManifestRepositoryImpl(
+    this._driveDao,
+    this._uploader,
+    this._folderRepository,
+    this._builder,
+    this._versionRevisionStatusUtils,
+  );
+
+  String? _existingManifestFileId;
```
kunstmusik, line 60: Think it would be better to get rid of this property as it seems like it could invite bugs. Would rather see the return type of checkNameConflict be modified so that the caller can know the result there (conflict, existing, no conflict).

lib/components/progress_bar.dart
```diff
@@ -17,25 +17,25 @@ class _ProgressBarState extends State<ProgressBar> {
...
-        builder: (context, snapshot) {
-          if (snapshot.hasData) {
-            _percentage =
-                ((snapshot.data!.progress * 100)).roundToDouble() / 100;
-          } else {
-            _percentage = 0;
-          }
+      stream: widget.percentage,
+      builder: (context, snapshot) {
+        if (snapshot.hasData) {
```
kunstmusik, line 22: Not important, but prefer a ternary here vs. if/else

lib/core/arfs/exceptions.dart
```diff
@@ -0,0 +1,16 @@
+abstract class ARFSException implements Exception {
```
kunstmusik, line 1: Checking, does this need to be ignore for Sentry?

packages/ardrive_uploader/lib/src/upload_controller.dart
```diff
@@ -403,9 +403,7 @@ class _UploadController implements UploadController {
     print('Upload Error');
   };
 
-  void Function(UploadTask task) _onCompleteTask = (UploadTask tasks) {
-    print('Upload Canceled');
-  };
+  void Function(UploadTask task) _onCompleteTask = (UploadTask tasks) {};
```
kunstmusik, line 406: This function has an empty body now, but the ones above it still have prints. Should they all be empty body functions? Are these default functions in case someone doesn't set them?

lib/blocs/create_manifest/create_manifest_cubit.dart
```diff
@@ -351,41 +214,31 @@ class CreateManifestCubit extends Cubit<CreateManifestState> {
...
+          uploadType: uploadType,
+          wallet: _auth.currentUser.wallet,
+        ),
+      );
 
       emit(CreateManifestSuccess());
-    } catch (err) {
-      addError(err);
+
+      return;
```
kunstmusik, line 241: This code changed from try/catch with addError() to just throwing any exception from _manifestRepository.uploadManifest(). Is that intended? There's still a call to addError() in prepareManifestTx(). 

Also, this return doesn't seem necessary as it's at the end of the method.

lib/core/arfs/repository/file_repository.dart
```diff
@@ -32,4 +37,15 @@ class _FileRepository implements FileRepository {
     final filePath = '$folderPath/${file.name}';
     return filePath;
   }
+
+  // TODO: implement unit tests for this method
```
kunstmusik, line 41: TODO mentioned here to add unit test. Is the plan to come back to this in another task or was it missed work that should be done in this PR?

lib/core/arfs/repository/folder_repository.dart
```diff
@@ -54,4 +74,71 @@ class _FolderRepository implements FolderRepository {
     // This will correctly handle the scenario when pathComponents is empty
     return pathComponents.join('/');
   }
+
+  // TODO: implement unit tests for this method
```
kunstmusik, line 78: Again in this file, TODOs listed for adding unit tests. Should these be done now in this PR or in a separate task?

lib/components/create_manifest_form.dart
```diff
@@ -29,19 +43,40 @@ Future<void> promptToCreateManifest(
...
+          context.read<DriveDao>(),
+          ArDriveUploader(
+            turboUploadUri:
+                Uri.parse(configService.config.defaultTurboUploadUrl!),
+            metadataGenerator: ARFSUploadMetadataGenerator(
+              tagsGenerator: ARFSTagsGenetator(
+                appInfoServices: AppInfoServices(),
+              ),
+            ),
+            arweave: Arweave(
```
kunstmusik, line 67: Noticing we've got a few places now where Arweave instance is created. Seems like we should consolidate and/or use the ArweaveService. Not necessary for this PR but worth noting for tech debt.

lib/manifest/domain/manifest_repository.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1739/commits/8cfe038afa21cf7ea73b4b7e2423ad5dab45e8d2

lib/core/arfs/exceptions.dart
thiagocarvalhodev: I don't think so. This should be used/extended for ARFS exceptions, like moving, renaming, resolving paths, etc.

packages/ardrive_uploader/lib/src/upload_controller.dart
thiagocarvalhodev: Those are defaults values that we can override by calling the respective public method, like this one:

```dart
@override
  void onCompleteTask(Function(UploadTask tasks) callback) {
    _onCompleteTask = callback;
  }
```

lib/components/create_manifest_form.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1739/commits/39d7dc64544cae124b5f806b43b47793db20ea80

lib/blocs/create_manifest/create_manifest_cubit.dart
thiagocarvalhodev: Nice catch!

https://github.com/ardriveapp/ardrive-web/pull/1739/commits/ac28404f83c5c6f89a59006e30d04dbd18536155

lib/components/progress_bar.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1739/commits/c7a97c505ee25c0dc4ac3c40398d2ca2cc9e5363

lib/core/arfs/repository/folder_repository.dart
thiagocarvalhodev: It'll be implemented in a next task.

lib/core/arfs/repository/file_repository.dart
thiagocarvalhodev: It'll be implemented in a next task.

kunstmusik (approved):

PR #1749, opened 2024-05-17
PE-5911: feat(upload)

thiagocarvalhodev:
- Enhance error handling
- Fixes an issue with upload folders (more in the pr comments)
- Removed some logs

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/1iofrch7u28so

### DIFF
--- a/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
+++ b/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
@@ -300,7 +300,6 @@ class _ArDriveUploader implements ArDriveUploader {
             task = task.copyWith(status: UploadStatus.failed);
             uploadController.updateProgress(task: task);
           }
-          uploadController.cancel();
         }
       });
     } else {
--- a/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
+++ b/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
@@ -287,9 +287,15 @@ class _ArDriveUploader implements ArDriveUploader {
     }
 
     if (folderUploadTask != null) {
-      // first sends the upload task for the folder and then uploads the files
-      uploadController.sendTask(folderUploadTask, wallet, onTaskCompleted: () {
-        uploadController.sendTasks(wallet);
+      /// first sends the upload task for the folder and then uploads the files
+      uploadController.sendTask(folderUploadTask, wallet,
+          onTaskCompleted: (success) {
+        logger.i('Folder upload task completed with success: $success');
+        if (success) {
+          uploadController.sendTasks(wallet);
+        } else {
+          uploadController.cancel();
+        }
       });
     } else {
       uploadController.sendTasks(wallet);
--- a/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
+++ b/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
@@ -294,6 +294,11 @@ class _ArDriveUploader implements ArDriveUploader {
         if (success) {
           uploadController.sendTasks(wallet);
         } else {
+          /// if the folder upload task fails, then all the files are marked as failed
+          for (var task in uploadController.notCompletedTasks) {
+            task = task.copyWith(status: UploadStatus.failed);
+          }
+          uploadController.updateProgress();
           uploadController.cancel();
         }
       });
--- a/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
+++ b/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:ardrive_io/ardrive_io.dart';
 import 'package:ardrive_uploader/ardrive_uploader.dart';
+import 'package:ardrive_uploader/src/upload_dispatcher.dart';
 import 'package:ardrive_uploader/src/utils/logger.dart';
 import 'package:ardrive_utils/ardrive_utils.dart';
 import 'package:arweave/arweave.dart';
--- a/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
+++ b/packages/ardrive_uploader/lib/src/ardrive_uploader.dart
@@ -298,8 +298,8 @@ class _ArDriveUploader implements ArDriveUploader {
           /// if the folder upload task fails, then all the files are marked as failed
           for (var task in uploadController.notCompletedTasks) {
             task = task.copyWith(status: UploadStatus.failed);
+            uploadController.updateProgress(task: task);
           }
-          uploadController.updateProgress();
           uploadController.cancel();
         }
       });
--- a/packages/ardrive_uploader/lib/src/arfs_upload_metadata.dart
+++ b/packages/ardrive_uploader/lib/src/arfs_upload_metadata.dart
@@ -1,15 +1,13 @@
+import 'package:ardrive_utils/ardrive_utils.dart';
 import 'package:arweave/arweave.dart';
 
 abstract class UploadMetadata {}
 
 class ARFSDriveUploadMetadata extends ARFSUploadMetadata {
   ARFSDriveUploadMetadata({
-    required super.entityMetadataTags,
     required super.name,
     required super.id,
     required super.isPrivate,
-    required super.dataItemTags,
-    required super.bundleTags,
   });
 
   // TODO: implement toJson
@@ -26,12 +24,9 @@ class ARFSFolderUploadMetatadata extends ARFSUploadMetadata {
   ARFSFolderUploadMetatadata({
     required this.driveId,
     this.parentFolderId,
-    required super.entityMetadataTags,
     required super.name,
     required super.id,
     required super.isPrivate,
-    required super.dataItemTags,
-    required super.bundleTags,
   });
 
   @override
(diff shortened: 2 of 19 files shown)


packages/ardrive_uploader/lib/src/ardrive_uploader.dart
```diff
@@ -287,9 +288,20 @@ class _ArDriveUploader implements ArDriveUploader {
...
+        if (success) {
+          uploadController.sendTasks(wallet);
+        } else {
+          /// if the folder upload task fails, then all the files are marked as failed
+          for (var task in uploadController.notCompletedTasks) {
+            task = task.copyWith(status: UploadStatus.failed);
+          }
+          uploadController.updateProgress();
+          uploadController.cancel();
+        }
```
thiagocarvalhodev, line 304: The previous method was wrong. The `onTaskCompleted` didn't take into account if the folder(s) was uploaded with success. So, even if the upload folder fails, the files would be uploaded.

packages/ardrive_uploader/lib/src/data_bundler.dart
```diff
@@ -399,62 +356,12 @@ class BDIDataBundler implements DataBundler<DataItemResult> {
     });
   }
 
-  @override
```
thiagocarvalhodev, line 402: This method was not being used.

packages/ardrive_uploader/lib/src/data_bundler.dart
```diff
@@ -381,10 +333,15 @@ class BDIDataBundler implements DataBundler<DataItemResult> {
...
+    final bundleTags = getBundleTags(
+      AppInfoServices(),
+      customBundleTags,
+    );
+
     final createBundledDataItem = createBundledDataItemTaskEither(
       dataItemFiles: dataItemsFiles,
       wallet: wallet,
-      tags: metadata.bundleTags.map((e) => createTag(e.name, e.value)).toList(),
+      tags: bundleTags.map((e) => createTag(e.name, e.value)).toList(),
```
thiagocarvalhodev, line 344: The bundle tags are now decoupled from the metadata classes.

packages/ardrive_uploader/lib/src/metadata_generator.dart
```diff
@@ -92,25 +82,22 @@ class ARFSUploadMetadataGenerator
...
+          lastModifiedDate: file.lastModifiedDate,
+          dataContentType: file.contentType,
+          driveId: arguments.driveId!,
+          parentFolderId: arguments.parentFolderId!,
+          name: file.name,
+          id: id,
+          licenseDefinitionTxId: arguments.licenseDefinitionTxId,
+          licenseAdditionalTags: arguments.licenseAdditionalTags)
+        ..setDataTags(tags['data-item']!)
+        ..setEntityMetadataTags(tags['entity']!);
```
thiagocarvalhodev, line 100: Sets the tags from the tags generator.

packages/ardrive_uploader/lib/src/ardrive_uploader.dart
```diff
@@ -287,9 +288,20 @@ class _ArDriveUploader implements ArDriveUploader {
...
+      /// first sends the upload task for the folder and then uploads the files
+      uploadController.sendTask(folderUploadTask, wallet,
+          onTaskCompleted: (success) {
+        logger.i('Folder upload task completed with success: $success');
+        if (success) {
+          uploadController.sendTasks(wallet);
+        } else {
+          /// if the folder upload task fails, then all the files are marked as failed
+          for (var task in uploadController.notCompletedTasks) {
+            task = task.copyWith(status: UploadStatus.failed);
```
kunstmusik, line 300: If I understand this correctly, this code seems strange. It looks like it's making a copy of the task but storing it in a local variable, then doing nothing further with that. Is this meant to update the notCompletedTasks?

packages/ardrive_uploader/lib/src/arfs_upload_metadata.dart
```diff
@@ -42,15 +44,61 @@ class ARFSFolderUploadMetatadata extends ARFSUploadMetadata {
...
+  /// The data tags for the file. These are the tags that will be added to the data transaction.
+  List<Tag> _dataTags = [];
+
+  /// The transaction ID for the data.
+  String? _dataTxId;
+
+  /// Gets the data transaction ID.
+  String? get dataTxId => _dataTxId;
+
+  /// Updates the data transaction ID.
```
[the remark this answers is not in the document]
thiagocarvalhodev: I thought having a named method would make more explicit the action of updating the tx id, since it is a very important property of that class

packages/ardrive_uploader/lib/src/ardrive_uploader.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1749/commits/833b6404a764ebace499f3d049d6331c51d52a12

thiagocarvalhodev: We must call the updateProgress passing the task with the new status

packages/ardrive_uploader/lib/src/upload_controller.dart
```diff
@@ -73,6 +80,10 @@ class _UploadController implements UploadController {
   final Map<String, UploadTask> _canceledTasks = {};
   final Map<String, UploadTask> _inProgressTasks = {};
 
+  @override
+  List<UploadTask> get notCompletedTasks =>
+      tasks.values.where((e) => e.status == UploadStatus.complete).toList();
```
kunstmusik, line 85: Doesn't this return those that *are* completed? Assuming this should be != ?

packages/ardrive_uploader/lib/src/upload_controller.dart
```diff
@@ -227,16 +241,22 @@ class _UploadController implements UploadController {
         );
 
         if (!uploadResult.success) {
-          final updatedTask = tasks[task.id]!;
+          final updatedTask =
```
kunstmusik, line 244: Should this not update with both error and status? Seems like it should be marked as failed before sending to _onFailedTask. Could set both fields with first copyWIth, then pass to both updateProgress and _onFailedTask.

packages/ardrive_uploader/lib/src/upload_controller.dart
```diff
@@ -317,7 +349,11 @@ class _UploadController implements UploadController {
...
```
kunstmusik, line 352: Same here about setting error and status before sending to and updatedProgress and onFailedTask. Also, order of calls here is different than earlier.

packages/ardrive_uploader/lib/src/upload_task.dart
```diff
@@ -0,0 +1,205 @@
...
+    List<ARFSUploadMetadata>? content,
+    SecretKey? encryptionKey,
+    List<(ARFSFolderUploadMetatadata, IOEntity)>? folders,
+    UploadTaskCancelToken? cancelToken,
+    UploadType? type,
+    Object? error,
+  }) {
+    return FolderUploadTask(
+      cancelToken: cancelToken ?? this.cancelToken,
+      folders: folders ?? this.folders,
```
kunstmusik, line 193: Interesting, this allows setting folders, but FileUploadTask does not allow setting file. Seems like it should be one way or the other.

packages/ardrive_uploader/lib/src/upload_controller.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1749/commits/0eeec628c9a55880f54eef6c39c4be157964522c

packages/ardrive_uploader/lib/src/upload_controller.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1749/commits/0eeec628c9a55880f54eef6c39c4be157964522c

packages/ardrive_uploader/lib/src/upload_controller.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1749/commits/aedc14ab32ba2102df1748c5d685965609d264c3

packages/ardrive_uploader/lib/src/upload_task.dart
thiagocarvalhodev: As we're using the copy with pattern, we must ensure we're copying all the properties of the object - at least the public ones. I'll add the file on the file class. Good catch!

kunstmusik (approved):

packages/ardrive_uploader/lib/src/upload_controller.dart
```diff
@@ -318,12 +318,12 @@ class _UploadController implements UploadController {
 
     init();
 
-    /// All folders goes in a single bundle. We are safe to send all the folders at once.
-    final folderTasks =
-        tasks.values.whereType<FolderUploadTask>().toList().first;
-
     /// If the tasks contains a folder, we must send the folder first
     if (containtsFolder) {
```
kunstmusik, line 322: Can you rename to "containsFolder"?

kunstmusik (approved):

PR #1751, opened 2024-05-24
PE-6147: feat(empty state): implement empty state

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/6vbacts030lto

### DIFF
--- a/lib/pages/drive_detail/components/drive_detail_folder_empty_card.dart
+++ b/lib/pages/drive_detail/components/drive_detail_folder_empty_card.dart
@@ -4,83 +4,454 @@ class DriveDetailFolderEmptyCard extends StatelessWidget {
   final bool promptToAddFiles;
   final String driveId;
   final String parentFolderId;
+  final bool isRootFolder;
 
   const DriveDetailFolderEmptyCard({
     super.key,
     this.promptToAddFiles = false,
     required this.driveId,
     required this.parentFolderId,
+    required this.isRootFolder,
   });
 
   @override
   Widget build(BuildContext context) => buildArDriveCard(context);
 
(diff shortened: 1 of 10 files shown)


lib/pages/drive_detail/components/drive_detail_folder_empty_card.dart
```diff
@@ -1,87 +1,704 @@
 part of '../drive_detail_page.dart';
 
-class DriveDetailFolderEmptyCard extends StatelessWidget {
+class DriveDetailFolderEmptyCard extends StatefulWidget {
```
kunstmusik, line 3: Noticing a lot of StatefulWidget in this file for Widgets that don't really have much state. Wonder if we should add a PlausibleTrackingWidget wrapper that is stateful that we can use to wrap StatelessWidgets. Might clean up the codebase a little bit.

thiagocarvalhodev: I liked that suggestion

kunstmusik:
General note: tested on Firebase URL and found on mobile I could scroll to see empty state cards, but not on other breakpoints.

lib/pages/drive_detail/components/drive_detail_folder_empty_card.dart
thiagocarvalhodev: https://github.com/ardriveapp/ardrive-web/pull/1751/commits/23a2ed7a3ba9b9d79ccb660c02529418707bc277

kunstmusik (approved):

PR #1755, opened 2024-05-29
PE-6222: add health check option on dev tools panel

thiagocarvalhodev:
We are implementing a drive health check tool for now only available on the dev tools. This tool will allow users to verify the health of a drive by checking the existence of files within that drive.

## How to use
After fully syncing your drive, type: `SHIFT + H` and it will open the dev tools panel. Inside the dev tools, click on `Run Health Check`. It will display the modal showing all the drives and the health of each one.

The `SHIFT + H` is required since we need the right `BuildContext` to display the modal properly.

This feature is still a work in progress.

Example:
image

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/4dcih33nq72k0


lib/pages/drive_detail/drive_detail_page.dart
```diff
@@ -146,30 +149,25 @@ class _DriveDetailPageState extends State<DriveDetailPage> {
...
+                  customShortcuts: [
+                    Shortcut(
+                      modifier: LogicalKeyboardKey.shiftLeft,
+                      key: LogicalKeyboardKey.keyH,
+                      action: () {
+                        ArDriveDevTools.instance
+                            .showDevTools(optionalContext: context);
+                      },
+                    ),
+                  ],
```
thiagocarvalhodev, line 162: Adds the shortcut for opening the dev tools with the right context.

lib/dev_tools/drives_health_check.dart
```diff
@@ -0,0 +1,527 @@
...
+    driveDao.select(driveDao.drives).get().then((drives) {
+      setState(() {
+        this.drives = drives;
+      });
+
+      processDrivesInSequency();
+    });
+  }
+
+  Future<void> processDrivesInSequency() async {
```
kunstmusik, line 40: Awkward name, suggest "processDrivesSequentially"

lib/dev_tools/drives_health_check.dart
```diff
@@ -0,0 +1,527 @@
...
+    for (final drive in drives) {
+      final status = DriveHealthCheckStatus(
+        drive: drive,
+        files: [],
+        totalFiles: 0,
+      );
+
+      driveStatuses.add(status);
+
+      setState(() {});
```
kunstmusik, line 52: could move this to outside of the loop, seems like it would flood updates if the user has a lot of drives

lib/dev_tools/drives_health_check.dart
```diff
@@ -0,0 +1,527 @@
...
+    }
+
+    for (final currentStatus in driveStatuses) {
+      final files = await (driveDao.select(driveDao.fileEntries)
+            ..where((tbl) => tbl.driveId.equals(currentStatus.drive.id)))
+          .get();
+      if (files.isEmpty) {
+        currentStatus.isLoading = false;
+
+        setState(() {});
```
kunstmusik, line 62: Looking at all of these setState calls makes me think this function might be better expressed as a Bloc with events

lib/dev_tools/drives_health_check.dart
```diff
@@ -0,0 +1,527 @@
...
+                mainAxisSize: MainAxisSize.min,
+                crossAxisAlignment: CrossAxisAlignment.start,
+                children: [
+                  Flexible(
+                    flex: 2,
+                    child: SingleChildScrollView(
+                      child: Column(
+                        mainAxisSize: MainAxisSize.min,
+                        children: [
+                          Text('Drives',
```
kunstmusik, line 106: Not necessary for this PR, but this is reminding me we might want to audit the codebase for localization of text strings.

lib/dev_tools/drives_health_check.dart
```diff
@@ -0,0 +1,527 @@
...
+    // Function to process files
+    void processNext() {
+      if (files.isNotEmpty) {
+        final file = files.removeAt(0);
+        checkHealth(file, driveStatus).then((_) {
+          controller.add(null);
+          setState(() {});
+        });
+      } else {
+        controller.close();
```
kunstmusik, line 285: Double-checking, it looks possible to me that a call to processNext could close the stream before a previous checkHealth completes. Can you verify?

lib/dev_tools/drives_health_check.dart
thiagocarvalhodev: That would be the ideal. As we're implementing this feature just as an internal tool and won't be available for the general users, I did the most simple implementation using the stateful widget. It's an experimental tool

lib/dev_tools/drives_health_check.dart
thiagocarvalhodev: Agreed

lib/dev_tools/drives_health_check.dart
thiagocarvalhodev: We're closing the stream when the files list is empty. So, there's no more items to be processed. 
The 
```dart
    await for (final _ in controller.stream) {
      processNext();
      setState(() {});
    }
```
will close the stream only when we processed all the items

kunstmusik: Leaving a note: looks like the latest changes still could have an issue awaiting for all to complete, but in real usage they'll all resolve and show up in the UI.

kunstmusik (dismissed):

kunstmusik (approved):

karlprieb (approved):

PR #1810, opened 2024-07-30
PE-6517: feat(snapshots)

thiagocarvalhodev:
- validate snapshot existence before syncing from snapshots

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/11ohs6433ophg

### DIFF
--- a/lib/sync/data/snapshot_validation_service.dart
+++ b/lib/sync/data/snapshot_validation_service.dart
@@ -19,7 +19,7 @@ class SnapshotValidationService {
       final appConfig = _configService.config;
 
       try {
-        final headers = {'Accept': '*/*'};
+        final headers = {'Accept': 'content-length'};
 
         final snapshotValidation = await http.head(
           Uri.parse(
--- a/lib/sync/data/snapshot_validation_service.dart
+++ b/lib/sync/data/snapshot_validation_service.dart
@@ -1,0 +1,66 @@
+import 'package:ardrive/services/config/config_service.dart';
+import 'package:ardrive/utils/logger.dart';
+import 'package:ardrive/utils/snapshots/snapshot_item.dart';
+import 'package:http/http.dart' as http;
+
+class SnapshotValidationService {
+  final ConfigService _configService;
+
+  SnapshotValidationService({
+    required ConfigService configService,
+  }) : _configService = configService;
+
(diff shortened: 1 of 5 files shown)


lib/sync/data/snapshot_validation_service.dart
```diff
@@ -0,0 +1,66 @@
...
+  Future<List<SnapshotItem>> validateSnapshotItems(
+    List<SnapshotItem> snapshotItems,
+  ) async {
+    List<SnapshotItem> snapshotsVerified = [];
+
+    final futures = snapshotItems.map((snapshotItem) async {
+      final appConfig = _configService.config;
+
+      try {
+        final headers = {'Accept': '*/*'};
```
karlprieb, line 22: Why don't you use the correct content type here?

karlprieb (commented):
What will be done with the snapshots that failed?
For what I can see we just ignore them, right?

thiagocarvalhodev:
> What will be done with the snapshots that failed?
For what I can see we just ignore them, right?

Exactly.

lib/sync/data/snapshot_validation_service.dart
```diff
@@ -0,0 +1,66 @@
...
+  Future<List<SnapshotItem>> validateSnapshotItems(
+    List<SnapshotItem> snapshotItems,
+  ) async {
+    List<SnapshotItem> snapshotsVerified = [];
+
+    final futures = snapshotItems.map((snapshotItem) async {
+      final appConfig = _configService.config;
+
+      try {
+        final headers = {'Accept': 'content-length'};
```
karlprieb, line 22: `Accept` headers are used to tell the server which content type you're willing to accept, not to signalize the headers you want.

lib/sync/data/snapshot_validation_service.dart
```diff
@@ -0,0 +1,63 @@
...
+
+            final headers = {
+              'Range': 'bytes=${lenght - 8}-$lenght',
+            };
+
+            final validationRequest = await http.get(
+              Uri.parse(
+                  '${appConfig.defaultArweaveGatewayUrl}${snapshotItem.txId}'),
+              headers: headers,
+            );
```
karlprieb, line 42: Why this get call is needed? Isn't the head call enough?

PR #1848, opened 2024-09-12
PE-6730: Decouple arns loading from the main upload flow

thiagocarvalhodev:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/4l9pl6jssh62g

### DIFF
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -105,24 +105,60 @@ class UploadCubit extends Cubit<UploadState> {
         isNextButtonEnabled: canUpload,
       ));
     } else if (state is UploadReadyToPrepare) {
-      final hasUndernames = (await _arnsRepository
-              .getAntRecordsForWallet(_auth.currentUser.walletAddress))
-          .isNotEmpty;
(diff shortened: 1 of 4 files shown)


lib/blocs/upload/upload_cubit.dart
```diff
@@ -105,24 +105,60 @@ class UploadCubit extends Cubit<UploadState> {
...
+
+          if (state is! UploadReady) {
+            logger.d('State is not UploadReady');
+            return;
+          }
+
+          final readyState = state as UploadReady;
+
+          emit(readyState.copyWith(
+              loadingArNSNames: false, showArnsCheckbox: showArnsCheckbox));
```
kunstmusik, line 141: Can't test this due to AO timeout issue, but should this also set showArnsSelection to true?

thiagocarvalhodev: Not yet. We first show the checkbox, then if the user checks it in the next modal we load the names

kunstmusik (approved):

PR #1872, opened 2024-09-25
PE-6802: feat(timestamps)

thiagocarvalhodev:
- use ISO and UTC formats to display dates in the app

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/1bedgacnqm31g


lib/pages/drive_detail/components/drive_detail_data_list.dart
```diff
@@ -200,8 +200,8 @@ Widget _buildDataListContent(
           name: row.name,
           typography: typography,
           size: row.size == null ? '-' : filesize(row.size),
-          lastUpdated: yMMdDateFormatter.format(row.lastUpdated),
-          dateCreated: yMMdDateFormatter.format(row.dateCreated),
+          lastUpdated: row.lastUpdated,
```
karlprieb, line 203: Is this intentional or you just forgot to use `formatDateToUtcString`?

kunstmusik:
Did a search of the branch code and found yMMDateFormatter is still used in a few places and wonder if they need to be updated or if it is fine to leave as-is. If you could confirm @thiagocarvalhodev that'd be great.

image

lib/pages/drive_detail/components/drive_detail_data_list.dart
thiagocarvalhodev: Before we were passing a String, now we're passing the DateTime, and inside the component we convert them

thiagocarvalhodev:
> Did a search of the branch code and found yMMDateFormatter is still used in a few places and wonder if they need to be updated or if it is fine to leave as-is. If you could confirm @thiagocarvalhodev that'd be great.
> 
> image

Good catch, @kunstmusik!! 

One of the files was not being used anywhere. I removed. There are just two places now on the item tile file. We will show on the drive explorer just the date instead of the timestamp, if the user hovers above the date we show the new formatted date

karlprieb (approved):

karlprieb (approved):

kunstmusik (approved):

PR #1873, opened 2024-09-26
PE-6786: change payment method selector

thiagocarvalhodev:
- use new icons on checkbox
- use new icons on license and change its style, also adds a question icon with a tooltip instead of a text link
- changes the upload payment method selector (only applies to the upload flow)
- FIXES:
  - the modal actions now have the correct title when checking the arns checkbox
  - arns checkbox now maintain its state
  - PE-6619

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/0ekvqmcgebt6g


LICENSE
karlprieb, line 1: Are we changing our license?

packages/ardrive_ui/lib/src/styles/icons/ar_drive_icons_icons.dart
```diff
@@ -1,4 +1,4 @@
-/// Flutter icons ArDriveIcons
+/// Flutter icons ArDriveIconsf
```
karlprieb, line 1: ArDriveIconsf?

thiagocarvalhodev: good catch! [95ee158](https://github.com/ardriveapp/ardrive-web/pull/1873/commits/95ee158abf1ad6e018f804780bd0637f1b39b3be)

karlprieb (approved):

PR #1890, opened 2024-10-10
PE-6751: fixes autoupdate manifest payment selector

thiagocarvalhodev:
- automatically start the upload if all manifest are free to upload

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/090vj6gqd4gi0

### DIFF
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -1260,6 +1260,10 @@ class UploadCubit extends Cubit<UploadState> {
           await prepareManifestUpload();
         }
 
+        UploadComplete(
+          manifestFiles: _selectedManifestFiles,
+        );
+
         PlausibleEventTracker.trackUploadSuccess();
       },
     );
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -174,7 +174,7 @@ class UploadCubit extends Cubit<UploadState> {
       return;
     }
 
-    if (manifestModels.any((element) => element.freeThanksToTurbo)) {
+    if (manifestModels.any((element) => !element.freeThanksToTurbo)) {
       emit(UploadManifestSelectPaymentMethod(
         files: manifestModels
             .map((e) =>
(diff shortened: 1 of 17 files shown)


lib/blocs/upload/upload_cubit.dart
```diff
@@ -717,6 +718,9 @@ class UploadCubit extends Cubit<UploadState> {
         ),
       );
     } else {
+      if (!isTest) {
+        await Future.delayed(const Duration(milliseconds: 100));
+      }
```
atticusofsparta, line 723: whats up with this?

lib/blocs/upload/upload_cubit.dart
```diff
@@ -814,12 +813,14 @@ class UploadCubit extends Cubit<UploadState> {
...
 
-    startUploadPreparation();
+    emit(UploadLoadingFilesSuccess());
   }
 
   Future<void> startUploadPreparation({
     bool isRetryingToPayWithTurbo = false,
   }) async {
+    await Future.delayed(const Duration(milliseconds: 100));
+
```
atticusofsparta, line 823: a pattern i see, hmmmmm

lib/blocs/upload/upload_cubit.dart
```diff
@@ -1223,53 +1240,13 @@ class UploadCubit extends Cubit<UploadState> {
...
-                undername: newUndername,
-                driveId: _targetDrive.id,
-                fileId: metadata.id,
-                processId: _selectedAntRecord!.processId,
-                uploadNewRevision: false,
-              );
-            }
+          if (tasks.first is FileUploadTask) {
+            await _postUploadFile(
+              task: tasks.first as FileUploadTask,
```
atticusofsparta, line 1247: Casting doesnt seem necessary here

thiagocarvalhodev: It is.

Dart’s type promotion has limitations:

1. Local Variables Only: Type promotion works only with local variables, not with properties, getters, or index accesses. Since tasks.first is a property access (even if tasks is a local variable), type promotion doesn’t apply.
2. Immutable Variables: The variable must not be mutated within the scope where the promotion is expected.

atticusofsparta: Fair enough - maybe could create a local variable with a typeguard - consider having a util for this that returns "FileUploadTask". But not a blocker.

atticusofsparta (approved):

PR #1914, opened 2024-11-15
PE-7106: feat(custom manifest)

thiagocarvalhodev:
- implement method to verify if the file is a custom manifest
- add checkbox on upload form to user set the file as a manifest
- upload file with the manifest content type

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/02a5g9lrg67v8

### DIFF
--- a/lib/blocs/upload/upload_cubit.dart
+++ b/lib/blocs/upload/upload_cubit.dart
@@ -11,6 +11,7 @@ import 'package:ardrive/core/activity_tracker.dart';
 import 'package:ardrive/core/upload/domain/repository/upload_repository.dart';
 import 'package:ardrive/core/upload/uploader.dart';
 import 'package:ardrive/core/upload/view/blocs/upload_manifest_options_bloc.dart';
+import 'package:ardrive/entities/constants.dart';
 import 'package:ardrive/main.dart';
 import 'package:ardrive/manifest/domain/manifest_repository.dart';
 import 'package:ardrive/models/forms/cc.dart';
@@ -21,6 +22,7 @@ import 'package:ardrive/services/config/config_service.dart';
 import 'package:ardrive/services/license/license.dart';
 import 'package:ardrive/turbo/services/upload_service.dart';
 import 'package:ardrive/utils/constants.dart';
+import 'package:ardrive/utils/is_custom_manifest.dart';
 import 'package:ardrive/utils/logger.dart';
 import 'package:ardrive/utils/plausible_event_tracker/plausible_custom_event_properties.dart';
 import 'package:ardrive/utils/plausible_event_tracker/plausible_event_tracker.dart';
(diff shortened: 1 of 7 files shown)


lib/blocs/upload/upload_cubit.dart
```diff
@@ -461,6 +473,8 @@ class UploadCubit extends Cubit<UploadState> {
             arnsRecords: _ants,
             showReviewButtonText: false,
             selectedManifestSelections: _selectedManifestModels,
+            isCustomManifest: isACustomManifest,
+            isUploadingCustomManifest: false,
```
kunstmusik, line 477: Should this be using isACustomManifest rather than false? Not sure why there's two values that look like they're doing the same thing

lib/blocs/upload/upload_cubit.dart
```diff
@@ -505,6 +519,8 @@ class UploadCubit extends Cubit<UploadState> {
             canShowSettings: showSettings,
             showReviewButtonText: false,
             selectedManifestSelections: _selectedManifestModels,
+            isCustomManifest: false, // only applies for single file uploads
+            isUploadingCustomManifest: false,
```
kunstmusik, line 523: Again not clear what's the difference between isCustomManifest and isUploadingCustomManifest. If they're supposed to be different I think they need better naming.

lib/utils/is_custom_manifest.dart
```diff
@@ -0,0 +1,22 @@
...
+import 'package:ardrive_io/ardrive_io.dart';
+
+/// Checks if a file is an Arweave manifest file by examining its content type and contents.
+///
+/// Returns true if the file has JSON content type and contains the string "arweave/paths",
+/// which indicates it follows the Arweave path manifest specification.
+Future<bool> isCustomManifest(IOFile file) async {
+  if (file.contentType == 'application/json') {
+    /// Read the first 100 bytes of the file
+    final first100Bytes = file.openReadStream(0, 100);
```
kunstmusik, line 12: Checking, is it possible for a manifest to be valid  with length < 100 bytes? If so, what happens here?

lib/utils/is_custom_manifest.dart
```diff
@@ -0,0 +1,22 @@
...
+
+    await for (var bytes in first100Bytes) {
```
kunstmusik, line 14: Little confused here as it looks like it is reading one byte at a time. Seems like this should convert everything into a string then check if that contains arweave/paths?

lib/utils/is_custom_manifest.dart
thiagocarvalhodev: Good catch! I don't think it's possible, but it will throw an exception.
I'm going to add a validation for this and a try catch block

kunstmusik (approved):

PR #1918, opened 2024-11-18
PE-7124: save drivekey in memory and fix upload failure loading arns names

thiagocarvalhodev:
…rns names

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/0mmvsvtqau60g

### DIFF
--- a/lib/services/arweave/arweave_service.dart
+++ b/lib/services/arweave/arweave_service.dart
@@ -4,6 +4,7 @@ import 'dart:convert';
 
 import 'package:ardrive/core/crypto/crypto.dart';
 import 'package:ardrive/entities/entities.dart';
+import 'package:ardrive/models/daos/drive_dao/drive_dao.dart';
 import 'package:ardrive/services/arweave/arweave_service_exception.dart';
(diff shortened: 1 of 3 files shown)


lib/services/arweave/arweave_service.dart
```diff
@@ -588,15 +590,24 @@ class ArweaveService {
...
-                  )
-                : null;
+        SecretKey? driveKey;
+
+        if (driveTx.getTag(EntityTag.drivePrivacy) == DrivePrivacyTag.private) {
+          driveKey = await _driveDao.getDriveKeyFromMemory(
+            driveTx.getTag(EntityTag.driveId)!,
+          );
+
+          driveKey ??= await _crypto.deriveDriveKey(
```
kunstmusik, line 600: This code reads a little strangely since it's always putting the value back into memory. Suggest to have it just be a conditional check on driveKey and if empty to do the deriveDriveKey and putKeyInMemory so that put is done only once.

kunstmusik (approved):

PR #1927, opened 2024-12-02
PE-7214: feat(primary name)

thiagocarvalhodev:
- load primary name

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/609k0t2no41m8


lib/components/profile_card.dart
```diff
@@ -548,15 +549,36 @@ class _ProfileCardState extends State<ProfileCard> {
...
-      variant: ButtonVariant.outline,
-      maxWidth: 100,
-      onPressed: () {
-        setState(() {
-          _showProfileCard = !_showProfileCard;
-        });
+    return BlocBuilder<ProfileNameBloc, ProfileNameState>(
+      builder: (context, state) {
+        final primaryName = state is ProfileNameLoaded
+            ? state.primaryName
```
kunstmusik, line 555: I think this won't truncate primary names to max 20 and add ... ? Or is it handled elsewhere?

lib/user/name/presentation/bloc/profile_name_bloc.dart
```diff
@@ -0,0 +1,75 @@
...
+      emit(
+        ProfileNameLoadedWithWalletAddress(
+          walletAddress,
+        ),
+      );
+    }
+  }
+
+  /// Truncates the primary name to 20 characters
+  String _truncatePrimaryName(String primaryName) {
```
kunstmusik, line 68: I see the truncation here, but does it add "..." elsewhere?

thiagocarvalhodev: We added the `overflow: TextOverflow.ellipsis` in our button, so it will apply the "..." automatically

kunstmusik (dismissed):
Would be good if the max width scaled more with the space available, as well as having a way to see the full name (tooltip for example), but I think this could go out and those can be addressed later.

kunstmusik (approved):

PR #1962, opened 2025-01-27
Fix: EthArEum login

kunstmusik:
* handle situation where multiple txs are made by wallet and settled in same block as initial signature recording tx

Verification system assumes first tx by wallet made is recording of verification signature. In some cases when a user made further tx's (e.g., added a folder, uploaded data), and those tx's were settled in the same block as the first tx, the GQL query would return a different tx than expected for verification. 

The change here does the following:

1. Get the blockheight of the first tx reported for the wallet
2. Query for all txs at that blockheight for the wallet
3. Filter txs by size == verification signature size
4. Download data for those txs and check against verification signature to check for a match


thiagocarvalhodev (approved):
LGTM

PR #2007, opened 2025-03-21
PE-7850: feat(manifest select the folder in view as target folder when creting a manifest

thiagocarvalhodev:


karlprieb (approved):

arielmelendez (changes requested):
This was broken when the changes for PE-7851 went in. Some rework needed.

PR #2017, opened 2025-04-24
PE-8014: don't report drive creation as successful in failure scenarios

arielmelendez:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/035d166ef9u6o


lib/blocs/drive_create/drive_create_cubit.dart
```diff
@@ -155,6 +155,7 @@ class DriveCreateCubit extends Cubit<DriveCreateState> {
       _drivesCubit.selectDrive(drive.id!);
     } catch (err) {
       addError(err);
+      return;
     }
 
     emit(DriveCreateSuccess(privacy: state.privacy));
```
kunstmusik, line 161: we can also move this emit to the end of the try block (after line 155)

PR #2061, opened 2025-08-20
PE-8473: fix: sync process improvements

vilenarios:
fix: improve sync cancellation and UX during post-sync operations
  - Add status messages for post-sync operations (ghost folders, licenses, transaction updates)
  - Display descriptive messages instead of just percentages during final sync stages
  - Reduce transaction status update timeout from 30s to 10s for better responsiveness
  - Add 5s timeout for individual GraphQL calls to prevent blocking
  - Fix cancellation exception propagation to ensure cancel modal always appears
  - Add more frequent cancellation checks throughout post-sync operations
  - Handle timeouts gracefully without showing error modals
  - Ensure cancellation exceptions are re-thrown through try-catch blocks

  The sync process now provides clearer feedback about what's happening during the
  final stages (92-100%) and allows cancellation to work properly even during
  long-running transaction status updates. Users will see messages like 'Creating
  ghost folders...' and 'Updating transaction statuses...' instead of just percentages.

  Resolves the '100% sync stuck' issue where cancellation wasn't working during
  post-sync operations.

--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/742cbb95pt0j0

### DIFF
--- a/lib/app_shell.dart
+++ b/lib/app_shell.dart
@@ -4,6 +4,7 @@ import 'package:ardrive/blocs/prompt_to_snapshot/prompt_to_snapshot_event.dart';
 import 'package:ardrive/components/migrate_private_drives_modal.dart';
 import 'package:ardrive/components/profile_card.dart';
 import 'package:ardrive/components/side_bar.dart';
+import 'package:ardrive/components/sync_failure_test_panel.dart';
 import 'package:ardrive/components/topbar/help_button.dart';
 import 'package:ardrive/misc/misc.dart';
 import 'package:ardrive/pages/drive_detail/components/hover_widget.dart';
@@ -16,6 +17,7 @@ import 'package:ardrive/utils/show_general_dialog.dart';
 import 'package:ardrive/utils/size_constants.dart';
 import 'package:ardrive_ui/ardrive_ui.dart';
 import 'package:ardrive_utils/ardrive_utils.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
@@ -97,7 +99,9 @@ class AppShellState extends State<AppShell> {
                   builder: (context, syncState) {
                     return Stack(children: [
                       scaffold,
-                      if (syncState is SyncInProgress)
+                      if (syncState is SyncInProgress || 
+                          syncState is SyncCancelled || 
+                          syncState is SyncCompleteWithErrors)
                         Stack(
                           children: [
                             SizedBox.expand(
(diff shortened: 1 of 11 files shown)


lib/app_shell.dart
```diff
@@ -192,6 +421,8 @@ class AppShellState extends State<AppShell> {
                               ),
                           ],
                         ),
+                      // Add the sync failure test panel (only visible in debug mode)
+                      if (kDebugMode) const SyncFailureTestPanel(),
```
thiagocarvalhodev, line 425: We could use the env variables and the dev tools panel to turn on/off this new feature. This way we can test in staging for example

lib/sync/domain/cubit/sync_cubit.dart
```diff
@@ -252,8 +259,28 @@ class SyncCubit extends Cubit<SyncState> {
...
+        // Clean up the cancellation token
+        _currentSyncToken?.dispose();
+        _currentSyncToken = null;
+        
+        emit(SyncCancelled(
+          drivesCompleted: _syncProgress.drivesSynced,
+          totalDrives: _syncProgress.drivesCount,
+          cancelledAt: DateTime.now(),
+        ));
+        _promptToSnapshotBloc.add(const SyncRunning(isRunning: false));
```
arielmelendez, line 273: Down below we call `unawaited(_updateContext());` which updates some logger context with info about the data that managed to get synced. First off, I'm wary of the fact that it's done unawaited... that aside, should we be doing that here also (in the case that a cancelled sync may still have had an impact on the _syncRepository)?

lib/sync/domain/cubit/sync_cubit.dart
```diff
@@ -299,6 +337,31 @@ class SyncCubit extends Cubit<SyncState> {
     }
   }
 
+  /// Cancel the current sync operation
+  void cancelSync() {
+    if (state is SyncInProgress && _currentSyncToken != null) {
+      logger.i('Requesting sync cancellation');
+      _currentSyncToken!.cancel();
+    }
```
arielmelendez, line 345: Some debug logs for the else case could be helpful if things get weird. e.g. `Sync attempted while state is <X> and _currentSyncToken <exists or doesn't exist>`

arielmelendez: Same comment for the two functions below.

lib/sync/domain/cubit/sync_cubit.dart
```diff
@@ -299,6 +337,31 @@ class SyncCubit extends Cubit<SyncState> {
...
+  }
+  
+  /// Clear the cancelled state and return to idle
+  void clearCancelledState() {
```
arielmelendez, line 349: Do these two public functions really need to require external intervention to get the state of the cubit where we want it? Said differently, should some other code path just manage this "automatically"? The worry is just that it creates more to do for external clients - i.e. something they might forget to do and/or do at the wrong time.

lib/sync/domain/cubit/sync_state.dart
```diff
@@ -20,3 +20,35 @@ class SyncFailure extends SyncState {
 class SyncEmpty extends SyncState {}
 
 class SyncWalletMismatch extends SyncState {}
+
+class SyncCancelled extends SyncState {
+  final int drivesCompleted;
```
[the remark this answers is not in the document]
arielmelendez: Related naming nit: if we keep these as counts, add either `num` or `Count` to the name.

lib/user/repositories/user_repository.dart
```diff
@@ -130,12 +130,26 @@ class _UserRepository implements UserRepository {
...
+      ]).then((res) => res[0] - res[1]);
+
+      return walletBalance;
+    } catch (e) {
+      // If getPendingTxFees fails (e.g., 504 error), fall back to just wallet balance
+      logger.w('Failed to get pending tx fees, using wallet balance only: $e');
+      
+      try {
+        // Try to at least get the wallet balance without pending fees
+        final walletBalance = await _arweave.getWalletBalance(walletAddress);
```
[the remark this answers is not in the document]
arielmelendez: aeb413c26

arielmelendez (approved):
Left a few questions and comments and pushed a few minor changes but overall should be ok if you've tested it well!

PR #2073, opened 2025-09-17
PE-8517: explose setting of GQL endpoint URL to users from advanced settings

kunstmusik:
--- Releases ---
Android release: https://appdistribution.firebase.google.com/testerapps/1:305132849030:android:6cf0cd5ec064fad3ffce07/releases/3957gvgvqmtm8


lib/components/profile_card.dart
```diff
@@ -559,11 +801,33 @@ class _ProfileCardState extends State<ProfileCard> {
   }
 
   void _closeProfileCardMobile() {
-    if (AppPlatform.isMobile) {
```
arielmelendez, line 562: Make sure this works as intended on mobile.

lib/components/profile_card.dart
```diff
@@ -559,11 +801,33 @@ class _ProfileCardState extends State<ProfileCard> {
...
+        onSave: (newEndpoint) {
+          const graphqlSuffix = '/graphql';
+          final normalizedEndpoint = newEndpoint.endsWith(graphqlSuffix)
+              ? newEndpoint.substring(
+                  0, newEndpoint.length - graphqlSuffix.length)
+              : newEndpoint;
+          final configService = context.read<ConfigService>();
+          configService.updateAppConfig(
+            configService.config.copyWith(
+              defaultArweaveGatewayUrl: normalizedEndpoint,
```
arielmelendez, line 822: Double check that this URL is only used for graphql. e.g. that's not being used for anything like getting chain data (i.e. headers, mempool depth, current block height, etc.)

lib/components/profile_card.dart
kunstmusik: Restored previous code, will test the mobile build created on GH actions and report back behavior here.

PR #2147, opened 2026-06-26
PE-9103: Skip GraphQL query for file download when tags are in local DB

vilenarios:
## Summary

For logged-in users, downloading a file triggered an unnecessary GraphQL query to fetch transaction tags (cipher, cipherIV, appName). These tags are already in the local DB's `file_revisions.customGQLTags` field from sync.

**Changes:**
- `personal_file_download_cubit`: reads tags from local DB first, falls back to GraphQL only if not cached
- `ardrive_downloader`: interface changed from `TransactionCommonMixin dataTx` to `String dataTxId` + `String? appName` — callers no longer need the full GraphQL transaction object
- Updated all callers: shared download cubit, thumbnail repository
- Shared file download still uses GraphQL (anonymous viewing has no local DB)

## Impact
- Logged-in file download: 0 GraphQL queries instead of 1-2
- Shared file download: unchanged (still needs network)
- Thumbnail generation: still fetches from network (thumbnail tx is different from file data tx)

## Test plan
- [x] `flutter analyze` — no issues
- [ ] Download a public file while logged in — verify no GraphQL queries in console
- [ ] Download a private file while logged in — verify decryption works
- [ ] Download a shared file (anonymous) — verify still works via GraphQL
- [ ] Thumbnail generation — verify still works

### DIFF
--- a/lib/blocs/file_download/file_download_cubit.dart
+++ b/lib/blocs/file_download/file_download_cubit.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:convert';
 
 import 'package:ardrive/core/arfs/entities/arfs_entities.dart';
 import 'package:ardrive/core/arfs/repository/arfs_repository.dart';
--- a/lib/blocs/file_download/personal_file_download_cubit.dart
+++ b/lib/blocs/file_download/personal_file_download_cubit.dart
@@ -142,15 +142,45 @@ class ProfileFileDownloadCubit extends FileDownloadCubit {
     String? cipher;
     String? cipherIvTag;
     SecretKey? fileKey;
+    String? appName;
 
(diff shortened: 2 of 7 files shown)


vilenarios:
Closing: the cipher/cipherIV tags needed for private file decryption are on the **data transaction**, not stored in the local DB's customGQLTags (which stores metadata transaction tags). The GQL call in personal_file_download_cubit is necessary for private files and is already skipped for public files. No optimization possible here.

vilenarios:
Closing: reviewed the current code on dev — the getTransactionDetails GQL call is already gated by \ (line 157). Public file downloads already make zero GQL calls. Private files need the data transaction's cipher/cipherIV tags which aren't stored locally. No optimization available.

PR #2162, opened 2026-07-05
PE-9129: Sync and login resiliency quick wins

vilenarios:
## Summary

First pass of the sync/data-fetching remediation plan: the highest-impact, lowest-risk fixes identified in the end-to-end sync analysis. Each change is a separate commit with its own tests; none alters happy-path behavior.

### Bug fixes
- **Login no longer fails or hangs when the balance endpoint is down** — `getUser` awaited `getWalletBalance` with no retry/timeout, locking out users with valid passwords during gateway blips. Balance is now best-effort (5s timeout, falls back to 0); `ArDriveAuth._updateBalance` refreshes it asynchronously right after login as before.
- **Wallets with 200+ drives no longer crash every sync** — `transactionParseBatchSize` computed `200 ~/ drivesCount`, which reaches 0 at 201 drives and makes `BatchProcessor` throw. Extracted `calculateTransactionParseBatchSize` clamped to ≥ 1.
- **`BulkImportResult.importedFiles` was always empty** — WorkerPool discards the execute callback's return value; successful imports are now collected in the callback.

### Performance
- **Drive syncs bounded to 5 concurrent** (was: every drive at once) via a new `runBoundedWorkers` helper mirroring `Future.wait(eagerError: false)` semantics — per-drive error handling, cancellation, and progress reporting unchanged.
- **`MetadataCache` constructed once** instead of rebuilt from SharedPreferences on every parsed batch of every drive.
- **Pin transaction info fetched 100 ids per request instead of 5** — added `first: 100` to the `InfoOfTransactionsToBePinned` query (it previously relied on the gateway's default page size, which is why the batch had to stay tiny), matching the `TransactionStatuses`/`LicenseAssertions` pattern.

## Test plan
- 13 new unit tests (batch-size clamp edge cases, balance-fetch failure paths, worker-pool bounds/failure isolation)
- `flutter analyze`: clean
- Full main-app suite on Flutter 3.19.6: 709 passed / 3 skipped / 0 failed
- Generated artemis code is gitignored; CI regenerates via `scr setup`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a **Refresh list** action to reload the available gateways.
  * Introduced cached/persisted gateway handling for improved availability.
  * Added configurable drive sync concurrency and drive-history pagination tuning.
* **Bug Fixes**
  * Improved sync robustness with bounded concurrent processing and safer pagination/cursor handling.
[...]

PR #2164, opened 2026-07-09
PE-9130: Point GraphQL at turbo-gateway.com directly with Goldsky fallback

vilenarios:
## Summary

Cuts the ardrive.net proxy hop out of the GraphQL path and hardens the fallback:

- **Primary GQL endpoint: `https://turbo-gateway.com`** (was `https://ardrive.net`, which proxies to turbo-gateway anyway). During the 2026-07-08 incident the ardrive.net proxy pool returned 503s while turbo-gateway itself recovered independently — the proxy was a single point of failure adding no value on this path. Data requests already go to turbo-gateway directly.
- **`configVersion` 2 → 3** in all three flavors so existing users' locally-stored configs are replaced on next app load (stored configs only refresh on version bump).
- **Fallback endpoint: `https://arweave-search.goldsky.com/graphql`** (was `https://arweave.net/graphql`, which proxies to Goldsky but rate-limits aggressively — probes got 429'd within a few requests).
- Documents Goldsky's page-size hazard discovered during endpoint testing: `first > 100` is silently clamped to 100 **with `hasNextPage` falsely reporting `false`** — a paginating client would silently truncate results. Fallback queries must stay ≤ 100/page (all current queries do; this matters for the planned page-size increase).

## Verification
- turbo-gateway.com/graphql serves the real sync query shape (owner + Drive-Id + block range, HEIGHT_ASC): 100 edges, correct `hasNextPage`.
- arweave-search.goldsky.com/graphql: healthy, correct pagination at `first: 100`.
- No test pins either endpoint URL; config JSONs validated.

## Rollout note
Merging to dev → staging.ardrive.io picks this up via the staging config; production users get it at the next release. Users with stored configs migrate automatically via the version bump.

https://claude.ai/code/session_0172nfTRDj7wgnhs44Lg6mxC

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default gateway and fallback GraphQL endpoint used by the app, improving request handling when the primary service is unavailable.
  * Bumped configuration versions across environments to keep settings aligned.

* **Chores**
  * Tightened a UI package dependency version range to improve compatibility and reduce upgrade-related issues.

### DIFF
--- a/assets/config/dev.json
+++ b/assets/config/dev.json
@@ -1,6 +1,6 @@
 {
-  "configVersion": 2,
-  "defaultArweaveGatewayUrl": "https://ardrive.net",
+  "configVersion": 3,
+  "defaultArweaveGatewayUrl": "https://turbo-gateway.com",
   "defaultArweaveGatewayForDataRequest": {
     "label": "Turbo Gateway",
     "url": "https://turbo-gateway.com"
--- a/assets/config/prod.json
+++ b/assets/config/prod.json
@@ -1,6 +1,6 @@
 {
-  "configVersion": 2,
-  "defaultArweaveGatewayUrl": "https://ardrive.net",
+  "configVersion": 3,
+  "defaultArweaveGatewayUrl": "https://turbo-gateway.com",
   "defaultArweaveGatewayForDataRequest": {
     "label": "Turbo Gateway",
     "url": "https://turbo-gateway.com"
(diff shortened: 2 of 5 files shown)


vilenarios:
@coderabbitai review

vilenarios:
Consolidated into #2162 so all sync improvements can be tested on one preview build. Branch merged there commit-for-commit.

PR #2165, opened 2026-07-09
PE-9131: Minimize Solana RPC usage in sync and gateway registry

vilenarios:
## Summary

Investigation of heavy traffic to ArDrive's QuikNode Solana RPC found two structural sources; this PR eliminates both.

### 1. Sync no longer sweeps ArNS on every run
Every sync (auto-sync ~5 min, tab-focus restarts, manual) called `getAntRecordsForWallet(update: true)` — deliberately bypassing the repository's 15-minute cache — walking every owned ArNS name on Solana (several RPC calls per name), plus a post-sync `saveAllFilesWithAssignedNames` pass. ArDrive currently has **no ArNS integration**, so this was pure RPC cost with no user-facing effect. Both call sites and the `ARNSRepository` dependency are removed from `SyncRepository`. On-demand ArNS lookups elsewhere (upload flows, profile name) are untouched; sync-time integration can return later with the feature.

### 2. GAR gateway list: fetch at most once ever, refresh only on demand
The AR.IO gateway list was re-fetched from Solana RPC every app session (data-fetch fallback, snapshot validation) and on every open of the gateway settings modal. It's now persisted in local storage:
- `DataGatewayFallback`: memory → persisted cache → single SDK fetch (persisted on success). Across sessions the network is hit at most once, ever.
- Gateway settings serve the cached list; a new **"Refresh list"** button (new `RefreshGateways` bloc event) force-fetches and persists — the only path that re-hits the RPC.
- `SnapshotValidationService` reads through the same shared cache.
- Corrupt/missing persisted entries fall back to a normal fetch; fetch failures are cached in memory only (never persisted) so the next session retries.

## Expected impact
Steady-state Solana RPC volume drops to ~zero for a signed-in user who isn't using crypto top-ups or explicitly refreshing the gateway list.

## Tests
- New `DataGatewayFallback` persistence suite (store-hit/no-RPC, fetch-once-persist, cross-session reuse, failure non-persistence, corrupt-entry recovery, forced refresh)
- `gar_repository_test` rewired to the shared cache (asserts no SDK call on plain `getGateways`)
- Sync optimization tests updated for the removed dependency
- Includes the shared equatable CI pin (identical commit as PRs #2162/#2164; merges cleanly in any order)

Note: local test execution was unavailable for this branch (WSL→Windows interop broke mid-session); CI is the validation gate for this PR — please wait for green before review.

https://claude.ai/code/session_0172nfTRDj7wgnhs44Lg6mxC

### DIFF
--- a/lib/gar/domain/repositories/gar_repository.dart
+++ b/lib/gar/domain/repositories/gar_repository.dart
@@ -8,6 +8,7 @@ import 'package:collection/collection.dart';
 
 abstract class GarRepository {
   Future<List<Gateway>> getGateways();
+  Future<List<Gateway>> refreshGateways();
   List<Gateway> searchGateways(String query);
   Future<Gateway> getSelectedGateway();
   Future<void> updateGateway(Gateway gateway);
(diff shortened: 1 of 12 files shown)


vilenarios:
Consolidated into #2162 so all sync improvements can be tested on one preview build. Branch merged there commit-for-commit.

PR #2166, opened 2026-07-16
PE-9132: Turbo free-tier support — allowance-aware upload UX and typed payment failures

vilenarios:
## Summary

Client readiness for Turbo's free tier (10 MiB per-wallet pool, 105 KiB per-item cap; see `docs/implementation_plan_turbo_free_tier.md`). The app now makes an **honest** free-vs-paid promise on every upload surface, backed by the `GET /v1/account/free` endpoint, and never leaves a payment rejection hanging.

Two bodies of work:

### 1. Typed payment failures & honest failure UX

- **Typed errors**: HTTP `402` → `TurboPaymentRequiredException` (app) / `UnderFundException` (uploader package); `429` → typed rate-limit. Both are **excluded from the uploader's retry loops** — retrying a payment rejection just multiplies metered load.
- **No more hangs**: every upload-touching surface (9 metadata ops, file/folder upload, snapshot, manifest, standalone + post-upload ArNS, single/multi thumbnail, bulk import) dismisses its progress UI on failure and, on a 402, shows the shared **"Free allowance used up → Buy Credits"** dialog instead of hanging or showing a generic error.
- **Move fixed**: gains a real failure state (it previously emitted *Success* after errors) and is reordered **post-then-commit** so a rejected move can't leave the local DB claiming a move the chain never saw.
- Surfaced payment errors that the shared `WorkerPool` was silently swallowing (bulk import, multi-thumbnail).

### 2. Allowance-aware free-vs-paid (this endpoint's payoff)

Previously "free" was decided from **item size alone**, so a user whose pool was exhausted was told an upload was free, had the payment selector hidden, and then hit a 402. Now free requires **both** size-eligibility **and** that the wallet's remaining allowance covers the upload.

| Situation | What the user sees |
|---|---|
| Covered by allowance | "…free thanks to Turbo" |
| Size-eligible, upload **exceeds** a non-zero allowance | payment selector + "This upload exceeds your free allowance and will need Credits or AR." |
| Allowance **used up** (zero) | payment selector + "Free allowance used up. This upload requires Credits or AR." |
| Item too large for the free tier | payment selector, no free message |

Wired across the file/folder, snapshot, and manifest paths (including manifest re-upload, which previously skipped payment selection entirely when it wrongly judged everything free). The four states are modeled as one `FreeUploadStatus` value rendered by one widget, so the surfaces can't drift apart.

### Design principles

- **Advisory, never a gate.** `bytesRemaining` is a point-in-time, wallet-level snapshot. It only decides what we *promise*; Turbo's `402` on upload remains the sole authority on what's actually charged.
- **Fails open.** An unreachable endpoint → `unknown` → the previous size-only behavior. A failed check never tells a user with allowance left that they must pay.
- **Honest, not precise.** The "exceeds" copy states the fact (upload > remaining) and the outcome (needs payment) without predicting *how much* ends up free — the client can't know that (Turbo bills server-side, granularity unexposed, plus a second per-IP pool `/v1/account/free` doesn't report). True whether Turbo frees part of the upload or none of it.

Endpoint host confirmed against prod: `GET payment.ardrive.io/v1/account/free` → `200 {"bytesRemaining":10485760}`.

## Tests

733 passing. New coverage for the status-code → exception mapping, `/v1/account/free` parsing (including `null` = unlimited, `0` = off, unparseable = unknown/fail-open), the `freeUploadStatusFor` derivation (boundaries + fail-open), and a widget test asserting each status renders the right message. Core logic and widget both mutation-verified.

## Notes for review

- The three new free-tier l10n keys currently carry English text in all six locales (branch convention); Localizely handles translation as a separate pass.
- `/v1/account/free` is placed on `turboPaymentUri` (verified in prod). Confirming with the Turbo team that the endpoint is a committed, stable contract would let us later tighten "exceeds your allowance" into an exact count — nothing depends on it otherwise.

https://claude.ai/code/session_01PsV2WFHZMGD9DUfX65cxuW

## Summary by CodeRabbit

* **New Features**
  * Added server-driven Turbo free-upload eligibility (server max item size) plus a “free allowance used up” flow for Turbo 402/429.
  * Introduced Turbo payment-required dialogs with localized copy and a “Buy Credits” action.
  * Added “free tier status” messaging above upload method selection.
* **Bug Fixes**
  * Improved payment-related error classification across create/rename/move, licensing, pinning, hide/unhide, upload, thumbnails, bulk import, and ArNS assignment so the correct dialog is shown.
  * Prevented blind retries for Turbo payment/rate-limit conditions.
* **Documentation**
  * Added a Turbo free-tier implementation plan.
* **Tests**
  * Added coverage for Turbo HTTP status-code → exception mapping and free-allowance parsing/behavior.


vilenarios:
Addressed the review findings in f40205b8:

- **Equatable props** — `isPaymentError` now included in `props` on all failure states (the highest-impact one: without it a payment failure emitted after a generic failure compared equal and the dialog wouldn't re-trigger).
- **License** — original exception preserved via `logger.e`; the failure card's action becomes "Buy Credits" → shared payment dialog on payment errors instead of retrying the rejected op.
- **Localization** — hardcoded generic error strings replaced with a shared `actionFailedTryAgain` key across all locales.

**Deferred (tracked, not in this PR):**
- **Multi-item move partial-failure semantics** (the Major): if one item in a multi-item move is rejected mid-batch, earlier items are already on-chain. This PR's post-then-commit reorder already removes the *local-state corruption* risk (nothing commits locally unless the network accepted it). True transactional/resumable multi-item moves are part of the burst-preflight follow-up noted in `docs/implementation_plan_turbo_free_tier.md`, not this PR's scope.
- **Per-locale translations**: new keys ship as English interim and are translated via the Localizely pipeline (`localizely.yml`), per repo convention.

PR #2172, opened 2026-07-29
PE-9132: Release v2.86.0

vilenarios:
# Release v2.86.0

Verified against the actual `master...dev` content diff (103 files) — this
release is the Turbo free-tier work plus a mobile fix and cleanup. The sync /
snapshot / tx-confirmation performance work shipped in **2.85.0** and is NOT
part of this release.

### 💳 Turbo free-tier support
The app now makes an **honest** free-vs-paid promise on every upload:
- Small files covered by your free allowance still upload free.
- When an upload is larger than your remaining allowance — or the allowance is
  used up — you're told clearly ("exceeds your free allowance" / "free
  allowance used up") and shown how to pay, instead of a surprise failure.
- Payment rejections are now handled cleanly across **every** upload surface
  (files/folders, metadata ops, snapshots, manifests, thumbnails, bulk import,
  ArNS name assignment): failures no longer hang — you get a clear dialog with
  a **Buy Credits** action.
- Move is fixed to fail honestly (it previously reported success after errors),
  and rename no longer hangs on a payment rejection.
- AR balance refreshes after a Turbo top-up from the profile dropdown.

### 📱 Mobile fix
The Arweave network-congestion popup no longer covers its own buttons on small
screens (the modal is now height-bounded and scrollable, and dismissible).

### 🧹 Cleanup
- Removed the feedback survey; added a link to the new help center.
- Small performance tweak: the Turbo free allowance is fetched concurrently
  with the balance during upload preparation, so the upload modal opens a touch
  faster.

---
_Full changelog: `v2.85.0...v2.86.0`_


vilenarios:
Superseded by a release/v2.86.0 branch → master, matching the v2.85.0 release process (keeps dev clean; resolves the squash-divergence conflict on the release branch).

PR #2161, opened 2026-07-04
PE-9103: Parallelize upload preparation pipeline

vilenarios:
## Summary

Optimizes the upload preparation phase by parallelizing sequential operations that were previously blocking on each other unnecessarily.

### Changes:

- **Parallel conflict detection**: All file conflict DB queries run concurrently via `Future.wait` instead of one sequential query per file. For 100 files: 100 sequential queries → 1 parallel batch.
- **Pre-computed file lengths**: All file lengths read in parallel at the start of `startUploadPreparation()` and cached in `_fileLengthCache`. Eliminates 2-3x redundant sequential reads per file across warning check, plan creation, and total size calculation.
- **Parallel cost estimation**: Turbo balance + AR bundle/file sizes fetched concurrently. AR and Turbo cost calculations run in parallel after sizes are computed.
- **Parallel upload plan creation**: AR and Turbo upload plans created simultaneously via `Future.wait` instead of sequentially.
- **Remove redundant operations**: Duplicate wallet mismatch check removed, ANT records not re-fetched if already populated, 100ms artificial UI delay reduced to `Duration.zero`.

### Impact

For multi-file uploads (50+ files), preparation time reduced from sequential O(n) DB queries to parallel O(1). Network-bound cost estimation parallelized (saves ~1-2s).

## Test plan

- [ ] Upload single file — verify preparation completes normally
- [ ] Upload 10+ files — verify faster preparation
- [ ] Upload folder with nested files — verify conflict detection works
- [ ] Upload with file name conflicts — verify conflict modal appears correctly
- [ ] Upload to private drive — verify encryption/key derivation works
- [ ] Turbo vs AR cost display — verify both costs shown correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved upload preparation by running planning, size, conflict checks, and cost estimation concurrently.
  * Added caching for file sizes so later steps reuse them without extra I/O.
  * Conflict detection now completes faster by processing per-file checks in parallel.
  * Upload payment evaluation now calculates AR and Turbo info in parallel and avoids extra network work when unnecessary.
  * Replaces a fixed delay with a UI-yield to keep the upload flow more responsive.

* **New Features**
  * Free-upload eligibility is now reported with clearer free-status details and eligibility checks.

PR #2173, opened 2026-07-30
PE-9211: Release v2.86.0 — file sharing rebuilt, downloads unblocked, sync resilience

vilenarios:
# Release v2.86.0

`dev` → `master`. Version `2.86.0` (`pubspec.yaml:6`). Deployed and verified on staging at `aca9707ac`.

A resilience release. Most of what is here exists because a gateway was slow, rate limiting, or answering with something other than what was asked for — and the app took that as a reason to fail rather than a reason to try elsewhere. Alongside it, file sharing is rebuilt around links that carry what the recipient needs.

## 🔒 Security

- **Private downloads on web were never authenticated, at any size.** Both ciphers went through a streaming decryptor whose AES-GCM branch trims the authentication tag and never checks it. GCM now buffers, verifies the MAC over every byte, and **writes nothing on failure**. This is the significant fix in the release and it is on.
- **The vendored pdf.js carried CVE-2024-4367** — a crafted PDF could run arbitrary JavaScript on the app origin, where a recipient's access key can sit in `sessionStorage`. Mitigated in the vendored build with a CI guard so it cannot silently revert. The real fix is pdf.js ≥ 4.2.67, which needs a newer Flutter than the pinned 3.19.6.
- **`plausible.js` sent `location.href`, fragment included, to plausible.io.** On a key-in-link share that string *is* the file key. The tag is removed.
- **The "Verified" badge was forgeable** by anyone able to post a transaction — the link's metadata was fetched with no owner constraint. Authorship is now checked, without adding a round trip before first paint.

### One thing that changed late, and honestly

An earlier draft of these notes promised chain-based integrity for AES-CTR and public downloads: the ANS-104 deep hash computed over the ciphertext as it streams, checked against the data item's signature. **That check shipped disabled**, and the reason matters:

- It reported healthy files as corrupt. Gateways return `anchor` and `recipient` as empty strings — verified live against goldsky, permagate.io, turbo-gateway.com and vilenarios.com — so any data item carrying either one hashes differently at verification than it did at signing. The verdict replaced the success dialog with "the file may be corrupted".
- Reaching it cost a GraphQL round trip the stream waited on before its first byte, and on a rate-limited connection it timed out every time.

The machinery remains behind a flag with nothing enabling it. **AES-GCM integrity is unaffected and still enforced** — it is computed locally over bytes in hand and needs no network. Restoring the signature check belongs with the transport work in the follow-ups.

## 🔗 File sharing

The access key no longer rides in the link by default. It is a separate artifact sent out of band, and the share dialog hands over both with copy that pushes them down different channels. Key-in-link stays as an explicit opt-in, and anything carrying key material is **masked until deliberately revealed** — handing a key over is the moment a screen is most likely to be shared, recorded or screenshotted.

A link now carries what the recipient's page needs, so a complete one paints with **zero blocking GraphQL**. Every field degrades independently, and v1 links resolve exactly as they always have.

**The recipient page** is a real state machine — locked, resolving, ready, not found, network trouble, damaged link. No eternal spinners, a wrong key says so rather than "file does not exist", and a recently-uploaded file retries instead of claiming it is missing. PDFs render inline, private video and audio preview, and recipients see real thumbnails.

Rebuilt around the file itself (#2186, #2189):

- **The file leads.** Identity and Download sit in a fixed header; details and version history are two tabs in a panel of fixed height, so nothing the reader touches resizes the card under the pointer on its way to Download.
- **Version history is a control, not a readout.** Any revision can be selected; the preview and the download target move together. A pinned link names its version pinned and still lets go. A renamed file no longer selects two rows at once.
- **Nothing blocks on history.** The version the link named renders from the first frame and the rest fills in around it. A failed lookup costs the list and nothing else.
- **Every state that cannot show the file** — still resolving, unsupported type, past the 100 MiB preview cap — answers with the file's own thumbnail over one sentence, instead of an empty pane.
- **Folder share links** are reachable from the file explorer, and a link naming a folder in a drive you have not attached yet lands on that folder once attached — not on another drive's folder.

## ⬇️ Downloads

- **No GraphQL before the bytes.** The arweave client's `download()` POSTs `{gateway}/graphql` before fetching anything — unconditionally, for a progress figure. That POST is CORS-sensitive, not every gateway allows it from an arbitrary origin, and it throws *before* the byte stream exists, so a gateway that could have served the file was recorded as failed and the next was tried until none were left. **Public files were failing to download outright.** Downloads now issue a direct `GET {gateway}/{txId}`.
- **The cipher travels in the link**, so a private download and a private preview no longer re-fetch tags the link already delivered.
- **Download dialogs say less and mean more.** A finished download says it finished. Two advisories are gone: "couldn't check this file against the original" appeared on every public download, and "Verified" on every successful private one — a label on 100% of outcomes carries no information.

## ⚡ Sync

Sync read every metadata item through a multi-gateway waterfall — configured gateway, two GAR gateways, then arweave.net, serially at 5s each — and building that list cost a Solana RPC. It now reads the configured gateway with two attempts, and the RPC is gone from the sync path.

The larger win was scheduling: the concurrency limiter was a barrier rather than a pool, so one slow item idled four workers. Over 300 items at a 10% failure rate, **321s → 83s**.

Then a round of failures that each cost a whole sync:

- **One drive's gateway hiccup no longer kills the whole sync** (#2179) — failures are isolated per drive.
- **Sync no longer goes to the gateway for what the snapshot already gave it** (#2184).
- **Read timeouts sized for metadata were being applied to 44 MiB snapshots** (#2183), so large snapshots timed out on connections that could have served them.
- **Login reads walk the gateway waterfall they were meant to skip** (#2181).
- **A drive whose root folder never synced no longer spins forever** (#2180).

**Known and documented:** sync still drops a file silently when its metadata cannot be read — empty bytes, a warning, and the watermark advances anyway. The block-height rewind is a ~2 hour recency window, not a retry. Skipped transaction ids now ride out on the sync result instead of vanishing, and `docs/SYNC_SKIPPED_ENTITY_PERSISTENCE.md` sets out the fix.

## 💳 Turbo free-tier support

The app now makes an **honest** free-vs-paid promise on every upload:

- Small files covered by your free allowance still upload free.
- When an upload is larger than your remaining allowance — or the allowance is used up — you're told clearly ("exceeds your free allowance" / "free allowance used up") and shown how to pay, instead of a surprise failure.
- Payment rejections are handled cleanly across **every** upload surface (files/folders, metadata ops, snapshots, manifests, thumbnails, bulk import, ArNS name assignment): failures no longer hang — you get a clear dialog with a **Buy Credits** action.
- Move is fixed to fail honestly (it previously reported success after errors), and rename no longer hangs on a payment rejection.
- AR balance refreshes after a Turbo top-up from the profile dropdown.
- The free-item cap is read from `/v1/info` rather than a stale config value.

## ⬆️ Uploads

- Upload preparation runs in parallel rather than serially (#2161).

## 📱 Mobile fix

The Arweave network-congestion popup no longer covers its own buttons on small screens (height-bounded, scrollable, dismissible).

## 🧹 Cleanup

- Removed the feedback survey; added a link to the new help center.
- The Turbo free allowance is fetched concurrently with the balance during upload preparation, so the upload modal opens a touch faster.
- `pica`, `ethers` and `pdf.js` are vendored — no runtime CDN dependency ships in a build that is permanent on Arweave.
- A manifest containing any non-Latin-1 character was unreadable, and now isn't.
- GraphQL retries drop from 8 attempts (~51s of backoff) to 5.
- The development flavor points at the `ar-io.dev` testnet.
- `CLAUDE.md` expanded with the repository facts that cost a reader time.

## Known follow-ups

Recorded rather than fixed here:

- **Syncs still crash for wallets with 200 or more drives.** `sync_repository.dart:502` sizes the transaction-parse batch as `200 ~/ (drivesCount - drivesSynced)`, which reaches `0` past 200 drives, and `BatchProcessor` throws on a zero batch size. The fix is a clamp; it sits in #2162 (PE-9129) with 28 other sync commits. That PR is **deferred out of this release** — it merges without textual conflict, but seven of its files were also rewritten by the sync work landing here, so it needs rebasing and re-reviewing against the merged state.
- **The data item signature check is disabled.** See the security section. Restoring it needs gateways that return `anchor`/`recipient`, and a transport that is not backstopped by a rate-limited `arweave.net`.
- **Downloads walk gateways in order rather than racing them.** Hedging was lost with the removal of the client's `download()` and is worth restoring on the direct path.
- **A private thumbnail still costs one GraphQL call.** The link carries the cipher for the *data* transaction; a thumbnail is its own transaction with its own IV.
- **`bundledIn` is carried in links but not used as a locator.** If a data transaction cannot be fetched from any gateway, the bundle it sits in is a second route to the same bytes.
- **GraphQL reads are bounded but not cancellable.** A timeout ends the wait, not the request.


vilenarios:
Superseded by #2191, which goes `dev` → `master` directly.

This branch was cut on 2026-08-11 and had drifted **9 commits** behind `dev` — missing the entire shared file page rebuild (#2189), the sharing hardening (#2186), all five PE-9205 sync fixes, and PE-9204's per-drive isolation. Merging it as-is would have shipped a v2.86.0 containing none of the sharing work and none of the download fix.

The description here was the starting point for #2191's notes, with one correction carried over: the chain-based integrity check it promised for AES-CTR and public downloads **shipped disabled**, because gateways return empty `anchor`/`recipient` and it reported healthy files as corrupt. AES-GCM authentication is unaffected and still enforced. Details in #2191.

PR #2174, opened 2026-07-30
PE-9132: Read Turbo free-item cap from /v1/info (fix 105 KiB free cap)

vilenarios:
## Summary

`TurboUploadService.refreshMaxItemBytes` read a **top-level `maxItemBytes`** key that `/v1/info` doesn't return — the per-item free cap is nested under **`freeTier.maxItemBytes`** (with a top-level `freeUploadLimitBytes` mirror). So the server value was never applied and the app silently fell back to the config value, enforcing a free-item cap of **100000 bytes (97.66 KiB) instead of Turbo's actual 107520 (105 KiB)**.

Effect: files between ~97.7 KiB and 105 KiB were charged when Turbo would have made them free. Surfaced during staging upload testing (100 KiB files hit "insufficient balance").

## Fix
Read `freeTier.maxItemBytes` → top-level `freeUploadLimitBytes` → `maxItemBytes` (defensive), accepting `int`/`num`, staying on the config fallback when absent or on error.

## Tests
7 unit tests: the real payload shape (proves the fix), each fallback, decoded-Map input, no-cap fallback, fail-safe on error, non-positive ignored.

Pre-existing bug (the server-cap fetch never worked), independent of the 2.86.0 release — can ship with it or as a fast-follow.

https://claude.ai/code/session_01PsV2WFHZMGD9DUfX65cxuW

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated free-upload size limits to use the server’s current limit when available.
  - Added support for multiple server response formats and fallback limit sources.
  - Invalid, missing, zero, negative, fractional, or unavailable limits no longer replace a valid existing limit.
  - Valid positive whole-number limits are now applied consistently.
  - Improved logging when retrieving the server limit.
  - Preserved the last known-good limit when later requests fail.

### DIFF
--- a/lib/turbo/services/upload_service.dart
+++ b/lib/turbo/services/upload_service.dart
@@ -149,12 +149,22 @@ class TurboUploadService {
       // code only checked top-level `maxItemBytes`, which the current service
       // does not return, so it silently fell back to the config value.
       final freeTier = data is Map ? data['freeTier'] : null;
-      final value = (freeTier is Map ? freeTier['maxItemBytes'] : null) ??
-          (data is Map
-              ? (data['freeUploadLimitBytes'] ?? data['maxItemBytes'])
-              : null);
-      if (value is num && value > 0) {
-        _serverMaxItemBytes = value.toInt();
+
+      // Each candidate is validated *before* the priority order is applied,
+      // not after. `??` picks the first non-null, so a present-but-unusable
+      // `freeTier.maxItemBytes` - a string, a negative, a zero - used to mask
+      // a perfectly good `freeUploadLimitBytes` sitting beside it and drop the
+      // service back to the stale config value this fix exists to stop using.
+      final value = [
+        freeTier is Map ? freeTier['maxItemBytes'] : null,
+        data is Map ? data['freeUploadLimitBytes'] : null,
+        data is Map ? data['maxItemBytes'] : null,
+      ]
+          .map(_wholePositiveBytes)
+          .firstWhere((bytes) => bytes != null, orElse: () => null);
+
+      if (value != null) {
+        _serverMaxItemBytes = value;
         logger.d('Turbo free item size limit from /v1/info: '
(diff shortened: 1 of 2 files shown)


vilenarios:
Addressed in `97af1d69` — two real defects fixed, one recommendation declined with reasoning. Analyze clean, 14 tests in this file, 747 in the suite.

**Validate before choosing — correct, and the sharper of the two.** The candidates were selected with `??` and validated *afterwards*, so any non-null preferred value won even when it was unusable. A payload like `{"freeTier":{"maxItemBytes":"oops"},"freeUploadLimitBytes":107520}` ignored the perfectly good number sitting beside it and fell back to the stale config value this PR exists to remove. Each candidate is now validated first and the first *valid* one wins, in the same priority order.

**Fractional values — correct, and worse than it looks.** `0.5` passed `> 0` and `toInt()` made it **0**. A zero cap is not the same as no cap: `maxFreeItemSizeBytes` returns `_serverMaxItemBytes ?? allowedDataItemSize`, so a zero would be returned *in preference to* the fallback and nothing at all would qualify for a free upload. Fractional values are now rejected rather than truncated. I also added a guard the review didn't mention: infinity and NaN reach `toInt()` under the old check, and `toInt()` throws on both.

**Declined: resetting `_serverMaxItemBytes` to null on error or on an unusable payload.** That would make a transient `/v1/info` failure silently reintroduce the stale 105 KiB cap this PR removes — and a cap the server actually reported is better information than the config value, which does not become more accurate because one request failed. On the first call the field is already null, so the reset only ever changes behaviour *after* a success, which is precisely the case where it does harm. Two tests pin the choice: a later failure and a later capless payload both keep the last good cap.

New regressions: unusable preferred value with a valid fallback, negative preferred value with a valid fallback, fractional-to-zero, large fractional, whole-number-as-double (JSON has one number type, so `107520.0` is a realistic payload), and the two persistence cases above.

PR #2175, opened 2026-08-06
PE-9200: Self-describing share links, an authenticated download path, and a recipient page built for strangers

vilenarios:
Rebuilds ArDrive's file sharing around a self-describing link, a recipient page written for someone who has never heard of ArDrive, and a download path that actually authenticates what it saves.

Design docs are in the branch: `docs/FILE_SHARING_VIEW_UX_REVIEW.md` (evidence) and `docs/FILE_SHARING_REDESIGN_PLAN.md` (design + phases).

## What changes for users

**The key is no longer in the link by default.** It travels out of band as a separate artifact, and the share dialog hands over link and key separately with copy pushing them down different channels. Key-in-link is an opt-in, off by default.

**A share link now carries what the page needs** — data and metadata transaction ids, owner, name, size, type, cipher, IV — so a complete link paints with **zero blocking GraphQL**. Every field degrades independently; a field that is missing or arrived mangled falls back to exactly the lookup a v1 link would have done. v1 links resolve exactly as they always have, and always will.

**The recipient page** is a real state machine: locked, resolving, ready, not found, network trouble, damaged link. No eternal spinners, a wrong key says so instead of "file does not exist", a not-yet-propagated upload retries instead of claiming the file is missing, and transaction ids live behind a drawer.

**Previews**: PDFs render inline (rasterised to images, never a frame), private video and audio preview on web, and recipients see real thumbnails.

**A generalized viewer** at `/view/{txId}` makes ArDrive a friendly front end for any Arweave transaction.

## Security

- **Private downloads on web were not authenticated, at any size.** Both ciphers were routed through a streaming decryptor whose GCM branch trims the MAC and never checks it. GCM now buffers, verifies, and writes nothing on failure; `AesGcmStream` is gone from the download path and disarmed. For CTR and public files, integrity now comes from the chain: the ANS-104 deep hash is computed over the ciphertext as it streams and checked against the data item's signature. A resumed download reports "could not verify" rather than guessing, and an Ethereum- or Solana-signed item does too — only bytes that contradict the signature report failure.
- **The vendored pdf.js was vulnerable to CVE-2024-4367** — a crafted PDF could run arbitrary JavaScript on the app origin, which is where a recipient's access key can sit. Mitigated in the vendored build. **The real fix is pdf.js ≥ 4.2.67, which needs a Flutter newer than the pinned 3.19.6.**
- **plausible.js was sending `location.href` — fragment included — to plausible.io.** On a key-in-link share that string is the file key. The tag is removed.
- **The "Verified" badge was forgeable** by anyone able to post a transaction: the link's `mtx` was fetched with no owner constraint. Authorship is now checked, without adding a pre-paint round trip.
- Untrusted transaction bytes never become script-capable content on the app origin: only bytes unlock an inline renderer, a declared type can only ever restrict, and HTML/SVG are offered on the gateway's own per-transaction origin inside a sandbox without `allow-same-origin`. `openUrl` now accepts only http/https, which also protects two pre-existing callers that pass non-constant URLs.

## Measured, not assumed

`Range` support is **not** universal — `turbo-gateway.com` returns `206`, `arweave.net` ignores the header and returns the whole body. So resume branches on the *response*, never on having sent the header: a `200` body starts at byte zero, and writing it at the resume offset would silently corrupt the file.

## Reviewer notes

- **Path routing is built and switched off.** It needs SPA rewrites that do not exist yet; enabling it is two lines and a test fails if only one moves. `PathUrlStrategy` also throws without a `<base>` element, and adding `base href="/"` breaks sub-path serving — so it is only viable from an origin root. Full prerequisite list in `lib/utils/app_url_strategy.dart`.
- **Mobile is unbuilt.** `build-ios`/`build-android` are `if: false` in CI. The mobile download fix (which also repairs a hang that predates this branch) is unit-tested but has not run on a device, and `pdfx` brings an Android module declaring `compileSdkVersion 31`.
- **One manual check worth doing:** the GCM path buffers a private file fully before requesting the save location, which may fall outside Chrome's transient-activation window. Affects private web downloads under 100 MiB.
- **Translations are machine-produced** and need a native-speaker pass, particularly the security copy.
- GraphQL retries drop 8 → 5. Eight is ~51s of backoff before the fallback endpoint is even tried.

## Verification

`flutter analyze` clean. 1,183 main-app tests, 43 in `ardrive_crypto`.

Three adversarial reviews (security, regression, test quality) ran over the whole diff and their findings are remediated in the final commit — including several tests that passed without proving anything, each now failing under the specific mutation that used to slip through.

## Summary by CodeRabbit

* **New Features**
  * Introduced redesigned file-sharing links with separate access keys, metadata, version pinning, and legacy-link compatibility.
  * Added clearer shared-file landing pages, unlock flows, verification indicators, revision history, retries, and remembered session keys.
  * Added raw transaction viewing with safe previews for images, PDFs, text, media, and sandboxed content.
  * Added inline PDF previews, thumbnail support, gateway fallbacks, and resumable downloads.
* **Bug Fixes**
  * Improved malformed-link handling, download integrity checks, preview size limits, URL safety, and localized download messaging.
* **Documentation**
  * Added detailed file-sharing redesign and user-experience documentation.


vilenarios:
@coderabbitai review

The review was skipped at 272 files. I've added a `.coderabbit.yaml` excluding vendored and generated paths — 173 of those files are the vendored pdf.js distribution (168 of them binary CJK `.bcmap` character maps), plus generated GraphQL and Drift artifacts. The hand-written diff is far smaller.

If the file count is still over the limit after filtering, say so and I'll split this into per-phase PRs instead.

vilenarios:
Rebased onto #2176 and retargeted: **99 files**, under CodeRabbit's limit.

#2176 carries only the vendored pdf.js build and the `pdfx` dependency — third-party assets with nothing to review line by line. This PR is now stacked on it and should merge after it.

Also dropped the 168 CJK `.bcmap` character maps in the process: 1.3 MB on a build that is permanent on Arweave, for a narrow class of older PDFs that use predefined CJK encodings. Anything with embedded fonts — the modern norm — is unaffected.

@coderabbitai review

vilenarios:
@coderabbitai review

CI is green here and on the base PR #2176. This PR is 99 files, under the limit. The earlier trigger hit your rate limit before the review ran.

vilenarios:
@coderabbitai full review

Incremental review won't help here: the automatic review already ran on these commits and skipped them when the PR was 273 files, so there are no *new* commits to look at — the diff just got smaller (99 files) by moving the vendored pdf.js into base PR #2176. A full review should pick it up.

vilenarios:
@coderabbitai full review

vilenarios:
All nine addressed in `9f5ac212` — seven fixed, two partially rejected with reasons. Analyze clean, 1206 main-app tests, 43 crypto tests.

**Fixed**

- **`DownloadIntegrityException` → `unknownError`** — real and user-visible. A file that failed its MAC check offered *Try Again*, which cannot help. New `integrityCheckFailed` reason renders a non-retryable modal saying nothing was saved; string localized into all six locales.
- **Unguarded limit check** — the future is discarded from the constructor, so a throwing `singleFileLimit` left the dialog on "starting" forever. Now try/caught, matching `ProfileFileDownloadCubit`.
- **Clock-derived `viewType`** — the sharpest catch of the nine. Millisecond resolution on web means `microsecondsSinceEpoch` is always a multiple of 1000, so two frames in the same millisecond collide. Now a monotonic counter, and the test strips the timestamp so the old implementation fails deterministically rather than depending on clock luck.
- **Eagerly acquired resources released only from a lazy generator** — correct, and it was a bug I introduced: the eager open exists so a dead gateway is reported *before* the save picker opens instead of as "Download cancelled" after. Kept the eager open, threaded the release through every settle path. Covered by a test using a dismissed-picker saver that never reads the stream.
- **Pending key surviving "forget"** — a real privacy defect. Test: tick, paste, untick, succeed → storage empty.
- **Stale remembered-key read** — now bails unless still locked, no unlock in flight, and the field empty. Worth noting the window is a single microtask with the real store, so it was latent rather than reachable today.
- **Oversized `thn`** — and the timeouts did *not* bound it: the per-gateway timeout doesn't cancel the buffered fetch beneath it, and the waterfall then starts the next gateway alongside it. Now a Range-bounded fetch that rejects a declared oversize without reading a byte. I did not take the suggested "fetch the size first" variant — `TransactionCommon.graphql` has no `data { size }`, and adding a GraphQL round trip before every thumbnail is exactly what this schema exists to avoid.
- **`openUrl` double-reporting** — it also threw into ~30 fire-and-forget call sites, where it became an unhandled async error and the user saw a dead link. Now returns `false`. The same duplicate-log pattern existed in `pdf_preview_widget.dart`, which the review only caught in one of the two places.
- **Boot shim logging a caught error** — the URL it parses ends in `#k=<key>`. Static message now.

**Rejected**

- **A `dispose` on `StreamedDataItemVerifier`** — `markUnverifiable` already is that entry point: it cancels the feed, which unwinds the verify and settles the verdict. A second name for one operation invites abandoning a check twice. Documented the dual role instead.
- **The `downloadToMemory` half of the resource finding** — that path listens on the statement immediately after the stream is built, with nothing in between that can throw, so the release could never run. A safeguard that is provably dead is worse than none; added a comment saying why.

PR #2176, opened 2026-08-06
PE-9200: Vendor pdf.js and add the pdfx dependency

vilenarios:
Split out of #2175 so that PR fits under CodeRabbit's 100-file limit. This carries only third-party assets and the dependency change — the code using them lives in #2175, which is based on this branch and should merge after it.

- **`pdfx: 2.6.0`**, pinned exactly. 2.7.0+ need `meta ^1.15.0` against the `meta 1.11.0` that Flutter 3.19.6 pins *exactly*, and 2.9.x additionally need Flutter ≥ 3.24.0 — so this is the newest version that resolves here. Verified: no other dependency moves.
- **pdf.js 2.12.313 vendored**, never a CDN. This build ships to Arweave where it is permanent and could never be patched, and a CDN script tag also breaks under the strict CSP an AR.IO gateway serves. Worker and cmap paths resolve from the script's own URL, so they survive being served under a `/{txid}/` subpath.
- **CJK cmaps deliberately excluded** — 1.3 MB across 168 files for a narrow class of older PDFs that use predefined CJK encodings. Anything with embedded fonts is unaffected.
- **CVE-2024-4367 mitigated**: `isEvalSupported` is patched to default `false` in the vendored build. This version compiles Type1 glyph paths with `new Function` from a `FontMatrix` it does not type-check, and the PDFs reaching it are attacker-chosen. Patched in the file rather than passed as an option because pdfx merges its option map with `Map.addAll` over a raw JS object, which throws under dart2js. **Re-apply on any rebuild, or move to pdf.js ≥ 4.2.67 once a Flutter upgrade lets pdfx target it.**

Cost, stated plainly: 1.3 MB permanent, of which `pdf.min.js` (238 KB) loads on **every page** — login included — because the pdfx web plugin asserts `pdfjsLib` exists at plugin-registration time. Only the 1 MB worker is deferred to when a PDF is actually opened.

## Summary by CodeRabbit

* **New Features**
  * Added PDF viewing support on macOS and web platforms.
  * Added web URL strategy support for improved navigation.
  * Added bundled PDF.js configuration and worker support for consistent rendering.
* **Security**
  * Updated PDF rendering configuration to mitigate a known PDF.js vulnerability.
* **Documentation**
  * Included PDF.js licensing information and compatibility documentation.
* **Reliability**
  * Added automated checks to ensure PDF.js assets remain bundled locally and securely configured.


vilenarios:
Thanks — two of three fixed in `edc445a8`, third is an artifact of the split.

**1. `pdfjs_config.js` excluded from review — fixed.** Correct catch: `!web/js/pdfjs/**` swept up our own file along with the vendored bundles. Narrowed to `*.min.js`, `LICENSE` and `cmaps/**`, with a note saying why `pdfjs_config.js` is deliberately left in scope.

**2. CVE patch has no CI guard — fixed, and this was the best finding of the three.** `test/web/vendored_pdfjs_test.dart` now fails if the vendored bundle drifts back to `isEvalSupported=!0`, with a message pointing at the two ways out (re-apply, or move to pdf.js ≥ 4.2.67). Verified by mutation: reverting the patch turns it red, restoring it turns it green. The comment in `pdfjs_config.js` asked the next person to remember; this asks CI instead.

**3. `index.html` doesn't load `pdfjs_config.js` — correct for this PR, and intentional.** This PR carries only the vendored assets and the `pdfx` dependency so that #2175 fits under the 100-file review limit. The script tags, the routing, and the viewer all live in #2175, which is *based on this branch* and merges after it. This PR is deliberately inert on its own — nothing here is reachable until #2175 lands. Worth confirming on the merged result rather than here.

PR #2179, opened 2026-08-10
PE-9203: Sync reads a single gateway, and stops letting one slow item block four

vilenarios:
Sync fetched every metadata item through the multi-gateway waterfall — configured gateway, up to two GAR gateways, then arweave.net, **serially, 5s each**. A flaky gateway made each item cost 20s, and a sync reads hundreds. Building that gateway list also cost a Solana RPC.

## What changed

**Sync reads one gateway.** Two attempts at 5s, then skip. A 404 isn't retried — the same host won't change its mind. **20s → 10.3s per failing item**, and the Solana RPC is gone from the per-item path.

**The bigger win was scheduling.** The concurrency limiter was a barrier, not a pool:

```dart
for (start = 0; start < items; start += 5)
  await Future.wait(...five items...);   // waits for ALL five
```

Each batch waited for its slowest member, so one 10s failure idled four workers for it. A sliding window keeps five in flight continuously.

| 300 items, 10% failing | wall clock |
|---|---|
| chunked barrier | 321.2s |
| sliding-window pool | **82.9s** |
| | **3.9× faster** |

Both schedulers produce byte-identical positional output — `entityDatas` is index-filled, and there's a test for out-of-order completion.

**Concurrency stays at 5.** The pool is the structural fix; raising the number is an independent knob. Turning both at once — while also removing the fallback — would raise sustained pressure on a single gateway at exactly the moment we removed the thing that absorbs a 429. One knob, measure, then the other.

**Snapshot validation loses its GAR branch.** It already fast-failed on 404; the remaining path existed only for transient errors, at the cost of that same RPC.

**The waterfall is untouched everywhere else** — downloads, previews, manifests, shared links, drive attach. A recipient with one dead gateway still gets their file. Caller audit is in the commit message.

## The thing this PR found

**Sync already drops files silently, today, on `dev`.** A failed metadata read returns `Uint8List(0)` rather than an error; those empty bytes fail to parse; the parse exception is logged as a warning; the entity never reaches the drive — and the watermark advances regardless. The block-height rewind (`kBlockHeightLookBack = 240`) is a **~2 hour recency window, not a retry mechanism**, and `_lastSync` is in-memory, so anything older than that is gone until a manual deep sync.

Skipped transaction ids and their drive ids now ride out on the sync result instead of vanishing. `docs/SYNC_SKIPPED_ENTITY_PERSISTENCE.md` sets out the fix — a table, a block-height range union, a backoff.

## No skipped-files UI here, deliberately

Per-file state can't be coherent while the list lives only in memory: rows would appear after one sync and vanish on restart, which is worse than none. It lands with the persistence that makes it stable. Worth knowing for the design — the GraphQL tags give us entity type, file id and parent folder id, but **not the name, size or dataTxId**, since those live in the body we failed to read. So the row can sit in the right folder, but it can't be named or downloaded.

## Verification

`flutter analyze` clean · 755 tests passing · 22 new, including `verifyNever(getGateways)` on the sync path and proof the waterfall paths still consult it.

## Summary by CodeRabbit

- **New Features**
  - Sync results now report entities whose metadata could not be retrieved, including counts and transaction details by drive.
  - Metadata requests use bounded concurrency and targeted retries to improve synchronization reliability.
  - Snapshot validation checks the configured gateway directly with clearer failure handling.

- **Bug Fixes**
  - Failed metadata reads are skipped instead of being treated as empty entity data.

- **Documentation**
  - Added design documentation for tracking and recovering skipped entities.

### DIFF
--- a/docs/SYNC_SKIPPED_ENTITY_PERSISTENCE.md
+++ b/docs/SYNC_SKIPPED_ENTITY_PERSISTENCE.md
@@ -1,0 +1,243 @@
+# Sync: skipped entities and why they can be lost permanently
+
+Status: design note for a follow-up PR. Nothing here is implemented yet.
+
+This note exists because of a bug found while making sync read from a single
+gateway (PE-9203). The bug **predates that change** and is invisible unless you
+go looking for it, so it is written down here in full.
+
+---
+
+## 1. The pre-existing silent drop
+
+When a sync cannot read an entity's metadata, that entity is dropped from the
+user's local drive state, silently, and in the common case permanently. There is
+no error surfaced, no retry across sessions, and no record that anything is
+missing.
+
+### Evidence
+
+**A failed read becomes empty bytes, not an error.**
+`ArweaveService._getEntityData` (`lib/services/arweave/arweave_service.dart`)
+historically ended with:
+
+```dart
+return getEntityDataFromNetwork(txId: txId).catchError((e) {
+  logger.e('Failed to get entity data from network', e);
+  return Uint8List(0);          // <-- failure becomes empty bytes
+});
+```
+
+**The empty bytes then fail to parse, and the parse error is swallowed.**
+In `createDriveEntityHistoryFromTransactions`, `FileEntity.fromTransaction(tx,
+Uint8List(0), ...)` throws `EntityTransactionParseException`, which is caught by
+the `on EntityTransactionParseException` handler and logged at warning level.
+The entity is never added to `blockHistory`, so it never reaches the database.
+
+**The drive's watermark advances anyway.**
+`_parseDriveTransactionsIntoDatabaseEntities`
+(`lib/sync/domain/repositories/sync_repository.dart`) writes
+`lastBlockHeight: Value(currentBlockHeight)` — both in the empty-transactions
+branch and in `endOfBatchCallback` — with no knowledge that anything was
+dropped. The next sync therefore starts *after* the block containing the item it
+failed to read.
+
+**The look-back only helps within a session boundary.**
+`_calculateSyncLastBlockHeight` rewinds by `kBlockHeightLookBack`
+(`lib/sync/constants.dart`, currently 240 blocks ≈ 2 hours) — but only when
+`_lastSync == null`:
+
+```dart
+int _calculateSyncLastBlockHeight(int lastBlockHeight) {
+  if (_lastSync != null) {
+    return lastBlockHeight;                                   // no rewind
+  }
+  return max(lastBlockHeight - kBlockHeightLookBack, 0);      // rewind
+}
+```
+
+`_lastSync` is a plain in-memory `DateTime?` on `_SyncRepository`. It is never
+persisted. So the rewind happens once per app session and covers only the last
+~2 hours of chain history.
+
+### What that means in practice
+
+| Situation | Recovered? |
+|---|---|
+| Item dropped, another sync runs in the same session | No — `_lastSync != null`, no rewind |
+| Item dropped, app restarted within ~2h of that block | Yes — first sync of the session rewinds 240 blocks |
+| Item dropped, app restarted more than ~2h later | **No — lost until a deep sync** |
+| User triggers a deep sync | Yes — `syncDeep` passes `lastBlockHeight: 0` |
+
+Deep sync is the only reliable recovery, it is user-initiated, and nothing tells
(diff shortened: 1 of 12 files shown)


vilenarios:
Went through all three. **One was right and is fixed in `a384d78e`; two I'm rejecting with evidence.**

### Fixed — vacuous snapshot validation test

Correct, and the sharpest of the three. `validateSnapshotItems([])` returns before the loop is entered (`snapshot_validation_service.dart:25-27` — `while (remaining.isNotEmpty)`), so the test made no HEAD request and exercised nothing it was named for.

It now passes a snapshot item, so the unreachable configured gateway is genuinely contacted and genuinely rejects — visible in the test output as `OS Error: The remote computer refused the network connection`, which the empty-list version never produced. That is exactly the point where `dev` would have reached for the GAR list.

I kept `verifyZeroInteractions(arioSDK)` but labelled it honestly: the service is never handed that SDK, so the assertion is decorative. The **compile-time** proof is the real one — `SnapshotValidationService(configService: ...)` only builds because `arioSDK` is no longer a parameter.

### Rejected — `.toInt()` on the clamped concurrency

`clamp` returns `num` on `num`, but Dart special-cases it: *the static type of `e.clamp(a, b)` is `int` when `e`, `a` and `b` are all `int`.* `maxConcurrentDataFetches` is `int` and the bounds are int literals, so both results are already `int`.

The proof is that the branch analyzes clean. `concurrency:` on `runPooled` is `int`, and `b +=` / `skip` / `take` all require `int` — if either value were `num`, those would be compile errors, not lint. Adding `.toInt()` would be a no-op on an already-`int` expression.

### Rejected — static gateway cache leaking across tests

`cachedGateways` is an **instance** field, not static — `data_gateway_fallback.dart:51`, `List<Gateway>? cachedGateways;` with no `static` modifier.

And `fallback = DataGatewayFallback(arioSDK: arioSDK)` is inside `setUp()` (`data_gateway_fallback_test.dart:35`), so every test gets a fresh instance with a fresh null cache. There is no order dependency between the tests at lines 144 and 156.

---

`flutter analyze` clean; sync and arweave suites green (67 tests).

vilenarios:
**Correction — you were right, and there was still a waterfall in the sync path.**

`getUniqueUserDriveEntities` — sync's drive-discovery phase — was calling `fetchData` (the full waterfall) **once per drive transaction, in parallel**:

```dart
driveTxs.map((e) => _gatewayFallback.fetchData(e.id, client)...)
```

So a user with a dozen drives opened a dozen fan-outs to GAR gateways on every sync. That is precisely the cost this PR exists to remove, sitting in the one sync path that was left untouched.

**The reason it was left is wrong.** It was believed to be shared with login. It is not — `getUniqueUserDriveEntities` has exactly one caller in the whole codebase, `_SyncRepository.updateUserDrives` (`sync_repository.dart:1294`). The login path is `getDriveSignatureForDrive` (`ardrive_auth.dart:323`), a different method, which keeps the waterfall and should.

Fixed in `5e99f7ac`. I also re-audited the rest by grep rather than trusting the earlier table: `getLatestDriveEntityWithId`, `getAllFileEntitiesWithId`, `getLatestFileEntityWithId` and `dataFromTxId` are **not reachable from `lib/sync/` at all**. Drive discovery was the last one.

So the claim in the PR description now actually holds: **no fallback fan-out and no Solana RPC anywhere in a sync.**

`flutter analyze` clean · 755 tests passing.

vilenarios:
**Valid, and fixed in `735c9b26`.**

Drive discovery fetches a drive signature for every private drive whose key isn't already in memory, and it did so via `getDriveSignatureForDrive` — the login path, which uses the waterfall. So the fan-out was back in the sync path by the side door, once per private drive.

Both readers now share one implementation and differ only in how they fetch:

- `getDriveSignatureForDrive` — waterfall, unchanged. This is `ArDriveAuth`, where one unreachable gateway must not cost someone their session.
- `getDriveSignatureForDriveOnSync` — configured gateway only, used by drive discovery.

Added the coverage you asked for: a test asserting the sync-path signature read never contacts the GAR, alongside the existing pair that proves `fetchData` still does.

**Worth recording why this survived two prior audits.** Both greps that hunted for waterfall callers excluded `arweave_service.dart` itself — and this call site is *inside* that file. A search that assumes the caller lives elsewhere cannot see an internal one. Two of the three misses in this PR share that root cause.

`flutter analyze` clean · 756 tests passing.

PR #2180, opened 2026-08-11
PE-9204: One drive's gateway hiccup killed the whole sync

vilenarios:
Reported from staging: a drive signature 404ed and **sync stopped dead before any drive had synced**. Retrying by hand found the transaction present — the gateway had it, it just hadn't indexed it yet.

Two causes.

## The failure wasn't contained

Drive discovery already skips a drive whose metadata can't be read (`if (driveResponses[i] == null) continue`). But the drive-signature read a few lines below was **the one unguarded `await` in that loop**. It threw past every handler to the method's outer catch — which rethrows — so `getUniqueUserDriveEntities` failed, `updateUserDrives` failed, and sync died at discovery.

One drive's bad luck took every drive with it, including public ones that would have synced fine. It now drops that drive for the pass, like every other failure there, and picks it up next time.

## A 404 was treated as final, and that was wrong

The 404 fast-path reasoned that *the same host will not turn a 404 into a 200*. That holds for a host that has answered definitively. It does **not** hold for one still indexing — which is precisely what was observed: 404 on the first read, present on the second.

A 404 now spends the second attempt like any other failure and is only reported as `TransactionNotFound` when every attempt agrees. A genuinely missing transaction still costs two attempts, no more.

## Why this got worse recently

Before #2179 this path used the multi-gateway waterfall, so a gateway that was a beat behind was covered by the next one. Single-gateway reads removed that cushion, and the 404 fast-path removed the retry too — so an indexing lag went straight to a hard failure with no second chance.

## Verification

`flutter analyze` clean · 1,320 tests passing.

The test asserting *"does not retry a 404"* encoded the behaviour this reverses; it's replaced by two — one proving a 404-then-200 succeeds, one proving a persistent 404 still raises `TransactionNotFound` in exactly two attempts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved private-drive discovery so a failed signature request no longer stops the entire discovery process.
  * Added retry handling for temporary 404 responses when fetching data from the configured gateway.
  * Persistent 404 responses still report that the requested transaction could not be found.

### DIFF
--- a/lib/services/arweave/arweave_service.dart
+++ b/lib/services/arweave/arweave_service.dart
@@ -1025,10 +1025,27 @@ class ArweaveService {
             final sigTypeTag = driveTx.getTag(EntityTag.signatureType) ?? '1';
             final signatureType = DriveSignatureType.fromString(sigTypeTag);
 
-            final driveSignature = signatureType == DriveSignatureType.v1
-                ? await getDriveSignatureForDriveOnSync(
-                    wallet, driveTx.getTag(EntityTag.driveId)!)
-                : null;
+            // Contained deliberately. This was the one unguarded await in the
+            // loop, and a gateway hiccup on a single drive's signature threw
+            // all the way out of drive discovery and killed the entire sync -
+            // before any drive had synced at all. Every other failure here
+            // drops one drive and carries on; this one now does too, and the
+            // drive is picked up on the next pass.
+            DriveSignatureEntity? driveSignature;
+
+            if (signatureType == DriveSignatureType.v1) {
+              try {
+                driveSignature = await getDriveSignatureForDriveOnSync(
+                    wallet, driveTx.getTag(EntityTag.driveId)!);
+              } catch (e) {
+                logger.w(
+                  'Could not read the drive signature for '
+                  '${driveTx.getTag(EntityTag.driveId)}; skipping this drive '
+                  'for this pass: $e',
+                );
+                continue;
+              }
+            }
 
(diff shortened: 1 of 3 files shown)


vilenarios:
All three valid — fixed in `e2268785`. Analyze clean, 1,322 tests passing.

**Cross-page dedup — and this one was caused by this PR.** `getUniqueUserDriveEntityTxs` dedupes *per page*, so the same Drive-Id can appear on several pages, newest first. The `continue` I added on a signature failure skipped the newest transaction **without recording the id**, so the older one on a later page was no longer deduped and would write stale metadata. That's worse than the drive being late — a silently outdated drive beats a temporarily absent one. The id is now claimed on failure, so nothing older stands in for it.

**`TransactionNotFound` claimed more than it knew.** My check was "the last attempt was a 404", so 500-then-404 reported the transaction as absent. That's an unwell gateway, not missing data. Every attempt must now agree, and a timeout or socket error disqualifies it too.

**The doc was stale** — it still said a 404 isn't retried, which this PR had already made untrue.

Added both scenarios you asked for: 500-then-404 and exception-then-404, each asserting a non-`TransactionNotFound` failure in exactly two gateway calls.

I did not add the cross-page regression test — `getUniqueUserDriveEntities` needs a wallet, crypto, a DriveDao and paginated GraphQL to construct, so a faithful test is a fixture exercise rather than a small one. The guard is three lines and covered by inspection; flagging that trade rather than claiming coverage I didn't write.

PR #2187, opened 2026-08-19
PE-9205: Design — the drive state artifact

vilenarios:
Design only, no code. A published, encrypted, parsed-state blob that a client imports in bulk instead of replaying a drive's history entity by entity.

**Additive by construction.** Snapshots remain the ArFS interchange format and the fallback; an unknown `State-Version` is inert, so old clients and other implementations are unaffected.

## The case, measured

Consuming a snapshot costs **per entity**: one AES decryption (`file_entity.dart:117`), one JSON parse, one insert. ~42,000 of each on the drive we've been testing — the ~80s of `Processing chunk of 1000 transactions` in its log. An artifact costs **one** decryption and a bulk insert.

| | on disk | gzipped |
|---|---|---|
| State artifact (42k files) | **31.31 MiB** | **6.41 MiB** |
| Snapshot `z78YIh…` (same drive) | 43.92 MiB | ~11.37 MiB |

The artifact figure is a Drift database of 42,000 files built, `VACUUM`ed and weighed — a model, not an export of a real user DB. Snapshot figures are the real transaction, with gzip measured on a sample of it.

And producing it is cheap in the way snapshots are not: it's already in the local database, so there's no chain re-read.

## Why security leads the document

The obvious implementation does permanent harm. The local DB holds:

```
profiles.encryptedWallet   +   profiles.keySalt      ← complete offline attack
drives.encryptedKey        +   drives.keyEncryptionIv    package on the password
```

Serialising the database publishes those **permanently**, with no delete. So: export from an explicit allowlist, never dump — guarded by a test that fails when the schema grows a table or column, because the realistic failure is a *future* migration, not this design today.

Two more that shape the format:

- **Authenticated encryption is mandatory**, and `maxSizeSupportedByGCMEncryption` is 100 MiB — a ~130k-file drive crosses into unauthenticated CTR, where a flipped bit becomes silently wrong local state.
- **Don't hand an untrusted file to SQLite.** Ship serialised rows, not a `.db`. This also avoids welding the wire format to `schemaVersion` — two independent reasons for the same call.

## Also covered

- **Sync** — the artifact is one more obscuring range; `HeightRange`/`obscuredBy` need no change. Import is a merge, not a replace. A sync reporting skipped entities must not produce one.
- **ardrive-core-js / CLI** — nothing breaks; the CLI is the more interesting *producer*, having real bandwidth and no browser connection limits. Reasoned from the protocol boundary; those repos weren't read.
- **ArNS** — already-built plumbing. `ArnsRepository.setUndernamesToFile` points an undername at an arbitrary `txId` today, so `state_<driveId>.myname → latest artifact` replaces a paginated query with one resolution. Caveats included: it's optional, updates cost a transaction, and a mutable pointer is a different trust object from an immutable transaction.
- **Relationship to the other two snapshot changes** — incremental snapshots is the cheapest of the three; this one is the largest and the only one that removes the 42,000 decryptions.

Every code reference was re-verified against `dev` before publishing. Open questions are listed rather than answered.

https://claude.ai/code/session_01DnYLXFocWgTt9M2CbGYSUP

## Summary by CodeRabbit

* **Documentation**
  * Added a design document describing encrypted drive-state artifacts.
  * Documented secure export and import requirements, versioning, synchronization, merging, rollback protection, and compatibility expectations.
  * Outlined artifact discovery and open implementation questions.

### DIFF
--- a/docs/DRIVE_STATE_ARTIFACT.md
+++ b/docs/DRIVE_STATE_ARTIFACT.md
@@ -1,64 +1,99 @@
 # The drive state artifact
 
-A published, encrypted, parsed-state blob that a client imports in bulk instead
-of replaying a drive's history entity by entity.
+A published, encrypted, signed blob of a drive's parsed state, which a client
+imports in bulk instead of replaying the drive's history entity by entity.
 
 **Additive.** Snapshots remain an ArFS interchange format and the fallback. A
-client that does not understand a state artifact, or cannot decrypt one, syncs
-exactly as it does today.
+client that does not understand a state artifact, cannot decrypt one, or never
+finds one syncs exactly as it does today.
 
 **The governing principle: this is a cache, and every failure is a fallback.**
-Unknown version, failed decryption, failed integrity check, missing tag,
-unexpected row - each means *ignore the artifact and sync normally*, never
-*fail the drive*. That single rule is what makes the addition foolproof: the
-worst outcome of anything going wrong is today's performance.
+Unknown version, failed signature, failed decryption, failed integrity check, an
+unexpected section — each means *ignore the artifact and sync normally*, never
+*fail the drive*. The worst outcome of anything going wrong is today's
+behaviour. Every design decision below follows from that.
 
 ---
 
-## 1. Why
+## 1. The case
 
-Sync is slow on large drives, and the cost is not the download.
+Three arguments, in increasing order of how much they matter.
 
-Consuming a snapshot costs, **per entity**: one AES decryption
-(`file_entity.dart:117`, inside `FileEntity.fromTransaction`), one JSON parse,
-one database insert. On a 42k-item drive that is ~42,000 of each — the ~80
-seconds of `Processing chunk of 1000 transactions` observed in a real sync.
-
-A state artifact costs **one** decryption and a bulk insert.
-
-### Measured size
+### 1.1 Size — the weakest of the three
 
 | | on disk | gzipped |
 |---|---|---|
-| State artifact (42k files) | **31.31 MiB** | **6.41 MiB** |
-| Snapshot `z78YIh…` (same drive) | 43.92 MiB | ~11.37 MiB |
-| | 1.40× smaller | 1.77× smaller |
+| State artifact | **34.63 MiB** | **6.65 MiB** |
+| Snapshot `z78YIh…` | 43.92 MiB | 11.37 MiB |
+| | 1.27× smaller | 1.71× smaller |
 
-The artifact figure is from a Drift database built with 42,000 file entries,
-revisions and network transactions, `VACUUM`ed and weighed — a model of the
-drive, not an export of the user's own database, and with one revision per
-file. The snapshot figures are the real transaction: 43.92 MiB on chain, and
-3.86× gzip measured on a 200 KB sample of it.
+Modelled on the real drive: 41,767 transactions, 98.4% files / 1.1% folders /
+0.5% drive, `bundledIn` populated (present on 188 of 189 sampled nodes), one
+revision per entity. Built with Drift, `VACUUM`ed, weighed.
 
-Size is the smaller win. The snapshot carries full GQL nodes with tags repeated
-per entity and base64-inflated metadata, which is why it loses even though JSON
-compresses better in ratio terms.
+The model is anchored at one end: the same sample gives 1058 bytes per snapshot
+entry, and 1058 × 41,767 = 44.2 MiB against the snapshot's real 43.92 MiB — a
+0.6% error, so the sample is representative.
 
-### Producing it is cheap, which is the real inversion
+**Known omissions, all of which would move the number up**: no
+`customJsonMetadata`/`customGQLTags` (the real nodes carry 11 tags, exactly the
+standard ArFS set, so these appear genuinely absent), no license or ArNS rows,
+and one revision per file. More revisions inflate both sides — a snapshot holds
+one entry per transaction, the database one row per revision — so the ratio
+should hold.
 
-A snapshot stores chain data, so producing one **re-reads the chain**: ~420
-paginated GraphQL queries plus ~42,000 metadata fetches for this drive, which
-does not complete on a throttled connection (see
-`SNAPSHOT_CREATION_FROM_SNAPSHOTS.md`).
+A 1.27× win on disk is modest. This is not the reason to build it.
 
-A state artifact is already sitting in the local database. Producing it is
-export → encrypt → upload. No GraphQL walk, no per-entity fetches.
+### 1.2 Work — the actual reason
+
+Reading a snapshot costs, **per entity**: one AES decryption
+(`file_entity.dart:117`, inside `FileEntity.fromTransaction`), one JSON parse,
+one insert. Private file metadata is encrypted with a per-file key, so the
+decryption is unavoidable and individual.
+
+On this drive that is ~41,000 of each — the ~80 seconds of `Processing chunk of
+1000 transactions` visible in a real sync log.
+
+An artifact costs **one** decryption and a bulk insert.
+
+Producing one is cheap in the way snapshots are not. A snapshot stores chain
+data, so producing it re-reads the chain: ~420 paginated GraphQL queries plus
+~42,000 metadata fetches for this drive, which does not complete on a throttled
+connection (`SNAPSHOT_CREATION_FROM_SNAPSHOTS.md`). An artifact is already in
+the local database — export, sign, compress, encrypt, upload.
+
+### 1.3 Privacy — an improvement, not a cost
+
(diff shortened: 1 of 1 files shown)


arielmelendez:
## Counter-proposal on the section format: make the payload a SQLite file

OQ1 is closed on the implementation branch as **D1 — JSON, gzipped**. I think that
call should be reopened, and I have measurements rather than opinions. The short
version: D1 gave three grounds, measurement contradicts the first, and the other
two argue the opposite way once the goal is unbounded drive size.

The proposed replacement is a **SQLite database file with a frozen artifact
schema** — not Drift's schema, and not a serialisation of rows. Everything else
in §3.3 is unchanged: `build → gzip → sign as a data item → AES256-GCM`.

### 1. "Buys little once compressed" is not what the numbers say

41,099 synthetic `file_revisions` rows, real column set, 43-char base64
transaction ids, 900 bundles, 459 folders, gzip -6. **One table**, so the
absolute figures are much smaller than the branch's seven-section 52.16 MiB —
the ratios are what transfer, not the values.

| format | raw | gzipped |
|---|---|---|
| SQLite, Drift's schema, `VACUUM` + index | 18.75 M | 10.14 M |
| **JSON row-wise — what D1 ships** | **27.25 M** | **7.50 M** |
| **SQLite, frozen schema, BLOB ids, no index** | **8.04 M** | **6.09 M** |
| JSON column-wise (arrays per column) | 15.06 M | 5.23 M |
| JSON column-wise + dictionaries | 8.08 M | 4.44 M |
| Parquet, binary ids + delta ints, zstd | — | 4.06 M |
| Hand-rolled binary columnar | 4.38 M | 3.86 M |

Two things worth pulling out. **Row-wise JSON is the worst end of a 1.94×
spread** — the shipped format, not a middle option. And **3.86 M is the entropy
floor**: that row's raw and compressed figures are only 1.13× apart, because
roughly a third of this payload is transaction ids and UUIDs, which are random
by construction and which no encoder can compress. Everything above that floor
is overhead, and the ceiling on any format choice here is under 2×.

Parquet was measured and rejected on its own: smallest of the off-the-shelf
options, but I could find no maintained Dart writer, the producer is a Flutter
web tab, and it has no notion of the named-section container §6 requires. Its
margin over transposed JSON is 9%, which never justified the implementation.

### 2. "Needs a spec and a codec on both sides" describes the status quo

That was D1's argument against binary. It is also an accurate description of
what the branch now carries: `drive_state_export.dart` and
`drive_state_import.dart` are 1,819 and 1,484 lines, and ardrive-core-js and
ardrive-cli must each reimplement both from a document.

ArFS already has one house serialisation format for this job, in the Snapshot
entity, and it is cumbersome and poorly supported. This is the same bet a
second time.

SQLite's file format is documented, frozen, and committed to be readable
through 2050, with a test suite far larger than its own source. `node:sqlite`
is built into Node. sql.js is already vendored in this repository.

### 3. The decisive argument is scale, and it is about the reader

Producing is close to a tie — both reach O(1) memory in drive size if written to
stream; SQLite just gets there for free.

Reading does not tie. `ATTACH` plus `INSERT INTO main.x SELECT … FROM artifact.x`
is bounded-memory by construction, runs in C, and allocates **no Dart objects at
all** — which is the cost §1.2 names as the actual reason for this feature. The
JSON equivalent needs either a streaming JSON parser, which the Dart SDK does
not have, or length-framed row groups we design, version and own.

Under a fixed-size goal that difference is affordable. Under "any drive size" it
is the whole question.

### 4. The WASM objection is already paid

I assumed a browser TypeScript reader would have to take on a WASM SQLite build,
and that this was decisive. It isn't — the repo already ships one:

```
web/sql-wasm.wasm   1,200,440 raw   457,879 gzip   385,501 brotli   (sql.js)
web/sqlite3.wasm    1,348,108 raw   505,121 gzip   439,650 brotli   (Drift's)
```

`web/index.html:332` loads `js/sql-wasm.js` unconditionally on every page load,
and has since 2022. The marginal cost to this client is zero.

### 5. Build up, never tear down

This is the part that must not be got wrong, and it is the one place the obvious
implementation is dangerous.

**SQLite does not erase deleted data.** `DROP TABLE` and `DELETE FROM` move pages
to the freelist; the bytes stay in the file. So a pipeline that copies the local
database and then removes `profiles` would publish `encryptedWallet` and
`keySalt` — the complete offline password-attack package §2.1 describes —
permanently and irrecoverably, unless a later `VACUUM` also ran. That makes the
safety rail conditional on someone remembering it.

Inverted, the risk disappears by construction:

```sql
ATTACH DATABASE '<vfs path>' AS artifact;
-- create the frozen tables in artifact
INSERT INTO artifact.file_revisions
  SELECT fileId, driveId, name, parentFolderId, size, …   -- explicit columns
  FROM main.file_revisions WHERE driveId = ?;
-- one statement per table; profiles is never named
INSERT INTO artifact.meta VALUES (?, ?, ?, ?);
DETACH DATABASE artifact;
```

Nothing withheld is ever written, so nothing has to be removed. **This also
supersedes D7**: the `INSERT … SELECT` column list *is* the typed projection,
expressed once, in SQL, in the statement that physically constructs the file.
That is strictly better than either the Dart projection D7 chose or the database
views D7 deferred, and the schema-drift test carries over unchanged.

### 6. Reading, and where §2.4 lands

§2.4's second argument survives intact and is why the schema must be frozen
rather than Drift's: the wire format stays decoupled from `schemaVersion`.
Shipping Drift's tables measured *worst on the board* above, so this is not only
a correctness point.

§2.4's first argument — the malformed-database CVE history — is narrower than it
reads, because the artifact is signed by the drive owner and, on private drives,
encrypted under the drive key. Feeding a hostile file requires the owner's wallet
or the drive key, and with either you already own the drive. The residual is a
compromised owner attacking someone who accepted a shared drive, which is real,
and which the read path should gate explicitly:

1. Decrypt, verify the data item signature against the drive owner, gunzip.
   Every failure is a fallback. Unchanged.
2. Open **read-only** with `SQLITE_DBCONFIG_DEFENSIVE` and
   `SQLITE_DBCONFIG_TRUSTED_SCHEMA=0`.
3. `PRAGMA integrity_check`, then refuse unless `sqlite_master` matches the
   frozen schema exactly — no views, no triggers, no virtual tables, no indexes.
4. Check `meta.version`, `meta.coverage`, `meta.privacy` against the tags,
   exactly as the branch already does against the signed payload.
5. `ATTACH`, then `INSERT INTO main.x SELECT …` per table, one transaction, same
   conflict modes as `drive_state_import.dart:956` uses today.

### Consequences for other decisions

- **D5 must be re-derived.** Its 100 MiB bound was refounded on the producer's
  memory — a peak near 950 MiB from a 263 MiB baseline. That peak is an artefact
  of row-wise JSON: 41k row objects, then 41k `Map`s, then a 52 MiB Dart string.
  `INSERT … SELECT` allocates none of them. Whatever bound survives should be
  measured against the new producer rather than inherited.
- **§6.1 becomes a `sqlite_master` check.** "A known section must be present even
  when empty" is exactly "the table exists and may hold zero rows", so
  absent-versus-empty stops being ambiguous on the wire at all.
- **D9 maps onto a `meta` table** with no change to the major/minor rule. A minor
  addition is a new nullable column or a new table older readers do not select —
  which requires explicit column lists everywhere, never `SELECT *`.
- **D2, D3, D4, D10, D11 are unaffected.** The tables carried and the rules about
  which ones travel are unchanged.

### Known costs

- **~37% more bytes** than the best JSON option — 6.09 M against 4.44 M — linear
  in drive size, paid on every publish.
- **A read gate that must be written deliberately**, per §6 above. It is the one
  failure here that would be worse than today's behaviour rather than equal to it.
- **A rewrite of two files that currently pass 1816 tests.**

### What has to be proven first

Two prototypes. Both cheap; either one failing sends this back to transposed JSON
with framed row groups, which is the fallback and still worth 1.69× over what
ships today.

1. `ATTACH` a second database in Drift's WASM VFS on web, build it, `DETACH`, and
   read its bytes back out.
2. `ATTACH` a received file in that same VFS and `INSERT INTO main.x SELECT …`
   across it.

Both are core SQLite, but the VFS is the part this repository has never
exercised, and the whole case rests on them holding on web.

---

If this is convincing, it belongs in `docs/drive-state/DECISIONS.md` on #2188 as
**D12**, reversing D1 and superseding D7. I have the table row written in the
house style and can post it there.

If it isn't, the cheap version of the same finding stands on its own and needs no
architectural argument: **transposing the existing JSON to arrays-per-column with
dictionaries takes it from 7.50 M to 4.44 M compressed, and 27.25 M to 8.08 M
raw**, with no new dependency and no change to any other client. The raw figure
is the one D5 refuses against.

Either way it is free to change now and a D9 major bump once anything reaches
chain.

<sub>Measurement scripts are synthetic and single-table; happy to re-run them against a real export if that would settle it.</sub>

PR #2188, opened 2026-08-19
PE-9205: Implement the drive state artifact

vilenarios:
Implements the drive state artifact specified in #2187 — an encrypted, signed snapshot of one drive's local rows, published to Arweave so another client can restore the drive without walking its GraphQL history.

**Includes #2187.** This branch carries the design document as well as the implementation, so the code's section citations resolve for a reviewer. It targets `dev` rather than the design branch because CodeRabbit reviews are disabled for non-`dev` bases, and a change this size should not skip review. #2187 can merge first or be closed as superseded — either way the doc lands once.

The spec on this branch is **corrected** against what was actually built: the operation order is gzip→sign, `Content-Encoding` must never be set, the payload carries a signed coverage claim, and a known section must be present even when empty. Those corrections are the diff against #2187.

**Nothing is published and nothing changes for users.** Both paths are behind `AppConfig` flags that default to `false` in all three flavours: `enableSyncFromDriveState` (read) and `enableDriveStatePublishing` (write). No artifact has ever been posted to any network — every upload collaborator is mocked in tests. The first upload is deliberately a human action.

## Trying it

`Ctrl+Shift+Q` opens dev tools; toggle `enableDriveStatePublishing`. The item then appears in the **New** menu for a private drive you own that has files in it. `configVersion` is deliberately **not** bumped — the flag ships as `false`, so a bump would deliver nothing while wiping every user's stored config.

## What it does

- **Export** one drive's rows — drive, folders, files, all three revision tables, licences — through explicit `selectOnly` projections. Key material cannot travel: a test reads the *generated SQL* and fails if a withheld column or a `SELECT *` appears in it.
- **Seal**: gzip → sign as an ANS-104 data item → AES256-GCM under the drive key.
- **Discover** by GraphQL, owner-filtered, with "the indexer did not answer" kept distinct from "this drive has no artifact".
- **Import**: validate everything before writing anything, then merge in one transaction.
- **Compose** into sync as its first source, ahead of snapshots, falling back silently at every failure.

## Decisions worth a reviewer's attention

**Revisions travel, and every revision does, not just the newest.** Entry rows alone restore a drive that renders an **empty file list** — `filesInFolderWithLicenseAndRevisionTransactions` INNER JOINs `network_transactions` through the newest `file_revisions` row. Measured on a real 42k-file drive: 1.05 revisions per entity, 4.7% with more than one, so full history costs ~5% more rows. **This is the one binding call made without a sign-off**; it is the superset, so narrowing to latest-only later is a producer-side change needing no reader change.

**`network_transactions` is regenerated on import, never carried.** It has no `driveId` — publishing it would publish rows about the user's *other* drives to everyone holding this drive's key. The importer derives it through the same helpers sync uses, and marks the rows mined; `_asMined` argues that trade in full.

**Coverage is signed, and the tags are checked against it.** `Block-Start`/`Block-End` are chosen by whoever posts the transaction and nobody signs them. Re-tagging a genuine artifact with a higher range would advance every importing client's watermark across blocks whose rows it never carried — files silently missing, nothing logged. The claim lives in the payload; a mismatch on either end is refused. The producer reads its watermark and its rows in one transaction, so it cannot tag a range its own payload contradicts.

**`Content-Encoding: gzip` must never be set on this entity**, though an earlier draft of the spec said to. `ar-io-node` indexes it from both L1 transactions and bundled data items and echoes it onto the data response; what the gateway serves is GCM ciphertext, so a browser would try to gunzip ciphertext and fail with `ERR_CONTENT_DECODING_FAILED`, with no opt-out. Tags are immutable — every artifact published with it would be permanently unfetchable.

**A known section must be present even when empty.** An absent section and an empty one are indistinguishable on the wire and mean opposite things. A producer built earlier on this branch signs a three-section payload that verifies, counts correctly, and claims honest coverage — and restores a drive whose file list is empty. Unknown sections are still ignored, so additive extension survives.

**A sync that skipped entities cannot publish.** Sync advances `lastBlockHeight` regardless of skips, so publishing from that state makes a gap permanent and immutable. Skip state is three-valued: "no sync has run" is not "clean".

## What two independent reviews found after the first push

CodeRabbit and an independent QA agent reviewed this separately. Both were worth running: between them they found ten defects that the lanes' own tests, and about eighty mutation checks, had all passed over. The pattern is consistent — the guards we wrote hold, and what got missed was things nobody thought to guard.

- **A ghost folder's children imported unreachable, permanently.** The export drops folders with no revision — which is what a ghost folder *is* — while keeping the files inside them, which do have revisions. Nothing lists a file except by its parent, so those files were invisible, and the import stamped the watermark past the range that would have healed them. Both reviewers found this independently. The consumer now materialises the stand-in the way sync does, so the drive lands in the state the producer is in: a ghost with a Fix button.
- **A second, ungated entry point** into the publish flow, predating the feature flag, so the item appeared with publishing switched off. The rail tests all passed — each asserted a property of a *function*, and none could see a call site nobody handed it. There is now a test that reads the source.
- **`syncDeep` was short-circuited by the artifact**, so the only "start over" remedy the interface offers re-applied the cause.
- **A probe-skipped drive read as "clean"** for the publish precondition, though that sync never examined it — which is the D3 rail that stops a gap being published permanently.
- **The merge compared against a timestamp the consumer had fabricated**, so a stand-in always outranked the real row and an artifact could never heal a ghost.
- **`enableSyncFromDriveState` could not be switched on by anyone** — no config key, no dev-tools toggle.
- **Only the newest candidate was ever tried**, so the window where an L1 artifact is indexed before it is retrievable closed the feature for that drive.
- **A modal overflowed 204px at 320×568**, carrying the Close button off a screen whose only job was explaining a refusal.

## Measured, not modelled

Nobody had run an export at real scale. `drive_state_scale_measurement_test.dart` now does, at 41,767 files, and it replaced the proposal's figures — which had been weighed on a VACUUMed SQLite file, from when the plan was to publish the database itself.

| | was documented | measured |
|---|---|---|
| serialised | 34.63 MiB | **52.16 MiB** |
| gzipped | 6.65 MiB | **9.55 MiB** |
| seal / import / render | — | 4s / 9s / 4s |

The size argument is therefore *weaker* than claimed, which only strengthens the document's own conclusion that size was never the reason to build this. The first run of that measurement reported 2.09 MiB at a 26× ratio, because its transaction ids interpolated a counter; real ids are 32 bytes of entropy and about a third of the payload.

It also showed the **producer** is the expensive half — peak near 950 MiB from a 263 MiB baseline — and that is the half running in a browser tab. And that an imported artifact was being re-downloaded and re-merged **every sync forever**, 8.5s to write zero rows.

## Verification

1816 passing, 5 skipped; `flutter analyze lib test` clean. Every guard was mutation-tested — the mutation applied, the failing tests recorded, the mutation reverted. That practice earned its keep here: it caught two tests that passed against buggy code, a guard that could not distinguish having Turbo credits from having the Turbo transport, and three defects that only appeared when independently-green branches were combined.

Known gaps, stated plainly: nobody has run the app, no artifact exists on chain, and the import is untested above ~10 rows.

## Downstream

`docs/drive-state/DELIVERY_PLAN.md` tracks ar-io-docs, ardrive-core-js and ardrive-cli. One item is independent of this feature and should ship on its own: the published ArFS spec calls the snapshot metadata field `dataJson`, while both implementations use `jsonMetadata` — anyone building from the docs today reads no metadata from any snapshot.

https://claude.ai/code/session_01LFV6xYmFz2meXf5EW2M1FB

## Summary by CodeRabbit

* **New Features**
  * Added encrypted drive-state artifact creation, discovery, synchronization, import, and publishing.
  * Added a guided publishing flow with cost estimates, AR/Turbo payment selection, affordability checks, and status feedback.
  * Added support for versioned drive-state formats, compression, integrity validation, and safe fallback behavior.
  * Added configuration controls for enabling drive-state syncing and publishing.

* **Bug Fixes**
  * Improved sync coverage tracking and snapshot handling.
  * Added safer recovery for partial uploads, malformed artifacts, skipped entities, and folder-structure conflicts.

* **Documentation**
  * Added drive-state artifact specifications, delivery planning, and updated size measurements.


vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai full review

vilenarios:
On the `configVersion` suggestion — declining deliberately, and recording why.

Both flags ship `false`, which is already the `AppConfig` constructor default. A stored config missing the key therefore already yields exactly the value the asset file would deliver, so a bump delivers nothing. What it *would* do is replace every user's stored config wholesale — including any dev-tools toggle they had set, and `enableSyncFromDriveState` in particular.

The advice becomes correct the moment either flag ships `true` in the asset file, and that is recorded next to the flags so whoever flips one sees it.

Two other notes from this review round: the `45 character path` hyphenation nit points at a passage that was replaced in `2cde10e2e` with measured figures, so it no longer applies. The stale `qa_findings_test.dart` header was real and is fixed in `1510ff435` — those tests all pass now, and the header still said every one was expected to fail.

PR #2189, opened 2026-08-19
PE-9211: The shared file page, rebuilt around the file

vilenarios:
Stacked on #2186 — base is `PE-9210-sharing-hardening`, so this diff is only the redesign.

The shared file page, rebuilt around the thing it exists to show. Three problems drove it: the card resized whenever you touched it, the version list was a readout you could not act on, and the page kept asking the gateway questions the share link had already answered.

## The card stops moving

It had three things that resized it — a preview toggle and two accordions — so anything you clicked moved everything else, including the Download button out from under the pointer on its way to press it.

The file's identity and Download move into a fixed header. Below that: the file on one side, everything about it on the other, as two tabs in a panel of fixed height. **File details** and **Version history** swap in place; the card does not move. On a phone the column already scrolls, so the panel has no fixed height there and transaction details grow the card instead of scrolling inside it.

## The preview is the page

The show/hide toggle is gone. It had quietly lost its purpose: the preview opens on arrival, so the bytes are already fetched by the time Hide could be pressed; the widget stayed mounted, so nothing was freed; and a recipient who did not want to see what a stranger sent them had already seen it. It saved no bandwidth, freed no memory and prevented nothing — it only asked the recipient to decide something.

The box keeps a job, just not as a control. Every state that cannot show the file — still resolving, type unsupported, past the 100 MiB cap — now gets the same picture-over-a-sentence placeholder, so a 580×430 pane never holds one floating line of text, and Download beside it stays the way forward.

Inline on the share page, the bar under an image preview also stops repeating the name and type on a 96px floor: the card above already carries both. Expand is the only thing in that bar the page cannot say for itself, so inline it is the only thing left in it — 40px back to the image (334 → 374). Full screen keeps the name, because the card is gone there. The file explorer is untouched.

## The version list is a control

- `showRevision` is the general case behind the freshness banner's two shortcuts, so a recipient can pick any version the file has. The banner stays honest because "a newer version exists" is re-derived from where the selection landed.
- **Opening the list waits on nothing.** The version the link named is already in hand, so it renders from the first frame and the rest fills in around it.
- **A failed lookup costs the list and nothing else.** The row the recipient was sent stays, stays selected, stays downloadable, with one line and a Retry — the state a rate-limited connection actually lands in.
- A **pinned** link names its version pinned and still lets go: pinning is the sharer saying which version they mean, not a limit on a recipient who can reach every other version on chain anyway.
- Selecting a version moves the preview with the download target, so the page never gives two answers to "what am I looking at".

## Stop asking what the link already said

`getTransactionDetails` was being called on the preview path purely to read the `Cipher` and `Cipher-IV` tags off the data transaction — the two values the link carries as `c` and `iv`, and the same lookup the download path already skips for them. On a connection the gateway rate limits, that is a call that can fail and cost the preview outright.

Private previews now decrypt with what they were given, falling back to the lookup only when they were given nothing — a v1 link, or a version the link never described. Guarded on the transaction the way the download is, so selecting another version leaves the link's cipher behind with the bytes it belonged to.

Where that leaves the page:

| Stage | Requests |
| --- | --- |
| First paint | **0** — `_resolveFromPayload` emits, *then* schedules background work |
| Behind the paint | 1 GraphQL (freshness banner), + license only if one exists |
| Preview (private) | 1 HTTP — **was 1 HTTP + 1 GraphQL** |
| Download | 1 HTTP + 1 GraphQL — the integrity check, not metadata (see below) |
| Thumbnail (private) | 1 HTTP + 1 GraphQL — see below |
| Version history | 1 GraphQL, only when opened, guarded against re-query |

The freshness query is the one that cannot be removed: a link cannot know about revisions made after it was created. It runs behind the paint, so it costs nothing on screen.

**Download's remaining query is not metadata.** It is `_dataItemVerifierFor` fetching the data item's *signature* to verify the bytes (`ardrive_downloader.dart:1131`, added in #2175). It cannot move into the link: the link is the untrusted input being checked, so a signature copied into it would be attacker-controlled. It is not self-certifying either — `verifyDataItem` first checks that `sha256(signature)` equals the data item id the link named, so an attacker cannot substitute their own keypair without a SHA-256 second preimage. It starts before the byte stream so it overlaps the download, and it is capped at 10s.

Worth recording: `FILE_SHARING_REDESIGN_PLAN.md:275` promised this would ride the existing `getTransactionDetails` call rather than being a new round trip. The two really are one query — `getTransactionDetails` delegates to `getTransactionDetailsWithSignature` — but the call sites never shared a result and nothing memoizes them, so a private download used to issue the identical query twice. This PR removed the other one.

## Consistency

Every row in both tabs measures exactly 44px — the copy control sets that floor as the smallest a touch target may be, and the rest meet it. That includes the transaction-details disclosure header, which came from an `ExpansionTile` whose `ListTile` floor measured **58** against the 44 either side of it, so the list changed rhythm at the header and changed back after it. None of the knobs `ExpansionTile` exposes reach 44 (`visualDensity` moves in steps of four and overshoots to 42), so the header is built from the same row primitive as everything else. A test measures all of them rather than asserting the constraint.

Dates read as dates (`Mar 3, 2024`) with the exact UTC instant in a tooltip — the pairing the drive explorer already uses.

## Follow-ups, not in here

- **A private thumbnail still costs one GraphQL call.** The link carries `c`/`iv` for the *data* transaction, but a thumbnail is its own transaction with its own IV. A `tiv` field would close it — the same trick, one field wider.
- **`bundledIn` is carried but not used as a locator.** The link has it as `in`, and today it only proves an item is bundled so the download can skip the lookup safely. If `/{dataTxId}` fails on every gateway, that bundle id is a second route to the same bytes — real work (ANS-104 offset parsing), real resilience.
- **Goldsky as a GraphQL fallback**, so the remaining queries are not backstopped by an `arweave.net` that rate-limits. This is now the highest-value item of the three: it fixes version history, the freshness banner *and* the download integrity check at once. Probing live for a bundled ArDrive data item, `signature` and `owner.key` both come back from goldsky, permagate.io, turbo-gateway.com and vilenarios.com — and `arweave.net` answers `429`. On a rate-limited connection the integrity check therefore burns its 10s timeout and reports `notVerified`, which is a transport problem rather than a problem with the check.

## Removed

`sharedFilePermanentFileSharing`, `sharedFileHidePreview` and `sharedFilePreviewPaneHint` are gone from all six locales along with the things they labelled.

## Summary by CodeRabbit

* **New Features**
  * Redesigned shared-file details and version history panels with responsive layouts.
  * Added revision selection, status labels, loading states, retry handling, and accessible controls.
  * Added support for encrypted shared-file previews using link-provided credentials.
  * Improved preview placeholders and compact fullscreen controls.

* **Bug Fixes**
  * Improved revision matching for renamed or moved files.
  * Downloads now avoid unnecessary integrity checks unless requested.
  * Prevented transient verification states for immediately completed downloads.
  * Improved handling of unavailable files across download gateways.

* **Localization**
  * Updated English text and removed obsolete shared-file and download verification messages.

### DIFF
--- a/lib/blocs/file_download/file_download_cubit.dart
+++ b/lib/blocs/file_download/file_download_cubit.dart
@@ -46,6 +46,11 @@ abstract class FileDownloadCubit extends Cubit<FileDownloadState> {
   /// arrives would trap the user in a download that has, in fact, finished.
   static const _verdictTimeout = Duration(seconds: 20);
 
+  /// How long a verdict may take before the user is told one is being waited
+  /// on. Short enough to appear promptly on the paths that are genuinely slow,
+  /// long enough that an already-settled verdict never shows the step at all.
+  static const _verdictAnnounceDelay = Duration(milliseconds: 250);
+
   /// Turns [downloader]'s mid-stream recoveries into something the user can
   /// see.
   ///
@@ -93,12 +98,40 @@ abstract class FileDownloadCubit extends Cubit<FileDownloadState> {
   }) async {
     if (isClosed || state is FileDownloadAborted) return;
 
-    emit(FileDownloadVerifying(fileName: fileName));
-
     DataItemIntegrityResult result;
 
+    final verdict = downloader.integrity;
+
+    // Announce the check only if it is actually going to take a moment.
+    //
+    // Every verdict this app can now reach is settled before the last byte is
+    // written: AES-GCM is MAC-checked during the download, and everything else
+    // has no check to run. Emitting "Checking this file" unconditionally meant
+    // a modal that appeared and replaced itself in the same breath - a flash
+    // of a step that never happened. The state is kept rather than deleted
+    // because it is correct the moment a caller asks for `verifyIntegrity`
+    // again, and then it is worth showing.
+    var answered = false;
+    // Swallowed here and read properly below: this is only asking "has it
+    // landed yet", and `integrity` is documented never to complete with an
+    // error anyway.
+    final settled = verdict.then<void>(
+      (_) => answered = true,
+      onError: (Object _) => answered = true,
+    );
+    await Future.any<void>([
+      settled,
+      Future<void>.delayed(_verdictAnnounceDelay),
+    ]);
+
+    if (isClosed || state is FileDownloadAborted) return;
+
+    if (!answered) {
+      emit(FileDownloadVerifying(fileName: fileName));
+    }
+
     try {
-      result = await downloader.integrity.timeout(
+      result = await verdict.timeout(
         _verdictTimeout,
         onTimeout: () => const DataItemIntegrityResult.notVerified(
           'The integrity check did not answer in time',
--- a/lib/blocs/fs_entry_preview/fs_entry_preview_cubit.dart
+++ b/lib/blocs/fs_entry_preview/fs_entry_preview_cubit.dart
@@ -32,6 +32,18 @@ class FsEntryPreviewCubit extends Cubit<FsEntryPreviewState> {
   final ProfileCubit _profileCubit;
   final ArDriveCrypto _crypto;
 
+  /// What the data transaction is encrypted with, when the caller already
+  /// knows.
+  ///
+  /// A share link carries the cipher and IV of the transaction it names, so a
+  /// recipient previewing a private file has them before the page loads. Given
+  /// them, the preview decrypts without asking the gateway to describe a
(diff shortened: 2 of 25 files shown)


vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

PR #2192, opened 2026-08-25
PE-9212: Private sharing — the link declares itself, the page stops hanging, and nothing is asked twice

vilenarios:
Private file sharing was broken end to end, in several independent ways that each looked like the others. Every one was found by testing against a gateway that rate limits and does not index every field — which turned out to be the honest environment, not a degraded one.

## What a recipient saw

| Symptom | Cause |
| --- | --- |
| A keyless private link loaded as a normal public page, no key prompt | the link never said the file was encrypted |
| Download produced a file the OS could not open | nothing decrypted it; the ciphertext was written under the file's own name |
| Preview sat on "Loading file details…" forever | it waited on metadata it did not need, through reads that could not time out |
| Preview said "could not load", download failed with `JSNull is not a subtype of String` | one unindexed field broke the query that carries the cipher tags |
| "Couldn't check for other versions" while every other query worked | the history walk gave up on the first page it could not use |

## The fixes

**A private link now says it is private.** `file_share_cubit` resolved a private file's cipher on a bounded, unawaited lookup and treated the link as complete when it failed — which it does on any rate-limited connection. `c` does two unrelated jobs, and only one was considered: beside `iv` it names the algorithm, but **on its own it is the only thing telling a keyless recipient that a key is needed at all**. Without it the page reads a private file as public.

The IV genuinely cannot be recovered offline — random per upload, only on the transaction, no cipher column in the local schema. The algorithm can be inferred from size, so the link now always declares a private file encrypted and leaves `iv` absent. The inference cannot cause a bad decrypt: everything that decrypts requires `hasCipherDetails` (`c` **and** `iv`), and this only runs when `iv` is missing. The sharer is also told, rather than handed a link that looks finished.

**A download refuses to save what it cannot decrypt.** With no key, and the file's own metadata never read, it asks the transaction what it is; a `Cipher` tag ends the download with a named failure and a dialog that says the file needs a key. Deliberately *not* on every public download — metadata that parsed is proof the file is public, because encrypted metadata does not parse.

**One unindexed field no longer breaks the cipher path.** A gateway answered a real `TransactionDetails` request with `"anchor": null`. The schema declares it `String!`, so Artemis fails deserialization for the **whole query** — and that query is where the preview and the download read `Cipher`/`Cipher-IV`. The four fields added in #2175 for the data item integrity check (which is off) now live on a separate `TransactionDetailsWithSignature`, so the cipher path selects nothing a gateway can break it with.

**The preview no longer waits for what it does not need.** It was gated on the file's own record returning, which a private file could wait on forever. It needs the data transaction, something to decide the type from, and the key — a v2 link carries all three. The drive id it appears to need is never read on this path: `FsEntryPreviewCubit` hands it to `_getFileKey`, which returns the key it was already given.

**Every read that gates the page is bounded.** Five `getLatestFileEntityWithId` call sites plus the metadata and owner lookups could hang forever, leaving the page on its placeholder with no way out. (Four found by CodeRabbit after the first was fixed.)

**Version history no longer gives up early.** The walk filtered each page to transactions carrying a valid ArFS version tag and `break`'d when that emptied a page. It reads `HEIGHT_ASC`, so the first page is the file's *oldest* transactions — the ones most likely to predate that tag. The newest-first query pages past them, which is why the two disagreed about the same file.

**A transaction is fetched once, not once per caller.** When a link lacks `c`/`iv`, the preview resolved them and then the download asked for the same transaction again — so previewing a file could make downloading it impossible on a rate-limited connection. `getTransactionDetails` now memoizes by id and hands back the in-flight future, so concurrent callers share one request. A miss is not cached. This completes the rule the link schema exists to serve: **if the link carries it, no query; if not, one query for that one thing; and never the same query twice.**

## From a review of the whole link path

**The encoder wrote names the reader discarded** — `_encodeText` truncates to 255 *bytes*, `sanitizeName` drops over 120 *characters*. A 150-character name was spent on the wire and thrown away. Now shortened to what the reader keeps, on a rune boundary.

**The rule keeping file keys off servers was prose only.** On the hash route everything after `#` stays in the browser; on the `/share/{fileId}` path route that query *is* sent, into access logs and `Referer` headers. Now the builder refuses to produce such a link at all — an assert would have been no guard in the one build where the leak would be real. Latent — that route is not live — but a trap set for whoever implements it.

Checked and found sound, recorded so nobody redoes it: the content-type table (64 entries, no duplicates, codes derived from the table rather than maintained beside it); base64url throughout, so the `+`-as-space trap cannot apply; a bounds-checked reader that latches truncation and never throws; and encoder output (~935 chars max) that cannot exceed the reader's 2048 limit.

## Making the four paths agree

A pass over what this PR actually touches — preview and download, on the drive explorer and on the shared file page — found the refusal above had no counterpart beside it, and that the declaration meant to prevent the whole failure had a gap of its own.

**The preview painted the ciphertext the download refused to save.** The explorer decides privacy from the drive record; the shared file page decides it from whether a key arrived. So on the one link this PR is about — private, missing `c` — the page reads it as public and hands it to a preview that fetches the bytes and renders them: mojibake in a document, a decoder error in an image or a PDF, while the download beside it gives a clear "needs a key". Two answers to one question, and the preview is the one the recipient sees first. It now makes the same check on the same tag, running *beside* the preview rather than in front of it, so no public link pays a round trip before its first paint. The transaction is memoized, so the preview and the download share one request rather than making two.

**`c` alone declares privacy — and the declaration skipped itself when only `iv` came back.** The guard returned early if *either* value was present, so a transaction answering with an IV and no cipher produced a link carrying `iv` and nothing to say the file was encrypted. That is precisely the link the declaration exists to prevent, since the recipient's locked state keys off `c` alone. It now gates on `c`, and drops an IV with no cipher to be used with.

**The one unbounded read in a PR about bounding reads.** The encryption preflight had neither a timeout nor a cancellation check, and it sits between the recipient pressing Download and anything happening. The memo made it worse rather than better: an entry is only removed once its future settles, so a request that never settled would be handed to every later caller for the life of the service — and `GraphQLRetry` has no timeout of its own. Bounded at the source, where it covers every caller, and the preflight now returns when the download has already been cancelled.

**The retry the dialog claimed to offer did not exist.** `retryCipherDetails` was never called from anywhere, and the sharer was shown a message telling them to close the dialog and share the file again. It is now a Try again button, and nearly free — the memo answers instantly for a transaction whose lookup has since succeeded.

**Both private image paths dereferenced a nullable file key.** `_getFileKey` returns null whenever the drive key cannot be produced, and the resulting crash was swallowed by a caller that reported it as an ordinary unpreviewable image with nothing in the log. They also left the static preview notifier spinning when they bailed out, which shows up under the *next* image the page previews.

## Verification

`flutter analyze` clean; **1435 passing / 4 skipped**. New coverage for: the ciphertext refusal and its two negative cases, private *image* decryption from the link cipher (images take their own branch and deliver bytes through a notifier, so the existing video test did not establish it), the memo including concurrent callers and the not-cached miss, the gateway's real `anchor: null` response, and three encoder round trips. From the pass above: a link built from an IV with no cipher, a retry that succeeds after a failure, the preview painting before its check answers and retracting when it does, its three negative cases, cancellation during the preflight, and the refusal to put a key in a path route's query.

Note for reviewers: `getTag` is an extension over `TransactionCommonMixin.tags`, so stubbing `getTag` on a mock is never consulted — the new tests use a fake carrying real tags.

## Known limits, not fixed here

- A private **download** still needs one query for the IV when the link could not carry it. Irreducible without the gateway; the fix is the Goldsky fallback.
- Links already shared without `c` cannot be repaired — the cipher is not in them. Recipients now get a clear "needs a key" message from the download and a retracted preview, rather than a corrupt file and rendered ciphertext, but those links still need re-sharing.
- Storing cipher/IV locally at upload would make private sharing work offline-first and remove this class of problem. Schema change; worth its own ticket.

## Summary by CodeRabbit

* **New Features**
  * Shared-file previews now verify encryption status and retract previews when encrypted content lacks an available key.
  * Downloads clearly identify encrypted files that require an access key and prevent unusable ciphertext from being saved.
  * Share links display recovery guidance when encryption details cannot be loaded, with a **Try again** option.
  * Link creation now safely handles oversized or unsafe file names and prevents accidental key exposure.
* **Bug Fixes**
  * Improved reliability for transaction lookups, metadata resolution, pagination, and shared-file loading.


vilenarios:
Closing and reopening to re-trigger CI. The `PR build and deploy` run created at 15:34 was orphaned by the GitHub Actions incident that began at 15:11 — the API reports it `queued` while refusing both cancel ("already completed") and re-run ("workflow file may be broken"). Other jobs on this branch have since drained normally, so this is one stuck run rather than a queue still burning down. No code change.

vilenarios:
@coderabbitai resume

vilenarios:
@coderabbitai review

PR #2202, opened 2026-08-28
fix: a preview timeout that measured duration, not silence

vilenarios:
A 48 MiB file could not be previewed from **any** gateway. Both attempts died at exactly the budget:

```
Gateway turbo-gateway.com failed for tx Y2qNfMZM...: TimeoutException after 0:00:10.000000
Gateway permagate.io      failed for tx Y2qNfMZM...: TimeoutException after 0:00:10.000000
```

Nothing was cold and nothing was broken. `getSandboxedTx` buffers the whole body behind one `Future`, and **a `Future` timeout is a deadline on the transfer** — so 10s meant *"this file must arrive at 5 MB/s"*. Previews are capped at 100 MiB, so any large file was abandoned mid-body by every gateway in turn while it was arriving perfectly well. Users saw "unpreviewable", and a second refresh sometimes beat the clock — which is why it looked intermittent.

Measured directly:

| | Result |
| --- | --- |
| `turbo-gateway.com/{tx}` | 302, 145 bytes, 0.39s |
| `{sandbox}.turbo-gateway.com/{tx}` | 200, **50,485,029 bytes**, 3.45s |

DNS 0.23s, TLS 0.47s — neither was the cost. The body was.

## The fix

The waterfall reads the body as a **stream**, and a `Stream` timeout fires only when no chunk has arrived for that long. The budget now means **silence**, which is what a timeout should mean: a slow body finishes, a dead gateway is still dropped in the same 10s.

`BrowserClient` in `http` 1.3.0 reads through a `ReadableStreamDefaultReader`, so this is chunk-wise on the web too, not only on the VM — verified before relying on it.

**No size thresholds and no larger per-read budget.** Those only move the cliff. The one constant that changed is the waterfall total, 25s → a 3 minute backstop; at 25s it was itself a deadline, needing 2 MB/s for the same file.

## Scope

Only the preview waterfall. `_syncFetch` keeps the buffered read and its own budget — sync's reads are a few hundred bytes each and there are hundreds of them, so a wall clock is the right shape there and one less moving part. `_tryGateway` is called from both, which is why the streamed read is a separate method rather than a change to it.

The sandboxed URL is preserved via the app's own `arweaveSandboxUrl` helper, so this does not start relying on the gateway's 302.

## Tests

Both halves pinned:
- a body arriving in chunks over 500ms **completes** under a 200ms budget
- a gateway that goes quiet for longer than its budget is **still dropped**

The first fails if the timeout ever goes back to being a deadline — checked by reverting it.

`flutter analyze` clean; **1440 passing / 4 skipped**.

## The 206 question, settled

`_tryGateway` accepts `200–208` without sending a `Range` header, which looked like it could return a truncated body as a complete file. Tested against every gateway in the waterfall — a plain GET with no `Range`:

| Gateway | Response |
| --- | --- |
| turbo-gateway.com | `200`, full `content-length: 50485029` |
| permagate.io | `200`, full length |
| ardrive.net | `200`, full length |
| arweave.net | `429 Too Many Requests` |

`206` comes back **only** in response to a `Range` request (`content-range: bytes 0-1023/50485029`), exactly as the spec requires. The preview path sends no `Range`, so it cannot receive one. Not reachable, and left alone.

Incidentally: `arweave.net` answered `429` here too — the last resort in this waterfall is rate limited for ordinary reads, which is the same liability #2195 removed from the GraphQL fallback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when retrieving data through fallback gateways.
  * Slow, steady transfers can now complete without timing out, while stalled connections still fail promptly.
  * Increased the overall retrieval window from 25 seconds to 3 minutes.
  * Preserved fallback behavior when gateways return errors or unavailable responses.

### DIFF
--- a/lib/services/arweave/data_gateway_fallback.dart
+++ b/lib/services/arweave/data_gateway_fallback.dart
@@ -132,15 +132,21 @@ class DataGatewayFallback {
   /// only way to tell a stall timeout from a deadline.
   final Duration _dataRequestTimeout;
 
+  /// Injectable for the same reason: the backstop is three minutes, and a test
+  /// has to be able to reach it to prove the connection is hung up.
+  final Duration _dataTotalTimeout;
+
   DataGatewayFallback({
     required ArioSDK arioSDK,
     Client Function()? clientFactory,
     @visibleForTesting Duration? syncRequestTimeout,
     @visibleForTesting Duration? syncLargeBodyRequestTimeout,
     @visibleForTesting Duration? dataRequestTimeout,
+    @visibleForTesting Duration? dataTotalTimeout,
   })  : _arioSDK = arioSDK,
         _clientFactory = clientFactory ?? Client.new,
         _dataRequestTimeout = dataRequestTimeout ?? _requestTimeout,
+        _dataTotalTimeout = dataTotalTimeout ?? _totalFetchTimeout,
         _syncRequestTimeout = syncRequestTimeout ?? _requestTimeout,
         _syncLargeBodyRequestTimeout =
             syncLargeBodyRequestTimeout ?? largeBodyRequestTimeout;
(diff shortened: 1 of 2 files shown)


vilenarios:
@coderabbitai review

PR #2208, opened 2026-08-29
feat: a sync nobody asked for no longer takes the app hostage

vilenarios:
Stacked on #2207 — review that one first; this PR's base is `sync-ui-1-surface-phases`, so the diff shown here is only the second commit.

## The problem

Every sync took the whole app hostage behind a full-screen scrim, including the ones nobody asked for. Log in and you are locked out until sync finishes, even though the drives you want are already in the local database from last session.

## What changes

A `SyncTrigger` (`userInitiated` / `background`) rides on `SyncInProgress` and `SyncCancelled`.

- **A sync you asked for** — Resync, deep resync, retry, attaching a drive — keeps the modal exactly as it is today.
- **A sync that just happens** — the login sync, and the periodic auto-sync — gets an indicator in the top bar instead, and the app stays usable.
- **`SyncCompleteWithErrors` still blocks.** Errors need an answer.

The refresh button becomes that indicator: a progress ring around the existing glyph, with the phase, percentage and elapsed time in the menu.

## Two things worth a reviewer's attention

**The default is `userInitiated`.** All thirteen call sites outside `SyncCubit` keep today's behaviour untouched — exactly one sync passes `background`. That is deliberate: a new sync path added later blocks (annoying, visible) rather than silently going quiet (a bug nobody notices).

**Mostly a move.** 473 of the ~1500 added lines are the sync modal relocating out of `app_shell.dart` into `lib/sync/presentation/sync_overlay.dart`, matching the feature-folder layout the repo already uses. It also makes `blocksTheApp` testable on its own, which is what pins the guarantee above.

## Also fixed here

- The **periodic auto-sync** was defaulting to `userInitiated`. Inert today (`autoSync` is false in every flavour) but it would have taken the app hostage the day that flag flipped — the exact bug this PR exists to kill.
- `SyncLoadingDrives` is `SyncCubit`'s *initial* state and is emitted by the metadata-only login path, so a scrim flashed up and vanished on **every login**. It no longer blocks.
- **The status was unreadable on a narrow phone.** `SyncButton` is also mounted in `MobileAppBar`, where there is no hover, and the menu is anchored to the trailing edge with `flutter_portal`'s `shiftToWithinBound` off by default. On a 320px screen the title started at **x = −48**. There is now a test pinning it, and `AxisFlag` is exported from `ardrive_ui` — `Aligned`'s own parameter type was missing from the `show` clause, so the option was unreachable for every consumer.
- The ring keeps visibly moving instead of freezing into a static arc — sync progress plateaus, and a motionless arc reads as hung.
- One name for the sync (`syncingAllDrives` / `syncingSingleDrive`) instead of a third near-synonym; whole percentages instead of `42.0%`; one glyph colour across idle and syncing.

## Verification

`flutter analyze` clean on the app and on `ardrive_ui`. Full suite **1487 passed, 4 skipped, 0 failed**.

Every new test was mutation-tested — the implementation line reverted, the failure observed, then restored. That caught two tests that passed for the wrong reason, and one that **hung instead of failing** when `SyncButton` was unwired from the cubit (a single-subscription `StreamController`, now broadcast: 8 failures in 3s instead of a wedge that burns CI's wall clock).

## Known, not fixed

Tapping the status text dismisses the menu. The overlay's dismiss barrier swallows taps for *all* dropdown content — ordinary items included — so this is pre-existing and fixing it means re-architecting a barrier every dropdown in the app depends on. Out of scope here.

## Not touched

No batching or concurrency arithmetic, no watermark or rewind, no snapshot validation, no extra network requests. *What* sync does is unchanged; only who it interrupts.

## Summary by CodeRabbit

* **New Features**
  * Added a unified sync overlay for user-initiated syncs, including progress, cancellation, and error recovery actions.
  * Background and login synchronization can now run without blocking the app.
  * Added a sync status indicator with animated progress, phase details, elapsed time, and resync access.
  * Added retry support for drives that fail during synchronization.
  * Dropdowns can now display an optional status header.

* **Tests**
  * Added coverage for sync indicators, trigger behavior, overlay visibility, and progress rendering.


vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

vilenarios:
Valid for this PR read in isolation, and resolved further up the same stack rather than left standing.

`SyncOverlay.blocksTheApp` does not exist by the top of this series — the blocking modal it gated was removed, and `SyncOverlay` now draws only a gateway indicator in non-production flavours. So the missed modal this describes has no modal to miss by the time the stack lands on `dev`.

The residue the finding points at is real and is also handled: a sync the user presses while one is already running is refused rather than queued (`startSync` returns false on `SyncInProgress`), the menu's actions are disabled for the duration, and the selection bar keeps its selection when a run is refused so the press costs nothing.

Merging: the intermediate state is not what ships.

PR #2209, opened 2026-08-29
feat: a sync that finished says what it found

vilenarios:
Stacked on #2208 (which is stacked on #2207). Review those first — this PR's base is `sync-ui-2-non-blocking`, so the diff here is only the third commit.

## The problem

A finished sync told the user nothing. On success the cubit emitted `SyncIdle()` and the modal simply vanished, so a sync that pulled in nothing looked exactly like one that pulled in five hundred files. The zero-change case is the one that matters most: a sync that says "nothing new" teaches people the next one is safe to ignore.

## What changes

`SyncComplete` carries what the sync found, and two surfaces report it — a pill at the sync button for a background sync, the modal ending on the result for one the user asked for. Both take themselves away after four seconds. Neither asks for a click.

| case | reads |
|---|---|
| nothing changed | `Up to date — nothing new` |
| nothing changed, named drive | `Photos is up to date — nothing new` |
| changes | `1 item changed` / `12 items changed` |
| changes, named drive | `12 items changed in Photos` |
| unreadable only | `2 updates could not be read` |
| both | `12 items changed · 2 updates could not be read` |

## Three things worth a reviewer's attention

**`SyncComplete extends SyncIdle`, deliberately.** `drive_detail_cubit.dart:144` and `sharing/sharing_file_listener.dart:40` both gate on `is SyncIdle` — the post-sync drive refresh and the shared-file handoff. A sibling state would have broken both silently. There is now a test feeding a `SyncComplete` through the drive-detail stream.

**It carries a `sequence`, not just a timestamp.** `SyncState` extends `Equatable` and bloc skips a state equal to the current one, so two "nothing new" syncs in a row would have been identical objects and the second would never have emitted — the exact case this PR exists for. `completedAt` alone was not enough: two mocked syncs run back-to-back land in the same millisecond and the test went intermittently red.

**A sync that failed says nothing.** The `catch` in `startSync` called `addError` without returning, so a sync that threw before any drive was queried fell through into the completion block with `hasErrors` false. That would have shown the "Failed to sync drive" snackbar and a "Sync Complete — nothing new" card at the same time. Both paths now carry `ranToCompletion`, and anything else emits a plain `SyncIdle()` — making no claim, exactly as before.

## The count is real, and it counts entities

`SyncProgress.entitiesSynced` was set nowhere in `lib/` — 0 in both factories, no `copyWith` anywhere. So every "N new items" string was dead and the app would have said "Up to date — nothing new" after syncing five hundred files.

It is now accumulated in `sync_repository.dart` beside the existing `_skippedEntityTxIdsByDrive`, from the revisions actually inserted. Two details that are easy to get wrong, both pinned by tests:

- **Not from the lists the batch hands back.** `_addNew*EntityRevisions` seed their return lists from the cache before testing whether anything changed, and every sync re-walks the last `kBlockHeightLookBack` blocks — so counting those reports a drive full of new items on a sync that wrote none. Test: a second sync over unchanged entities reports `0`; counting the returned lists gives `2`.
- **Entity ids, not revisions.** One file arriving with three historical versions writes three revisions but is one item to the reader, and a first sync walks a drive's whole history. Test: `Expected: <2> Actual: <3>` if revisions are counted.

Counting only — no change to control flow, batch sizes, transaction boundaries, the watermark, snapshot validation, progress arithmetic, or any query.

## Why the strings say "changed" and "updates"

A revision is written for a rename, move, hide, new version, thumbnail or licence — so "new items" would be false in the common case of a sync that picked up renames from another device. And the skipped count is a set of **transaction** ids, so three unreadable revisions of one file are three of them; calling those "items" would send someone hunting for three missing files.

## Verification

`flutter analyze` clean. Full suite **1535 passed, 4 skipped, 0 failed**.

Every new test was mutation-tested — implementation line reverted, failure observed, restored. The end-to-end count test drives a real `SyncRepository` over a real in-memory `DriveDao` with only `ArweaveService` mocked, and asserts the reported number against actual database row counts, because building the state by hand is exactly how the fake count got through the first review.

## Known, not fixed

The summary's `Material` has no colour, so it paints `canvasColor` at the corners — but the pre-existing `SyncCancelled` modal does the same thing. Worth one pass across all of them rather than making the new one the odd one out.

## Summary by CodeRabbit

* **New Features**
  * Added completion summaries for syncs, showing newly synced items and updates that could not be read.
  * Added distinct messaging for up-to-date syncs and single-drive syncs.
  * Added non-blocking summary notifications that automatically dismiss and refresh for each completed sync.
  * Added localized text for sync results.

* **Bug Fixes**
  * Completed syncs now refresh drive details correctly, even when metadata updates are unavailable.

### DIFF
--- a/lib/components/app_top_bar.dart
+++ b/lib/components/app_top_bar.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:ardrive/blocs/drive_detail/drive_detail_cubit.dart';
 import 'package:ardrive/blocs/drives/drives_cubit.dart';
 import 'package:ardrive/blocs/hide/global_hide_bloc.dart';
@@ -9,6 +11,7 @@ import 'package:ardrive/search/search_modal.dart';
 import 'package:ardrive/search/search_text_field.dart';
 import 'package:ardrive/sync/domain/cubit/sync_cubit.dart';
 import 'package:ardrive/sync/domain/sync_progress.dart';
+import 'package:ardrive/sync/presentation/sync_summary.dart';
 import 'package:ardrive/user/name/presentation/bloc/profile_name_bloc.dart';
 import 'package:ardrive/utils/app_localizations_wrapper.dart';
 import 'package:ardrive/utils/plausible_event_tracker/plausible_custom_event_properties.dart';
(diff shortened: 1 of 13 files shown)


vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

PR #2210, opened 2026-08-29
feat: a progress bar that keeps moving, and says when it cannot

vilenarios:
Last of four. Stacked on #2209 → #2208 → #2207, so **this PR's preview build contains the whole series** — it is the one link to test.

This is the only PR in the series that changes a number sync computes rather than adding one, so it is briefed and reviewed harder than the others.

## The problem

The drive walk ran 0 → 0.90, and then the entire tail was three fixed jumps: `0.92` for ghost folders, `0.96` for transaction statuses, `1.0` for completion. So the bar climbed fast and then sat perfectly still through phases of unknown length — transaction status updates go to the gateway and are bounded only by a 30s timeout. A bar that has not moved in half a minute is what a hung app looks like.

## Where the bar's travel goes now, and why

| phase | share | why |
|---|---|---|
| drive walk | 0 → 0.85 | The only phase whose length grows with the user's history, and the only one already reporting continuously. Travel belongs where progress is measured. |
| ghost folders | 0.85 → 0.88 | Local writes over an in-memory map, usually empty. Reports per row. |
| snapshot / preference reads | 0.88 → 0.92 | Local queries, reported per chunk. |
| pending-tx local reads | 0.92 → 0.97 | Largest tail share — the last stretch backed by work that has actually finished. |
| gateway confirmations | 0.97 → 0.99 | **Smallest share, deliberately the inverse of its duration.** Points here would be crossed on no evidence. |
| completion | 1.0 | Reached exactly, and only on success. |

## The honest part

No weighting makes a determinate bar move through a phase with no measurable progress. An earlier cut of this PR gave that phase the *largest* share and produced the biggest jump in the bar's history — freeze at 84% for up to 30s, then leap 15 points, which reads worse than the 96%-then-4-points it replaced.

So the gateway phase now says it cannot be measured. `SyncProgress.isIndeterminate` is set for its duration; the modal renders an indeterminate bar and the top-bar ring drops its value so it sweeps as well as turns. The user sees motion and a phase name instead of a number sitting still. No ETA, and no invented time-based ramp — an estimate that lies is worse than the stopwatch already there.

## Two regressions caught in review, both fixed

**The high-water mark had no ceiling.** Progress is now clamped monotonically at the point of publication, but the walk's value is unbounded above: it derives from block heights against a head read *once* at the start of a sync that can run for minutes, so a transaction in a newer block pushes it past 1. Reproduced at 1.2 then 2.4 — after which *every* later emission, "Sync complete" included, was pinned at 2.4. A bar reading 240% that never moved again, and a successful sync that never reported 1.0. Fixed by bounding to `[0,1]` before the value can become the mark, and by capping the walk at its own phase end so an overshoot below 1.0 cannot pin the tail either. Both cases are tested.

**The new reporting aborted real work.** `Future.timeout` does not cancel its source, so after the 30s status-update timeout the abandoned loop kept running while the controller closed; the new progress callback then threw into it and killed the remaining confirmation batches — invisibly, because a fired timeout drops late errors. Confirmed with 15001 pending transactions and a slow gateway: 3 of 4 batches issued instead of 4, leaving uploads "pending" for another cycle. The sink now no-ops once closed, and a test drives that exact scenario.

## Verification

`flutter analyze` clean. Full suite **1559 passed, 4 skipped, 0 failed**.

Tests run through a real `SyncRepository` over a real in-memory `DriveDao` with only `ArweaveService` mocked. Every guarantee is asserted on **both** the all-drives and single-drive paths, which are separate code with separate callees.

Two rounds of review found tests that passed without guarding anything — the original "tail advances" assertions counted the three fixed phase boundaries, so deleting every intra-phase callback left them green. They now count distinct values **strictly inside one phase's open interval**, which only intra-phase reporting can satisfy, and each was proved by deleting only the callback and observing the failure.

## Known, not fixed

Swapping widget types across the indeterminate branch destroys `LinearPercentIndicator`'s state, so the determinate bar returns from empty. It is at most a frame or two at the very end of a sync and no visible duration could be demonstrated; the obvious fix (`animateToInitialPercent`) does not exist in percent_indicator 4.2.3.

Cancelling inside the gateway phase leaves `isIndeterminate` set on the stream's final event. Harmless today — the cubit emits `SyncCancelled`, the overlay swaps to the cancelled modal, and the next sync resets before emitting. Clearing it would mean touching the cancellation path, which this PR's constraints put off limits.

### DIFF
--- a/lib/blocs/drive_detail/drive_detail_cubit.dart
+++ b/lib/blocs/drive_detail/drive_detail_cubit.dart
@@ -108,6 +108,9 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
       });
     } else {
       Future.microtask(() async {
+        // No loading state to emit first: the cubit is constructed in
+        // DriveDetailLoadInProgress, so the screen is already saying it is
+        // working before this wait begins.
         // Wait for any current sync to complete before checking drive state
         await _syncCubit.waitCurrentSync();
 
@@ -208,6 +211,10 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
   }
 
   void showEmptyDriveDetail() async {
+    // Nothing to announce before this wait either. This runs when there is no
+    // drive to show, from a state that is already DriveDetailLoadInProgress,
+    // and all it can do afterwards is narrow that to "empty" - so emitting a
+    // loading state here would claim work that is not happening.
     await _syncCubit.waitCurrentSync();
 
     // Check if state has already changed (e.g., drives were loaded during sync)
@@ -227,7 +234,9 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
     // First check current drive state before waiting for sync
     var drive = await _driveDao.driveById(driveId: driveId).getSingleOrNull();
 
-    // If drive doesn't exist locally at all, wait for sync to discover it
+    // If drive doesn't exist locally at all, wait for sync to discover it.
+    // This one already emits before it waits, which is the shape openFolder
+    // now has too.
     if (drive == null) {
       emit(DriveDetailLoadInProgress());
       await _syncCubit.waitCurrentSync();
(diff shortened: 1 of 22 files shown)


vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai full review

PR #2214, opened 2026-08-31
feat: a login lands on a list of drives, not inside one

vilenarios:
Stacked on #2213. Eighth and last of the sync-UX series — the one the rest was building toward.

## What it does

A login lands on a **list of your drives** instead of inside one the app hasn't looked at yet.

Each row carries the drive's name, whether it's public or private, whether it's shared with you, when it was last synced, how many items it holds, its size, and when it was created. A drive that has never been walked says so, and withholds the numbers it cannot know.

## Why this is more than a nicer landing page

An audit of the login story found a long list of ways a user gets stranded or misinformed, and nearly all of them share one cause: **the app opens a drive before it knows enough to do so honestly.**

| What the user saw | Because |
|---|---|
| “Getting Started” | the page needed *a drive* to render, and the list hadn't arrived |
| “Drive Not Synced” | it opened a drive whose contents were never fetched |
| A blank screen | a bookmarked `/drives/{id}` wasn't in an empty table |
| A drive you'd hidden | `lastSelectedDriveId` was restored without checking visibility |

A drives list has none of those failure modes, because the drive list is **the one thing login actually fetches**. `updateUserDrives` is cheap — drive entities only, no history walk — and it still runs with “sync drives on login” switched off.

## Decisions behind it

- **The sidebar doesn't change.** It stays navigation only. Drive metadata lives here; duplicating it in two places is how the two drift apart. It does still *navigate* — a sidebar tap opens that drive from the list.
- **Size and item counts are withheld until a drive has been walked.** `COUNT` and `SUM` only see synced rows, so “0 bytes / no items” on an unsynced drive is the same confident wrong answer this series exists to remove.
- **One list, with a marker for shared drives.** Not two sections; the smaller group gets missed.
- **First login gets a single “Sync All Drives”.** Every drive reads “Never synced”, so the page offers the thing the old auto-sync used to do — asked, rather than assumed.
- **No schema change.** Per-drive last-synced is a JSON map in the existing key-value store. A `.drift` column needs `build_runner` plus a `schemaVersion` bump, and this repo's migration fixtures only cover v17–v19.

## Deep links are untouched

`/drives/{id}`, `/drives/{id}/folders/{id}` and file share links are permanent public API, and links already in the wild must keep resolving. The parser gained one branch — the bare `/drives` segment — and the existing `length > 1` branch that resolves a drive id is **byte-identical**. 15 routing tests cover it.

## States

Reusing the vocabulary from #2213 rather than inventing a fourth: **loading** says “Loading your drives…”, **empty** is today's Getting Started (correct only for a genuinely new account), **failed** says the list could not be loaded and offers Try Again, and **populated** is the list.

## What review caught

Two rendering passes across the full width band found four things reading could not have:

- **The retry told the user they had no drives.** Tapping Try Again painted “Getting Started” for the whole retry, because `syncMetadataOnly()` emits `SyncLoadingDrives` *before* making the request, which re-fired the subscription onto an empty table. That is the exact prohibition of #2213, reintroduced in this PR's own retry path.
- **The sidebar was inert on the new landing page** — full of tappable drives, none of which did anything.
- **The header and the rows disagreed about the layout**, so a 972–1004px window showed a table header over stacked cards.
- **A landscape phone showed no drives at all** — the sync-all card filled the viewport and nothing scrolled.

And after the fixes, three more in the screens *around* the list: the empty state overflowed 720–877px (a breakpoint measured at 720 for a pair of cards needing ~828 — introduced by extracting them out of `ScreenTypeLayout`); both the loading and error headlines fell through to Material's body colour and became the faintest text on their own screen in both themes; and “Try Again” was still the full-bleed red slab that had just been capped on the button beside it.

## Verification

`flutter analyze` clean. Full suite **1760 passed, 3 skipped, 0 failed**.

Every test proved by reverting its implementation line. The layout tests read the **horizontal** axis and assert the body really is the width under test, so a clamped surface fails loudly rather than passing quietly — both traps this series produced before. The header/row agreement is checked at 186 widths in 2px steps.

## Known

There is no in-app affordance back to the list — browser back works and the URL is `/drives`. Adding one means touching the sidebar or top bar, which the sidebar decision rules out, so it is left for a follow-up.


vilenarios:
@coderabbitai full review

PR #2215, opened 2026-09-03
PE-9205: A sync you can read through, and choose the drives for

vilenarios:
Stacked on #2214. The five steps from `docs/SYNC_PROGRESSIVE_READABILITY.md`, executed end to end.

## What changes for a user

- **A drive the sync has finished can be opened**, while the rest of the run continues. A ten-drive sync no longer holds the first drive shut for the nine that follow.
- **A drive being synced can be read, and fills in as it goes**, instead of sitting behind a panel until the run ends.
- **Navigating a folder mid-sync just works.** No wait, no loading card.
- **Sync the four drives you care about**, not all ten. One run over four, not four runs.
- **A partial failure says what it cost and fixes itself in one press** — "1 of 2 drives could not be read", plus what is still true, plus a button over exactly the drives that failed.

## Fault tolerance

Each claim rests on a property, not a hope:

- `syncedDriveIds` is appended **only** on the path that reached the end, so the widened gate can never release a drive that failed or was cancelled.
- Dropping redraw ticks is safe only because another is coming. The tick with nothing behind it is a run's last, so a run that ends having dropped anything **re-reads the folder once**.
- A folder is read from committed transactions either way, so what comes back mid-sync is consistent — just less than the folder will eventually hold.
- Choosing drives is one run, not several. A second sync is still refused rather than queued.

## Three faults found while building it

1. A four-of-ten run carries no single drive id, so it read as an all-drives sync — ten rows claiming to sync, ten drives held for a run over four.
2. Pressing **Sync selected** during a sync cleared the ticks and was then refused, costing a selection for nothing.
3. **Clear** was wired to toggle-select-all, which selects everything when some are unticked.

## Verification

2012 tests (up 29), `flutter analyze` clean on `lib` and `test`.

## Known gap

The 500ms redraw interval is a **starting point, not a measured optimum** — the tuning method is in the doc comment. Measuring it needs a browser and a wallet with thousands of revisions.

## Summary by CodeRabbit

- **New Features**
  - Select multiple drives to sync, deep-resync, or clear selection.
  - Stop active syncs from the sync menu and drive detail view.
  - Retry drives that fail during a multi-drive sync.
  - Pending upload confirmations are monitored automatically.
- **Bug Fixes**
  - Drive contents remain available while syncing, with clearer loading status.
  - Sync updates are smoother and more reliable during uploads.
  - Storage totals now exclude hidden and unrelated drives and include accurate file counts.
  - GraphQL requests recover more reliably from rate limits and server errors.
  - Detaching a syncing drive now stops its sync and explains recovery steps.

### DIFF
--- a/lib/blocs/drive_detail/drive_detail_cubit.dart
+++ b/lib/blocs/drive_detail/drive_detail_cubit.dart
@@ -593,6 +593,12 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
           }
 
           if (_activityTracker.isUploading) {
+            // Dropped, and owed back. An upload writes its file while this
+            // flag is up, so the tick carrying it is the one refused here -
+            // and nothing else is going to write that folder afterwards, so
+            // no later tick will carry it either. Remembering the debt is what
+            // lets the refresh at the end of the upload settle it.
+            _droppedFolderRedraw = true;
             return;
           }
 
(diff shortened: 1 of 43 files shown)


vilenarios:
@coderabbitai full review

PR #2218, opened 2026-09-05
PE-9205: What a returning or brand-new user is told about their drives

vilenarios:
Two independent things a returning or brand-new user sees, both found in staging.

## Telling a returning reader what changed

Coming back to an existing session restores everything from the local database and syncs nothing, which is right - IndexedDB survives a closed tab, so counts, sizes and block heights are all still there, and `autoSync` is false in every flavour. The cost is that nothing on the page says how old any of it is. Close the tab on Tuesday, come back on Friday, and Tuesday's file counts are drawn with exactly the confidence of current ones. The only cue is a `Last synced` cell you have to read and do arithmetic on.

The one prompt this page has is gated on `nothingHasEverBeenSynced` - *every* drive unwalked - so it only ever fires on a first login. Returning readers get silence.

So: one query, at the exact moment the app currently goes quiet (the "nothing owed" branch of `_syncOnlyIfThereIsWork`). `probeActiveDriveIds` already exists and the sync already uses it to skip unchanged drives; the new `probeDrivesWithChanges` runs the same per-owner query read-only - it writes nothing, touches no sync state and starts no sync. If drives have moved, a neutral notice above the table names how many and offers to sync **only those**. If nothing moved, nothing is said.

An unanswerable probe also says nothing. That is deliberately the opposite of the fallback inside a sync, where `isComplete: false` means *sync everything*: there the cost of guessing wrong is a slower sync, here it would be a banner claiming changes nobody has confirmed, which is a nag.

Two bugs surfaced while testing it, both mine:

- `close()` shuts the subject down before it awaits its way to `super.close()`, so there is a window where `isClosed` is still false and adding throws `Cannot add new events after calling close` - out of an un-awaited future, where nothing catches it. Now guarded on the subject, not the cubit.
- The failure swallow lived in the repository, leaving the cubit's fire-and-forget call unprotected: an exception there was an unhandled async error rather than a failed request. Now caught at the boundary that is actually un-awaited.

## The new drive's two offers

On "You're on chain!" the buttons were `ButtonVariant.secondary`, which on the dark card is `solidGrey700` against a `solidGrey800` background - while the **disabled** token is `solidGrey600`. The button that worked was less prominent than one that would not, which is a fair description of "doesn't look clickable". Both are primary now. This page has exactly two cards and its whole job is to get one of them pressed, so two primaries is the design rather than a competition.

On a phone those two offers were 283px squares stacked under a heading and a paragraph: **239px of scroll on a 375x667 screen, with Create Folder at y=866** - two hundred pixels below the fold, on the one page whose job is to say "well done, now do one of these two things". Laid on their side they take about 130px each and the whole page is visible at once.

The `SingleChildScrollView` stays: at a large enough text scale something has to give, and a page that scrolls degrades better than one that overflows.

The four-card empty states are deliberately left alone. They have the same grey-on-grey problem, but four red buttons is a wall rather than a fix.

## Testing

13 new tests. The phone layout is *measured* rather than asserted - `maxScrollExtent` is 0 and both buttons are proven on screen; removing the fix puts the numbers back to 239px and y=866. The two withdrawal guards on the notice were each checked by deleting them and confirming the matching test fails.

Full suite green (2075 passing), `flutter analyze lib test` clean.

## Summary by CodeRabbit

- **New Features**
  - Drives now show unread network changes and offer targeted syncing.
  - Drive lists display available file counts and sizes before a full scan completes.
  - Pending uploads can be rechecked from the file menu.
  - Mobile drive search is now accessed from the app bar.
  - Sync status is refreshed when no other sync is running.

- **Style**
  - Improved new-drive empty states with compact mobile actions and clearer primary buttons.

- **Tests**
  - Added coverage for unread changes, targeted syncing, upload status refreshes, search access, and responsive layouts.

### DIFF
--- a/lib/app_shell.dart
+++ b/lib/app_shell.dart
@@ -7,10 +7,12 @@ import 'package:ardrive/components/side_bar.dart';
 import 'package:ardrive/components/sync_failure_test_panel.dart';
 import 'package:ardrive/misc/misc.dart';
 import 'package:ardrive/pages/drive_detail/components/hover_widget.dart';
+import 'package:ardrive/search/search_modal.dart';
 import 'package:ardrive/shared/blocs/banner/app_banner_bloc.dart';
 import 'package:ardrive/shared/blocs/private_drive_migration/private_drive_migration_bloc.dart';
 import 'package:ardrive/sync/domain/cubit/sync_cubit.dart';
 import 'package:ardrive/sync/presentation/sync_overlay.dart';
+import 'package:ardrive/utils/app_localizations_wrapper.dart';
 import 'package:ardrive/utils/logger.dart';
 import 'package:ardrive/utils/show_general_dialog.dart';
 import 'package:ardrive_ui/ardrive_ui.dart';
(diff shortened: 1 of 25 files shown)


vilenarios:
Three more, pushed onto this branch rather than stacked behind it - they touch the same `drives_list` cubit and state, and a PR based on a non-`dev` branch does not get reviewed here. Happy to split if you would rather review them apart.

### A drive you filled yourself is not a drive nothing is known about

A new user creates their first drive, uploads five files, opens Your Drives, and sees **`Never synced` · − · −**. The figures are not missing; they are withheld. `writeFileEntity` put those rows in `fileEntries` on upload and `driveContentSummaries()` reads them fine - the cubit was discarding them because `createDrive` inserts without a `lastBlockHeight` and the schema defaults it to zero.

And since it was their only drive, `nothingHasEverBeenSynced` was trivially true, so **"Nothing has been synced yet - their contents have not been fetched yet"** went up over contents they had just put there.

`hasBeenWalked` was answering two questions with one flag. *Has anything read this from chain?* - a block height knows that, and is right. *Do the local numbers mean anything?* - different question, and for a drive this device created and filled the answer is yes. Figures now show when the drive is walked **or** has a non-zero local count: a zero from a walked drive still means empty, a zero from an unwalked one still means unknown.

`Never synced` stays in the last-synced column. It is true, and nothing has checked the chain. It just no longer sits beside two dashes, which is what made it read as a failure.

### Mobile search moves into the bar

Desktop search is already global chrome - a field in `AppTopBar` beside the hide toggle. Mobile's was page furniture inside the explorer, taking a 60px band above the file list on the layout with the least height to give. The field did nothing until submitted, at which point it opened the same sheet the icon now opens.

`showSearch` is **opt-in**, and that is the load-bearing part: `MobileAppBar` is also worn by the drives list and the no-drives page. The drives-list subtree provides a `DriveDetailCubit` built against the root path rather than a chosen drive, and the no-drives page has nothing to search - so an always-on icon would offer search where it could only disappoint. Also drops a `TextEditingController` the explorer held and never disposed.

### Asking whether an upload landed, without a sync

A pending file could only be resolved by waiting up to twenty minutes or running a whole sync. `_updateTransactionStatuses` already runs wallet-wide and reads its pending transactions from the local tables rather than from the walk, so it never needed the sync around it. `refreshTransactionStatuses` exposes that pass alone - no history read, no watermark moved. It cannot find a file somebody else uploaded, only settle one already known about, which is exactly what a pending dot asks.

On the file kebab, **only while that file is pending**. Refused while a sync runs (that sync ends with this very pass), and on refusal or failure the row keeps its status and the app claims nothing.

**A correction worth recording:** I had said the twenty-minute watch does a full drive walk to confirm an upload, and that status-only would be a large saving. That was wrong - the probe skips drives with no newly-mined activity and the status pass runs regardless, so that path already costs about three queries. What did not exist was any way to ask *on demand*. The saving is the twenty minutes, not the queries.

### Testing

Full suite green (**2085 passing**), `flutter analyze lib test` clean.

Each guard was checked by removing it: reverting the figures fix gives `null` where 2 is expected; removing the one-at-a-time guard fails the decline test. I also caught and rewrote a vacuous test for the status refresh - it was verifying the mock rather than the cubit.

vilenarios:
@coderabbitai review

PR #2223, opened 2026-09-09
PE-9205: A search result from All Drives never actually opened its folder

vilenarios:
**#2221 shipped search on the drives list with navigation that does not work.** A file or folder result lands at the root of the right drive rather than at the thing that was searched for. This fixes that, and adds the file highlighting.

## The bug

`requestFolder` set a pending folder and `onDriveSelected` honoured it — but **`onDriveSelected` is only mounted on the explorer's branch of the router** and never sees a drive opened from the list. The road that actually runs is `OpenDriveOnSelection`, which hears the selection and calls `openDriveFromList` — and that cleared the pending folder outright.

**Why the tests passed:** they called `requestFolder` then `onDriveSelected` directly. A path that exists, and not the one a reader takes. The new group goes through `openDriveFromList`, and **three of its tests fail against what is live on staging right now**.

## A first attempt broke an existing test, correctly

Making `openDriveFromList` honour *any* pending folder broke `drives_list_routing_test`: *"opening a drive from the list opens it at its root rather than at whatever folder was last seen."*

That test is right. A folder deep link and a search request were sharing one pair of fields, and they are **not the same intent**:

- Arriving through `/drives/X/folders/Y` and later tapping X on the list should open X at its **root**. A link is not an instruction to resume.
- A search result is the opposite — the reader picked that folder out of a list of them.

Search requests now have their own fields, so `openDriveFromList` honours one and discards the other. The existing test needed no changes, which is what says the invariant was real and the shortcut was the problem.

## Highlighting the file

`DriveDetailCubit` takes an `initialSelectedItemId`, carried on the same one-shot road, so a file result arrives selected the way it does from inside a drive.

It is **keyed by drive id**: `_driveDetailCubit` is built twice over — once for the drives list against `rootPath`, once for the explorer against the chosen drive — and unkeyed, the list's cubit would swallow a selection meant for the explorer.

## A crash that was already there

`openFolder`'s `firstWhere` had no `orElse`, so a selection that is not in the folder **threw out of the folder load**. There are ordinary reasons for that: the file moved or was deleted between the search and the tap, or hidden items are being filtered out. Failing to highlight a row is a disappointment; throwing is a broken screen.

## Two more found by attacking this before shipping

Neither would have been caught by CI or by the happy path working:

1. **`clearState` did not clear the new fields.** Every sibling pending field is cleared on logout; a search from one session could have reached the next person to sign in on that device.
2. **A stale highlight window.** A request honoured in a frame where the explorer never built sat waiting, and would have highlighted a stale file the next time that drive opened. `showDrivesList` abandons it now.

## Testing

Full suite green (**2106 passing**), `flutter analyze lib test` clean.

Every guard verified by removal: the folder fix, the logout clear and the abandon-on-return all fail their tests when the code is taken out. The stale test group from #2221 was deleted rather than repaired — it was the one that gave false confidence.

## Summary by CodeRabbit

* **New Features**
  * Search results now open the containing folder and automatically select the specific matching item.
  * Item selection is preserved while navigating from the drives list to the requested folder.

* **Bug Fixes**
  * Opening a search result no longer fails when the item is unavailable or filtered out; the folder opens without an item selected.
  * Pending search selections are cleared when leaving the drives list or signing out, preventing stale selections from affecting later navigation.

### DIFF
--- a/lib/app_shell.dart
+++ b/lib/app_shell.dart
@@ -318,7 +318,8 @@ class _MobileSearchButton extends StatelessWidget {
   const _MobileSearchButton({this.onNavigateToFolder});
 
   /// See [MobileAppBar.onSearchNavigateToFolder].
-  final void Function(String driveId, String folderId)? onNavigateToFolder;
+  final void Function(String driveId, String folderId, {String? itemId})?
+      onNavigateToFolder;
 
   @override
   Widget build(BuildContext context) {
@@ -372,7 +373,7 @@ class MobileAppBar extends StatelessWidget implements PreferredSizeWidget {
   /// long-lived and can be told to open a folder in another drive. The drives
   /// list passes one, because selecting a drive there replaces the subtree and
   /// tears that cubit down mid-navigation. See [FileSearchModal.onNavigateToFolder].
-  final void Function(String driveId, String folderId)?
+  final void Function(String driveId, String folderId, {String? itemId})?
       onSearchNavigateToFolder;
 
   @override
--- a/lib/blocs/drive_detail/drive_detail_cubit.dart
+++ b/lib/blocs/drive_detail/drive_detail_cubit.dart
@@ -95,6 +95,14 @@ class DriveDetailCubit extends Cubit<DriveDetailState> {
   DriveDetailCubit({
     required String driveId,
     String? initialFolderId,
+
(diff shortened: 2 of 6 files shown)


vilenarios:
@coderabbitai review

PR #2225, opened 2026-09-10
Release v2.88.0

vilenarios:
Promotes `dev` to production. **95 commits since v2.87.1**, 17 merged PRs.

> **Do not squash.** This is the release merge.

## What's new

**A login lands on your drives, not inside one.** A new Your Drives page lists every drive with its file count, size and when it was last read, and syncs any selection of them rather than all or one.

**Sync you can read.** The phases that used to sit silent say what they are doing, a finished sync says what it found, and the progress bar keeps moving or admits when it cannot measure something. A sync nobody asked for no longer holds the app: you can navigate, open drives and upload while one runs.

**Search on All Drives.** Search has always looked across every drive, but was only reachable from inside one. It is now on the drives list too, desktop and mobile, and a result opens the file where it lives with the file selected.

**Coming back to an existing session.** Nothing re-syncs and nothing is lost. The app now tells you if drives changed while you were away, and offers to sync just those.

**A drive you just made shows what you put in it.** Creating a drive and uploading to it showed "Never synced" with no file count and no size, because nothing had read it from chain yet. The files were always there. They are shown now.

**Check an upload without a full sync.** A file waiting to be confirmed has a Check upload status action.

**Smaller things**
- The new drive's welcome page fits on a phone, and its buttons look like buttons
- A drive opened from the list is scrolled into view in the sidebar
- Shared links get a proper preview card when posted
- "Pinned" in a shared file's version list is now "This version"

## Under the hood

- An upload that cannot be found is called failed after two hours rather than eight
- Status is re-checked every 20 minutes while an upload is pending, and only then
- GraphQL retries only where a retry can change the answer, and the fallback endpoint now fires when it should rather than depending on a string match
- Faster CI: change to preview roughly halved

## Notes for upgraders

- No database migration: `schemaVersion` unchanged at 29
- No config change: `configVersion` unchanged at 3
- Uploads and encryption untouched

## Pre-flight

| | |
|---|---|
| `schemaVersion` | **29** both sides, no `.drift` changes. Migration fixtures only cover v17–v19, so an unnoticed bump would have shipped untested |
| `configVersion` | **3** both sides, no asset changes. No stale-config surprise for existing users |
| Uploads / crypto | `ardrive_uploader`, `ardrive_crypto`, `lib/core/upload`, `lib/core/crypto`, `lib/blocs/upload`: **zero lines changed** |
| Staging | Green through #2223, all deploys successful |
| Version | Bumped to 2.88.0 in #2224 |

## Accuracy of the notes above

Each claim was checked against `master` rather than written from memory. Two drafts were cut as a result:

- *"Sync skips drives with no new activity"* — `probeActiveDriveIds` is **already in master**, so it is not new here.
- *"Falls back to a second endpoint sooner"* — the change makes the fallback fire **when it should**; it previously string-matched `error.toString()` for status codes. Not a timing improvement.


vilenarios:
Replaced by the release branch, which follows the convention in this repo's history: a `release-vX.Y.Z` branch off `dev` with `master` merged into it, so conflicts are resolved on the release branch rather than against production. See #2204 and #2198 for the same shape.

PR #2231, opened 2026-09-16
PE-9205: Offer New Drive in the New menu with no drive open

vilenarios:
Users opened **All Drives**, pressed **New**, and found one row: *Advanced*, holding one action, *Attach Drive*. Several read that as a broken app.

## Why

Every item in that menu except two is gated on a drive being open, so on All Drives the uploads, folders, notes, pins, manifests and snapshots correctly drop out. Two things were wrong beyond that:

- **New Drive was gated on the drive list having loaded** (`drivesState is DrivesLoadSuccess`). Creating a drive does not depend on knowing which drives already exist.
- **Attach Drive sat under Advanced**, which with no drive open holds exactly one item. A menu whose only row opens a submenu to a single action reads as empty.

## What changed

1. **New Drive appears whenever you are logged in**, disabled only when the wallet cannot pay for it. Not gated on the drive list's state.
2. **With no drive open, Advanced is flattened**, so Attach Drive sits at the top level instead of behind a submenu.

3. **The mobile plus button does the same.** It runs its own item list and still nested Attach Drive under Advanced. That button lives on a drive page, so it only lands there while the drive has not loaded or cannot be read, but the rule is the reader's rather than the surface's. Found by CodeRabbit.

4. **The rule at the top of the mobile sheet is gone.** It separates the upload items from the rest, and with no drive open the group above it was empty, so it drew a hairline above the first item. Found by reading the menu end to end.

5. **New Drive is never greyed.** A wallet that cannot pay is not a reason to withhold the action, and the getting started cards on the same page already open the same dialog ungated.
6. **An empty wallet is sent to top up.** That dialog's only action was Cancel, so the path ended there. It now offers Buy Credits into the top-up modal, the same offer every other payment refusal makes.

Nothing in the page body changed. The New menu is the contextual menu for creating things, so a second control on the drives list would only repeat it.

## Honest note

This removes New Drive's dependence on `DrivesCubit` state rather than explaining it. I did not establish which state that page sits in when the menu is built, only that the gate was wrong either way, since the item should appear regardless.

## Testing

- New: with no drive open the menu offers New Drive and puts Attach Drive at the top level with no Advanced row, asserted for the sidebar menu and for the mobile plus button, which also asserts that sheet opens with an item rather than a rule.
- 5 menu tests and the full suite (**2,164 passing**) green, `flutter analyze lib test` clean.
- Not covered by a test: the dialog's Buy Credits action. It is one action on one branch, verified by analyze and by matching `showTurboPaymentRequiredDialog`; the top-up flow it opens is its own.
- Both behaviours checked by reverting the code and confirming a test fails.

https://claude.ai/code/session_01AAvnj5uGsaASPN47BViH8D

## Summary by CodeRabbit

* **New Features**
  * “New Drive” is now available whenever a profile is logged in, including when no drive is open, and remains enabled regardless of wallet balance.
  * Advanced actions appear directly in the menu when no drive is selected.
  * Added a “Buy Credits” option to the drive creation dialog when the account has no available balance.


vilenarios:
Valid, and fixed in 637b36733. `_getPlusButtonItems` now applies the same `hasDriveInView` rule as `_getNewMenuItems`: with no drive in view its advanced items render directly, and the Advanced row is only built when there is one.

On reachability, since it changes how much this matters rather than whether it is right: that menu is only built with `isBottomNavigationButton: true`, which is set once, on the drive detail page. So it lands in the no-drive branch only while that drive has not loaded or cannot be read, not on the drives list. The rule still belongs to the reader rather than to the surface, so both menus flatten alike.

`new_button_no_drive_test.dart` covers it with `isBottomNavigationButton: true` alongside the sidebar cases.

One note on the rest of this PR, since it changed under you: the New Drive button this branch added to the drives list body has been removed at the maintainer's direction. The New menu is the contextual menu for creating things, and a second control on the page repeated it. The page files are back to exactly what `dev` has, and the branch now touches only `new_button.dart` and that test.

vilenarios:
@coderabbitai review

Issue #2206, opened 2026-08-28
Private media previews show no progress while buffering and decrypting


vilenarios:
## Symptom

Previewing a **private** video or audio file shows a bare spinner with no progress for as long as the file takes to arrive. On a 47 MiB video at ~2.5 MB/s that is **~20 seconds of nothing**, then playback starts abruptly.

Nothing indicates that work is happening, how much is left, or that waiting is worthwhile. It reads as frozen, and the natural response — refresh — throws away the download and starts it over.

## Why it matters now

It didn't used to be visible. Before v2.87.1 a file that size **timed out at 10 seconds** and said "this file can't be previewed here" ([#2202](https://github.com/ardriveapp/ardrive-web/pull/2202) fixed that; the budget now measures silence rather than duration). The wait was always there — the timeout was hiding it behind a fast failure.

So we have converted *"fails quickly and wrongly"* into *"succeeds, but appears frozen"*. Better, and not yet good.

## Why private media is different

Public media never has this problem. `_mediaUrl` returns the gateway URL directly and the browser's own player streams it, with its own buffering and progress:

```dart
if (!isPrivate || isPinFile) {
  return previewUrl;
}
```

A private file has no URL any player can use — every gateway holds only ciphertext. So the app must **fetch the whole body, decrypt it in memory, wrap it in a blob URL**, and only then hand it over. `lib/blocs/fs_entry_preview/fs_entry_preview_cubit.dart:913-943`:

```dart
emit(const FsEntryPreviewLoading());          // ← the only state emitted
dataBytes      = await _fetchPreviewBytes(...);   // seconds to tens of seconds
decryptedBytes = await _decodePrivateData(...);   // opaque
objectUrl      = _objectUrls.create(...);
```

`FsEntryPreviewLoading` carries nothing — it is `super(previewUrl: '')` and its `props` are empty, so there is no place to put a percentage even if we had one.

## Scope

Four call sites emit that same bare `FsEntryPreviewLoading` (`:339` PDF, `:913` media, `:1022` document, `:1109` email). Ranked by how much it hurts:

1. **Private video / audio** — the worst, and the only one where the file is routinely tens of MiB
2. **PDF** — buffers whole for rasterising, *public ones too*
3. Documents / email — usually small, capped well below the media limit
4. Images — buffered and capped at 100 MiB, but progress matters less when the payoff is instant

[...]

PR #2238, opened 2026-09-22
PE-9205: Fetch a preview once, and stream it in the browser

vilenarios:
Groundwork for #2206. Its progress bar needs two things this PR fixes, and both are bugs in their own right.

## 1. A preview was downloaded twice, and again on every row write

The drive explorer previews a file once immediately, and again for every row its database watch emits. Drift emits the current row as soon as the watch starts. PDFs were guarded against this; private media, documents and email were not. Measured with a test against the cubit:

| Private video in a drive | Before | After |
|---|---|---|
| Opening it | 2 downloads | 1 |
| Each later write to its row (upload confirmed, sync) | 1 more download, and **the playing video restarts** | Nothing |
| A download that failed, then a row write | Retries | Retries |

For a private file that's up to 100 MiB per download, held in memory and decrypted. Documents and email flashed the spinner and reset the text's scroll.

**Change:** PDF's `_pdfDataTxId` guard becomes `_claim` / `_release`, shared by PDF, private media, documents and email. The claim is taken before the first `await`, so two calls arriving together can't both pass it. It is released on every failure, so a later write can still retry.

## 2. In the browser, gateway reads weren't streamed

`DataGatewayFallback` used `package:http`'s `Client()`. In the browser that's `BrowserClient`, which in the version we can use (1.2.1) is built on `XMLHttpRequest`: `send` completes only when the **last** byte has arrived, and the body comes as one chunk. So:

- **#2202's silence budget was still a deadline on web.** `send(request).timeout(10s)` covered the whole download, so a large private video on an ordinary connection timed out on the first gateway. The comment claiming `BrowserClient` "reads through a `ReadableStream`" is true only of http ≥ 1.3.
- **There was nothing to measure progress with.**

Upgrading `http` isn't possible: the fetch-based `BrowserClient` needs Dart 3.4, and Flutter 3.19.6 ships 3.3.4.

**Change:** `DataGatewayFallback` defaults to `gatewayHttpClient()`. That's `FetchClient` from ArDrive's `fetch_client` fork in the browser, which arweave-dart already uses to stream downloads, and the platform client elsewhere, via a conditional import. It's created with `RequestMode.cors`; its default, `noCors`, makes every response body opaque. `fetch_client` moves from override-only to a direct dependency; the lockfile resolves the same commit.

The swap covers every `fetchData` caller (previews, shared-link reads, downloads through the fallback), `fetchDataAtMost` and the manifest fetch. Sync reads go through `_tryGateway` with the Arweave client and are unaffected.

## Testing

- **3 new cubit tests** drive the explorer with a real stream of database rows. The existing media tests stubbed that stream as empty, which is how the double fetch went unnoticed.
- **Mutations:** removing the guard, removing the release, or unguarding documents each fails a test.
- **Not tested:** the email guard, because the parser is JS interop and doesn't run on the VM. The web client can't be unit-tested on the VM either; see below.
- `flutter analyze` is clean and the full suite passes (**2,310**), on Flutter 3.19.6.

## Please check on the preview build. This is the part only a browser proves

With the Network tab open:
1. **Public PDF, image, text and video previews still load.** This is the regression risk: CORS mode, gateway redirects to sandbox subdomains.
2. **A private video opens with one data request, not two.** Let it play while the drive syncs: it doesn't restart.
3. **Throttle to "Fast 3G" and open a private file of 30+ MiB.** It should load instead of failing over to another gateway after 10 s.
4. A shared-link page for a file and a download still work.

## Known and not changed here

In the explorer, the database watch re-previews using the **originally selected** item rather than the fresh row. So a new revision arriving while the panel is open never loaded the new data: before, it fetched the old data again; now, it doesn't fetch. Worth its own fix.

https://claude.ai/code/session_012m4TLHc7h1FW1ANDSLTFjf

## Summary by CodeRabbit

* **New Features**
  * Web downloads now stream data as it arrives, improving responsiveness when loading gateway content.
* **Bug Fixes**
  * Previews for videos, documents, and other files are less likely to trigger duplicate downloads when updates arrive repeatedly.
  * Failed previews can be retried after a later update, while successful previews avoid disrupting media that is already playing.


vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai review

